PR #15, opened 2022-06-17
feat: Flagd http provider

beeme1mr:


libs/providers/flagd-rest/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Installation
+
+```
+$ npm install @openfeature/flagd-rest-provider
+```
+
+## Building
+
+Run `nx package providers-flagd-rest` to build the library.
```
toddbaert, line 13: We should probably either wrap this in a package json script, or add `npx` to the command. I dont have `nx` install globally on purpose to use project local vertsions.

libs/providers/flagd-rest/src/lib/flagd-rest-provider.spec.ts
```diff
@@ -0,0 +1,151 @@
+import {
+  ErrorCode,
+  FlagValueType,
+  TypeMismatchError,
+} from '@openfeature/nodejs-sdk';
+import * as nock from 'nock';
+import { FlagdRESTProvider } from './flagd-rest-provider';
+
+nock.disableNetConnect();
```
toddbaert, line 9: I think you should do this in a before, and re-enabled it after.

toddbaert (approved):

PR #66, opened 2022-07-21
feat: flagd provider

james-milligan:
Signed-off-by: James-Milligan <james@omnant.co.uk>


libs/providers/flagd/.vscode/configurationCache.log
```diff
@@ -0,0 +1 @@
+{"buildTargets":["generate"],"launchTargets":[],"customConfigurationProvider":{"workspaceBrowse":{"browsePath":[],"compilerArgs":[]},"fileIndex":[]}}
```
skyerus, line 1: These .vscode files should be omitted

libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -0,0 +1,65 @@
+import {
+  EvaluationContext,
+  Provider,
+  ResolutionDetails,
+  ProviderOptions
+} from '@openfeature/nodejs-sdk';
+import {IService} from './service/IService'
+import HTTPService from './service/http/service'
+
+export interface FlagdProviderOpions extends ProviderOptions {
```
skyerus, line 10: typo

libs/providers/flagd/.vscode/configurationCache.log
skyerus, line 1: These too

libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  constructor(options?: FlagdProviderOptions) {
+      if (!options || options.service == undefined) {
+        this.service = new HTTPService()
+        return
+      }
+      this.service = options.service
+  }
+
+  contextTransformer = () => ({});
```
beeme1mr, line 29: This can be removed if it's not being used.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libs/providers/flagd/schemas"]
+	path = libs/providers/flagd/schemas
+	url = https://github.com/james-milligan/schemas.git
```
beeme1mr, line 3: Update to the real schema repo when you're ready to merge.

libs/providers/flagd/src/proto/ts/schema/v1/schema.client.ts
```diff
@@ -0,0 +1,76 @@
+// @generated by protobuf-ts 2.2.2 with parameter long_type_string,generate_dependencies
```
beeme1mr, line 1: Avoid version controlling generated code.

libs/providers/flagd/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@openfeature/flagd-provider",
+  "version": "0.0.1",
+  "type": "commonjs",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [[ $(npm show $npm_package_name@$npm_package_version version) = $(npm run current-version -s) ]]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "dependencies": {
+    "@grpc/grpc-js": "^1.6.7",
```
beeme1mr, line 10: These dependencies must be added to the root package.json

libs/providers/flagd/src/lib/flagd-provider.ts
james-milligan: removed

.gitmodules
james-milligan: awaiting changes in https://github.com/open-feature/schemas/pull/13

libs/providers/flagd/src/proto/ts/schema/v1/schema.client.ts
james-milligan: removed + updated nx to include generation step on test lint and package actions

libs/providers/flagd/package.json
james-milligan: added

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
thomaspoignant, line 1: I am not sure that we want to commit the vscode file?
Shouldn't we be agnostic from the IDE?

libs/providers/flagd/package.json
thomaspoignant: You should also remove them from this `package.json`

libs/providers/flagd/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@openfeature/flagd-provider",
+  "version": "0.0.1",
+  "type": "commonjs",
+  "scripts": {
```
thomaspoignant, line 5: You may have to add a `peerDependencies` block as @beeme1mr did in this commit https://github.com/open-feature/node-sdk-contrib/pull/64/commits/75dc1c5ccc06bcaa10273bad8b80ed8b1ea1bb50
```json
  "peerDependencies": {
    "@openfeature/nodejs-sdk": "^0.0.1"
  }
```

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export class GRPCService implements IService {
+    client: ServiceClient;
+    constructor(host?: string, port?: number, credentials?: grpc.ChannelCredentials) {
+        this.client = new client(host, port, credentials)
+    }
+
+
+    async ResolveBoolean(flagKey: string, defaultValue: boolean, context: EvaluationContext): Promise<ResolutionDetails<boolean>> {
+      // eslint-disable-next-line prefer-const
```
thomaspoignant, line 26: why do you have to disable eslint here ?

.vscode/settings.json
james-milligan: removed in latest commit

libs/providers/flagd/src/lib/service/grpc/service.ts
james-milligan: removed in latest commit

libs/providers/flagd/package.json
james-milligan: added and removed other dependencies

libs/providers/flagd/project.json
```diff
@@ -0,0 +1,93 @@
...
+    },
+    "package": {
+      "executor": "@nrwl/web:rollup",
+      "outputs": ["{options.outputPath}"],
+      "options": {
+        "project": "libs/providers/flagd/package.json",
+        "outputPath": "dist/libs/providers/flagd",
+        "entryFile": "libs/providers/flagd/src/index.ts",
+        "tsConfig": "libs/providers/flagd/tsconfig.lib.json",
+        "compiler": "babel",
```
beeme1mr, line 67: You may want to consider adding ` "buildableProjectDepsInPackageJsonType": "dependencies"`. NX will automatically add dependencies to the packaged project.json. You can then explicitly set peer dependencies in the package.json before hand (which you're already doing).

libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -0,0 +1,63 @@
+import {
+  EvaluationContext,
+  Provider,
+  ResolutionDetails,
+  ProviderOptions
+} from '@openfeature/nodejs-sdk';
+import {IService} from './service/IService'
+import HTTPService from './service/http/service'
+
+export interface FlagdProviderOptions extends ProviderOptions {
```
beeme1mr, line 10: Why extend provider options? It doesn't look only the service is being used.

libs/providers/flagd/src/lib/service/grpc/client.ts
```diff
@@ -0,0 +1,21 @@
+import * as grpc from '@grpc/grpc-js';
+import {ServiceClient} from '../../../proto/ts/schema/v1/schema.client'
+import { GrpcTransport } from "@protobuf-ts/grpc-transport";
+
+export default class GRPCClient extends ServiceClient {
```
beeme1mr, line 5: I would recommend avoiding default exports.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    async ResolveBoolean(flagKey: string, defaultValue: boolean, context: EvaluationContext): Promise<ResolutionDetails<boolean>> {
+        const res: ResolutionDetails<boolean> = {
+            value: defaultValue
+        }
+        const req: ResolveBooleanRequest = {
+            flagKey,
+            context: Struct.fromJsonString(JSON.stringify(context)),
+        }
+        await this.client.resolveBoolean(req, undefined).response.then(callRes => {
```
beeme1mr, line 33: I would recommend not merging async/await with promises like this. With await, you can refactor this to something like this:

```typescript 
try {
    try {
      const { response } = await this.client.resolveBoolean({
        flagKey,
        context: Struct.fromJsonString(JSON.stringify(context)),
      });
      return {
        value: response.value,
        reason: response.reason,
        variant: response.variant,
      };
    } catch (err: unknown) {
      return {
        reason: 'ERROR',
        errorCode: (err as Partial<RpcError>)?.code ?? StandardResolutionReasons.UNKNOWN,
        variant: 'default_value',
        value: defaultValue,
      }
    }
}
```

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,162 @@
...
+  ResolveObjectResponse
+} from '../../../proto/ts/schema/v1/schema'
+
+
+export default class HTTPService implements IService {
+    private url: string
+
+    constructor(host?: string, port?: number) {
+        if (host == undefined) {
+            host = "localhost"
```
beeme1mr, line 17:
```suggestion
            host = "http://localhost"
```
Axios complained about a missing protocol when using the default value.

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,162 @@
...
+        }
+        if (port == undefined) {
+            port = 8080
+        }
+        this.url = `${host}:${port}`
+    }
+
+    async ResolveBoolean(flagKey: string, defaultValue: boolean, context: EvaluationContext): Promise<ResolutionDetails<boolean>> {
+      // eslint-disable-next-line prefer-const
```
beeme1mr, line 26: See above comment about refactoring this.

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,162 @@
...
+            console.error(err)
+            response.reason = "ERROR"
+            response.errorCode = ErrorCode.GENERAL
+        })
+        return response
+    }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function CheckResponse(res: AxiosResponse<any>, valueType: string): boolean {
```
beeme1mr, line 143:
```suggestion
function CheckResponse(res: AxiosResponse<ResolutionDetails<unknown>>, valueType: string): boolean {
```

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,162 @@
...
+    if ((res.data)
+    && (res.data.value && typeof res.data.value == valueType)
```
beeme1mr, line 145:
```suggestion
    && (res.data.value && typeof res.data.value === valueType)
```

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,162 @@
...
+    && (res.data.variant && typeof res.data.variant == "string")
```
beeme1mr, line 146:
```suggestion
    && (res.data.variant && typeof res.data.variant === "string")
```

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,162 @@
...
+    && (res.data.reason && typeof res.data.reason == "string")
```
beeme1mr, line 147:
```suggestion
    && (res.data.reason && typeof res.data.reason === "string")
```

libs/providers/go-feature-flag/project.json
```diff
@@ -73,4 +73,4 @@
     }
   },
   "tags": []
-}
```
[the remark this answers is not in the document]
james-milligan: removed

libs/providers/flagd/project.json
james-milligan: added

libs/providers/flagd/src/lib/flagd-provider.ts
james-milligan: have removed the extend, is now standalone interface

libs/providers/flagd/src/lib/service/grpc/client.ts
james-milligan: removed and updated imports

libs/providers/flagd/src/lib/service/grpc/service.ts
james-milligan: refactored and mirrored in the http implementation

libs/providers/flagd/Makefile
```diff
@@ -0,0 +1,4 @@
+generate:
```
beeme1mr, line 1: Do you need this file anymore?

james-milligan: removed in latest commit

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function CheckResponse(res: AxiosResponse<ResolutionDetails<unknown>> | AxiosResponse<ResolveObjectResponse>, valueType: string): boolean {
+    if ((res.data)
+    && (res.data.value && typeof res.data.value === valueType)
+    && (res.data.variant && typeof res.data.variant === "string")
+    && (res.data.reason && typeof res.data.reason === "string")
+    ) {
+        return true
+    }
```
toddbaert, line 162: I think you can simplify this with [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html)

libs/providers/flagd/src/lib/service/IService.ts
```diff
@@ -0,0 +1,11 @@
...
+    EvaluationContext,
+    ResolutionDetails
+} from '@openfeature/nodejs-sdk'
+
+export interface IService {
+    ResolveBoolean(flagKey: string, defaultValue: boolean, context: EvaluationContext): Promise<ResolutionDetails<boolean>>
+    ResolveString(flagKey: string, defaultValue: string, context: EvaluationContext): Promise<ResolutionDetails<string>>
+    ResolveNumber(flagKey: string, defaultValue: number, context: EvaluationContext): Promise<ResolutionDetails<number>>
+    ResolveObject<T extends object>(flagKey: string, defaultValue: T, context: EvaluationContext): Promise<ResolutionDetails<T>>
+}
```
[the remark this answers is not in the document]
james-milligan: this interface has been made internal, as of the latest commit the only exports from the provider is the FlagdProvider itself (and its options interface)

libs/providers/flagd/src/lib/service/grpc/client.ts
```diff
@@ -0,0 +1,21 @@
...
+    constructor(host?: string, port?: number, channelCredentials?: grpc.ChannelCredentials) {
+        if (host == undefined) {
+            host = "localhost"
+        }
+        if (port == undefined) {
+            port = 8080
+        }
+        if (channelCredentials == undefined) {
+            channelCredentials = grpc.credentials.createInsecure()
+        }
```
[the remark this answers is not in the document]
james-milligan: I've replaced this will default values in the latest commit

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,165 @@
...
+    private url: string
+
+    constructor(host?: string, port?: number) {
+        if (host == undefined) {
+            host = "http://localhost"
+        }
+        if (port == undefined) {
+            port = 8080
+        }
+        this.url = `${host}:${port}`
```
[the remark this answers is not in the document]
james-milligan: I've replaced this will default values in the latest commit

libs/providers/flagd/src/lib/service/http/service.ts
james-milligan: ive trimmed it right down - and added a small unit test for it, happy to use optional chaining, but no longer checks for res.data.value first - if its undefined its not a string so returns false etc

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -0,0 +1,104 @@
...
+        return {
+          value: response.value,
+          reason: response.reason,
+          variant: response.variant,
+        };
+      } catch (err: unknown) {
+        return {
+          reason: StandardResolutionReasons.ERROR,
+          errorCode: (err as Partial<RpcError>)?.code ?? StandardResolutionReasons.UNKNOWN,
+          variant: 'default_value',
```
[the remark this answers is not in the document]
james-milligan: I've removed this value, I agree, if the error field is populated then you'll know you have the default value anyway

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    async ResolveBoolean(flagKey: string, defaultValue: boolean, context: EvaluationContext): Promise<ResolutionDetails<boolean>> {
+      try {
+        const res = await axios.post<ResolveBooleanResponse>(
+          `${this.url}/flags/${flagKey}/resolve/boolean`,
```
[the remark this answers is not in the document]
james-milligan: I cannot find any reference to it, so have added it to err on the side of caution

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -103,7 +103,7 @@ export class HTTPService implements Service {
   async resolveObject<T extends object>(flagKey: string, defaultValue: T, context: EvaluationContext): Promise<ResolutionDetails<T>> {
     try {
       const res = await axios.post<ResolveObjectResponse>(
-        `${this.url}/flags/${flagKey}/resolve/object`,
+        encodeURI(`${this.url}/flags/${flagKey}/resolve/object`),
```
toddbaert, line 106: I think you need to do this everywhere you are constructing URLs for safety. If a flag name had a `/` in it, it would break. TBH I would just use `encodeURIComponent` on just the flag name.

james-milligan: I have made this change, however after testing with flagd it seems that keys specifically containing ```/``` will still be parsed wrong by grpc-router, I will investigate the cause of this and open an issue.

beeme1mr: It might be better to update the schema to restrict the characters that are allowed to be used as the key.

https://github.com/open-feature/schemas/blob/main/json/flagd-definitions.yaml

We could restrict it to something like `a-z`, `A-Z`, `0-9`, `_`, and `-`.

toddbaert: > I have made this change, however after testing with flagd it seems that keys specifically containing / will still be parsed wrong by grpc-router, I will investigate the cause of this and open an issue.

> It might be better to update the schema to restrict the characters that are allowed to be used as the key.

Yep, I'm OK with this too. I did verify the HTTP framework decoded path params containing encoded chars properly, but it also had trouble with `/` specifically (even encoded). I'm fine with not supporting that char.

libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+        host,
+        port,
+      })
+    } else {
+      this.service = new GRPCService({
+        host,
+        port
+      })
+    }
+  }
```
toddbaert, line 46: Like this much better.

toddbaert (approved):
I tested this locally and it works as expected. The interface/config is also better IMHO. Approved, great work!

libs/providers/flagd/src/lib/service/http/service.ts
```diff
@@ -0,0 +1,155 @@
...
+
+interface HTTPServiceOptions {
+  host: string;
+  port: number;
+}
+export class HTTPService implements Service {
+  private url: string
+
+  constructor(options: HTTPServiceOptions) {
+      this.url = `http://${options.host}:${options.port}`
```
toddbaert, line 19: There is one small issue here - this locks us to http. We want to support HTTPS as well. Consider adding a protocol option or accepting a full URL.

toddbaert (changes requested):
https://github.com/open-feature/node-sdk-contrib/pull/66/files#r931240946

toddbaert (approved):

toddbaert:
Works great for me. I'm going to merge.

PR #69, opened 2022-07-21
chore: Use package, not build

toddbaert:
I was using "build" which happened to exist in OTel, but isn't actually the right target.


beeme1mr (approved):
Good catch!

PR #97, opened 2022-08-26
feat: add Unix socket support

beeme1mr:
This PR adds support for Unix sockets when using gRPC.

I did notice a few items that should be discussed and addressed in a future PR.

- [x] HTTP is used by default but it should likely be gRPC
- [x] Consider completely removing HTTP since it requires additional dependencies.
- [ ] Include environment variable support in flagd

Signed-off-by: Michael Beemer <beeme1mr@users.noreply.github.com>

### DIFF
--- a/libs/providers/flagd/src/lib/flagd-provider.spec.ts
+++ b/libs/providers/flagd/src/lib/flagd-provider.spec.ts
@@ -1,10 +1,9 @@
-import {
-  Client, OpenFeature, ErrorCode
-} from '@openfeature/nodejs-sdk';
-import type { UnaryCall } from "@protobuf-ts/runtime-rpc";
+import { Client, OpenFeature, ErrorCode } from '@openfeature/nodejs-sdk';
+import type { UnaryCall } from '@protobuf-ts/runtime-rpc';
 import { Struct } from '../proto/ts/google/protobuf/struct';
 import { ServiceClient } from '../proto/ts/schema/v1/schema.client';
-import { ResolveBooleanRequest,
+import {
+  ResolveBooleanRequest,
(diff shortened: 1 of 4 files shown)


james-milligan (approved):

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -13,15 +13,16 @@ import { Service } from '../Service';
...
 export class GRPCService implements Service {
   client: ServiceClient;
 
   constructor(options: GRPCServiceOptions) {
-    const { host, port } = options;
+    const { host, port, socketPath } = options;
     this.client = new ServiceClient(
       new GrpcTransport({
-        host: `${host}:${port}`,
+        host: socketPath ? `unix://${socketPath}` : `${host}:${port}`,
```
toddbaert, line 25: Can we add a test for this?

toddbaert (approved):

libs/providers/flagd/src/lib/flagd-provider.spec.ts
```diff
@@ -36,105 +36,179 @@ const OBJECT_KEY = 'object-flag';
...
+);
+
+jest.mock('@protobuf-ts/grpc-transport');
+
+import { GrpcTransport } from '@protobuf-ts/grpc-transport';
 
 describe(FlagdProvider.name, () => {
+  describe('GRPC Service Options', () => {
+    beforeEach(() => {
+      jest.clearAllMocks();
```
toddbaert, line 57: Maybe this time I'll remember which jest method to call. Probably not though.

PR #100, opened 2022-08-29
feat: add Composite provider

beeme1mr:
## Overview

The federated provider is based on the "super provider" concept that has been brought up in a number of community meetings. This PR adds an provider that supports an arbitrary number of child providers. Flags are evaluated based on the order providers are registered.

## Limitations

- All child provider hooks are run.

### DIFF
--- a/libs/providers/federated/src/lib/federated-provider.ts
+++ b/libs/providers/federated/src/lib/federated-provider.ts
@@ -1,0 +1,90 @@
+import {
+  EvaluationContext,
+  Provider,
+  ResolutionDetails,
+  Hook,
+} from '@openfeature/nodejs-sdk';
+
+export class FederatedProvider implements Provider {
+  metadata = {
+    name: FederatedProvider.name,
+  };
+
+  hooks: Hook[] = [];
+
+  constructor(private readonly providers: Provider[]) {
+    providers.forEach((p) => this.hooks.push(...(p.hooks ?? [])));
+  }
+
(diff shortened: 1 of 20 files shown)


libs/providers/federated/src/lib/federated-provider.ts
```diff
@@ -0,0 +1,90 @@
...
+    defaultValue: number,
+    context: EvaluationContext
+  ): Promise<ResolutionDetails<number>> {
+    for (const provider of this.providers) {
+      const result = await provider.resolveNumberEvaluation(
+        flagKey,
+        defaultValue,
+        context
+      );
+      if (!result.errorCode) {
```
justinabrahms, line 66: You may want to do some logging in the error cases here.

justinabrahms (approved):

justinabrahms:
optional nitpick: "Composite provider"? federated makes me think it's decentralized.

libs/providers/federated/src/lib/federated-provider.ts
beeme1mr: That's a good idea. I'll do that once [this PR](https://github.com/open-feature/node-sdk/pull/219) is merged.

beeme1mr:
@justinabrahms, the use of federation in this context was inspired by [GraphQL](https://www.apollographql.com/docs/federation/). Happy to change it to something else if it's confusing though.

FYI @toddbaert

PR #102, opened 2022-09-12
fix!: Fix object parsing, remove HTTP

toddbaert:
- fixes: https://github.com/open-feature/node-sdk-contrib/issues/101
- removes http service (BREAKING)
- removes axios peer dep in flagd provider
- adds GRPC testing

Closes: https://github.com/open-feature/node-sdk-contrib/issues/101


libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -3,45 +3,38 @@ import {
...
   metadata = {
-    name: 'flagD Provider',
+    name: 'flagd Provider',
   };
 
   private readonly service: Service;
 
-  constructor(options?: FlagdProviderOptions) {
-    const { service, host, port, protocol }: FlagdProviderOptions = {
+  constructor(options?: FlagdProviderOptions, service?: GRPCService) {
```
toddbaert, line 24: This is a non-breaking changes that allows the injection of a GRPCService instance. Few will probably use this, but it makes testing much easier.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -8,21 +9,24 @@ import {
...
+  protocol: Protocol
 }
+
 export class GRPCService implements Service {
   client: ServiceClient;
 
-  constructor(options: GRPCServiceOptions) {
-    const { host, port } = options;
-    this.client = new ServiceClient(
+  constructor(options: GRPCServiceOptions, client?: ServiceClient) {
```
toddbaert, line 24: Again allowing injection of `ServiceClient` for easy testing (we can inject a mock)

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -126,4 +130,9 @@ export class GRPCService implements Service {
       };
     }
   }
+
+  private convertContext(context: EvaluationContext): Struct {
+    // JsonValue closely matches EvaluationContext, this is a safe cast.
+    return Struct.fromJson(context as JsonValue);
+  }
```
toddbaert, line 137: Reduced some duped code here, also instead of stringifying, then converting to struct, we just convert directly to struct which should improve performance.

beeme1mr (approved):

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -110,10 +114,10 @@ export class GRPCService implements Service {
     try {
       const { response } = await this.client.resolveObject({
         flagKey,
-        context: Struct.fromJsonString(JSON.stringify(context)),
+        context: this.convertContext(context),
       });
       return {
-        value: response.value as T,
+        value: (response.value ? Struct.toJson(response.value) : undefined) as T,
```
toddbaert, line 120: This is the actual fix.

beeme1mr: Should we throw if value is undefined? I think an undefined value is invalid.

toddbaert: I wasn't sure about this. Why do you think an undefined value is invalid?

beeme1mr: The value is required but since it's a generic I suppose technically it can be undefined.

https://github.com/open-feature/node-sdk/blob/main/src/types.ts#L202

It just doesn't seem like the behavior I would expect. In what situation would an undefined value be useful? Is this something we had to worry about with the Java provider?

toddbaert: > The value is required but since it's a generic I suppose technically it can be undefined.

That's part of the problem. I think we need to refine the type to not be generic, but something more representative of structured data, like it is in the Java/dotnet SDK. In Java/dotnet, a node (`Value`) can be returned which represents a JSON `null`. I think we should refine the type to be more like the `Evaluation Context` type, which extends this:

`export type JSONValue = null | string | number | boolean | Date | { [x: string]: JSONValue } | Array<JSONValue>;`.

This is also more in-line with the other SDKs, where the context and the object resolver return type are similarly structured (though not exactly the same).

> It just doesn't seem like the behavior I would expect. In what situation would an undefined value be useful? Is this something we had to worry about with the Java provider?

In other SDKs we've structured things very flexibly, so anything like a JSON node can be returned (this also worked great for other data-interchange formats, see [GRPC's object type](https://github.com/timostamm/protobuf-ts/blob/3eea598471fb9c891c8b6da06bc1cbe18bc3457e/packages/runtime/src/json-typings.ts#L10)) so `null`ish values make sense. To your point, this isn't the same as undefined though. I'm not sure what would happen if the value returned in the GRPC message was completely missing in Java.

toddbaert: Throwing parse error now.

beeme1mr (approved):

james-milligan (approved):

PR #114, opened 2022-10-03
feat!: migrate to sdk 0.5.0

toddbaert:
- update providers to be compatible with 0.5.0 changes


libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
```diff
@@ -227,10 +227,12 @@ export class GoFeatureFlagProvider implements Provider {
...
       variant: apiResponseData.variationType,
-      reason: apiResponseData.reason.toString(),
+      reason: apiResponseData.reason?.toString() || 'UNKNOWN'
     };
-    if (apiResponseData.errorCode) {
-      sdkResponse.errorCode = apiResponseData.errorCode.toString();
+    if (Object.values(ErrorCode).includes(apiResponseData.errorCode as ErrorCode)) {
+      sdkResponse.errorCode = ErrorCode[apiResponseData.errorCode as ErrorCode];
+    } else if (apiResponseData.errorCode) {
+      sdkResponse.errorCode = ErrorCode.GENERAL;
```
toddbaert, line 235: This is probably not going to happen, but it looks like the existing code would return error codes that goff sent it the HTTP response.

Now, those errors code are an enum, so I have to do some mapping and if I can't map it, I use `GENERAL`.

libs/providers/flagd/package.json
```diff
@@ -7,6 +7,6 @@
     "current-version": "echo $npm_package_version"
   },
   "peerDependencies": {
-    "@openfeature/js-sdk": "^0.4.0"
+    "@openfeature/js-sdk": "~0.5.0"
```
toddbaert, line 10: require `0.5.x` since we are treating minors as majors sub v0.

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
thomaspoignant: Yes it should probably throw in this case let me look at that in another PR if you want.

thomaspoignant (approved):

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
toddbaert: Ya, I think it's a but unexpected, but this is about your REST API so it's up to you! to decide exactly how this should behave.

PR #179, opened 2022-12-09
feat!: update the otel hook to be spec compliant

### LINKED ISSUE #171: Add OTel hook
Now that [OTel 2529](https://github.com/open-telemetry/opentelemetry-specification/pull/2529) is merged, we should update the (or create a new) Open Telemetry hook.

Basically, this would entail creating a hook which:

- in the `after` stage: gets the current span and adds an appropriately named event (`feature_flag`), with the attributes (`feature_flag.key`, `feature_flag.provider_name`, and `feature_flag.variant`)
- in the `error` stage, records the exception

In pseudocode, this looks something like:

```
class OTelHook implements Hook {
  after(context, details) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.addEvent("feature_flag", { "feature_flag.key": details.key, /*  other props */ } );
  }

  error(err) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.recordException(err);
  }
}
```
See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md

Note: For the this repo, we may want to deprecate the old hook.

beeme1mr:
## Overview

🚨 This is a breaking change!🚨 

The OpenTelemetry semantic convention specification now includes a section on [feature flags](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/feature-flags/). The hook no longer creates a new span but attaches a span event to an existing span.

It also changes the attribute names in the following ways:

- `flag_key` => `key`
- evaluated_variant => variant
- `evaluated_value` is no longer used. A stringified version of the value is used if a variant is available. 

## This PR

- Updates the OTel hook to be spec [compliant](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/feature-flags/).
- Updates Jest tooling for the package
- Updates the package readme

### Related Issues

Fixes #171 

Signed-off-by: Michael Beemer <beeme1mr@users.noreply.github.com>


libs/hooks/open-telemetry/README.md
```diff
@@ -1,21 +1,59 @@
 # OpenTelemetry Hook
 
+The OpenTelemetry hook for OpenFeature provides a [spec compliant][otel-spec] way to automatically add a feature flag evaluation to a span as a span event. Since feature flags are dynamic and affect runtime behavior, it’s important to collect relevant feature flag telemetry signals. This can be used to determine the impact a feature has on a request, enabling enhanced observability use cases, such as A/B testing or progressive feature releases.
+
```
toddbaert, line 4: This is a good blurb.

libs/hooks/open-telemetry/README.md
```diff
@@ -1,21 +1,59 @@
...
+Confirm that the following peer dependencies are installed.
 
 ```
 $ npm install @openfeature/js-sdk @opentelemetry/api
 ```
 
-## Building
+## Usage
+
+OpenFeature provider various ways to register hooks. The location that a hook is registered affects when the hook is run. It's recommended to register the `OpenTelemetryHook` globally in most situations but it's possible to only enable the hook on specific clients. You should **never** register the `OpenTelemetryHook` globally and on a client.
```
toddbaert, line 21: We could probably implement a clever mechanism to detect if the Hook is run multiple times in an evaluation and warn, but it might not be worth it.

libs/hooks/open-telemetry/project.json
```diff
@@ -81,7 +81,9 @@
       ],
       "options": {
         "jestConfig": "libs/hooks/open-telemetry/jest.config.ts",
-        "passWithNoTests": true
+        "passWithNoTests": true,
+        "codeCoverage": true,
+        "coverageDirectory": "coverage/libs/hooks/open-telemetry"
```
toddbaert, line 86: I wonder if we should add this to all projects, and the template.

libs/hooks/open-telemetry/src/lib/open-telemetry-hook.ts
```diff
@@ -1,65 +1,37 @@
...
+  after(hookContext: HookContext, evaluationDetails: EvaluationDetails<FlagValue>) {
+    const currentTrace = trace.getActiveSpan();
+    if (currentTrace) {
+      let variant = evaluationDetails.variant;
+
+      if (!variant) {
+        if (typeof evaluationDetails.value === 'string') {
+          variant = evaluationDetails.value;
+        } else {
+          variant = JSON.stringify(evaluationDetails.value);
```
toddbaert, line 21: This will break with circular structures.

libs/hooks/open-telemetry/project.json
beeme1mr: Yes, probably but I didn't want to pollute this PR.

libs/hooks/open-telemetry/src/lib/open-telemetry-hook.ts
toddbaert, line 21: You could refactor this into 2 ternaries, which would allow you to make `variant` a const, also... but some people hate nested ternaries.

toddbaert (approved):
Approved with comments, please especially consider: https://github.com/open-feature/js-sdk-contrib/pull/179#discussion_r1044692735

libs/hooks/open-telemetry/README.md
beeme1mr: I don't think it's worth it since it's a misconfiguration if you register any plugin at multiple layers.

libs/hooks/open-telemetry/project.json
beeme1mr: A follow up issue has been created.

libs/hooks/open-telemetry/src/lib/open-telemetry-hook.ts
beeme1mr: Yeah, I had a nested ternary but that seemed a bit nasty too. Can you provide a code snippet of what you had in mind?

beeme1mr: Can the evaluation details value ever be a circular structure?

toddbaert: It's JS so at the end of the day, providers can return whatever they want if they want to cast unsafely.

toddbaert: Something like:

```typescript
const variant = (evaluationDetails.variant) ?
        evaluationDetails.variant : (typeof evaluationDetails.value === 'string') ? 
        evaluationDetails.value : JSON.stringify(evaluationDetails.value);
```

toddbaert: Some of the braces here aren't needed, but I find they make nested ternaries easier to read.

Again, some people HATE them though.

beeme1mr: I'm just going to leave it. Prettier makes your proposal look nasty, unfortunately. Also, SonarLint is also complaining about nested ternaries.

tcarrio: If a circular reference is being provided, they can specify how it should be serialized with the `toJSON` method.

> If the value has a toJSON() method, it's responsible to define what data will be serialized. Instead of the object being serialized, the value returned by the toJSON() method when called will be serialized. JSON.stringify() calls toJSON with one parameter, the key

[Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)

If we are concerned about this not being straightforward, the actual interface could be updated to hint at the optional `toJSON` property.

toddbaert: I'm not too concerned about it TBH, I just thought it was worth pointing out.

tcarrio: Since this is directly on EvaluationDetails it only impacts providers.

I don't see this as likely but it's possible. Wouldn't hurt to be ahead of it.

toddbaert: Thinking about it more, I think you should ignore this possibility.

Issue #198, opened 2022-12-25
[flagd-provider] Can't import flagd provider in ESM Typescript project


talzion12:
Trying to import FlagdProvider in Typescript with ESM causes the following error:
```
node_modules/@openfeature/flagd-provider/src/index.d.ts:1:15 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

1 export * from './lib/flagd-provider';
                ~~~~~~~~~~~~~~~~~~~~~~

src/main.ts:2:10 - error TS2305: Module '"@openfeature/flagd-provider"' has no exported member 'FlagdProvider'.

2 import { FlagdProvider } from '@openfeature/flagd-provider';
           ~~~~~~~~~~~~~


Found 2 errors in 2 files.

Errors  Files
     1  node_modules/@openfeature/flagd-provider/src/index.d.ts:1
     1  src/main.ts:2
```

When I switch my project to CommonJS the problem is resolved.
I'm not sure if this is a bug or if I'm doing something wrong here.

### Steps to reproduce

* Create Typescript ESM project.
* Import FlagdProvider like so: `import { FlagdProvider } from '@openfeature/flagd-provider';`.

I also created a simple [example repo](https://github.com/talzion12/openfeature-example). Running `npm run build` shows the error.

beeme1mr:
Hi @talzion12, thanks for reporting this. Sorry you keep running into ESM issues! The team has limited availability this week but we'll address it as soon as possible.

talzion12:
@beeme1mr Haha yes ESM is causing me quite a bit of issues 😅
No problem, in the meantime I created a custom provider.
Thanks!

toddbaert:
I'll look into this today and hopefully publish a fix.

toddbaert:
Thanks again for the report @talzion12 . Looks like this might have been a [change/regression in NX](https://github.com/open-feature/js-sdk-contrib/pull/201).

I published `0.7.3` which resolves the issue in your example project.

toddbaert:
I'm still seeing issues with this. I have a fix I'll push tomorrow.

PR #201, opened 2022-12-29
fix: fix ESM and web polyfills issue

toddbaert:
Fixes: https://github.com/open-feature/js-sdk-contrib/issues/198

This _seems_ to be the result of a breaking change in the nx `@nrwl/web:rollup` executor, which previously added an exports statement. I have moved to `@nrwl/rollup:rollup` with the `tsc` compiler, which generates the exports correctly and also fixes some [inappropriate polyfills](https://github.com/nrwl/nx/issues/10304) the `@nrwl/web:rollup` executor was adding.

### DIFF
--- a/libs/hooks/open-telemetry/project.json
+++ b/libs/hooks/open-telemetry/project.json
@@ -4,7 +4,7 @@
   "projectType": "library",
   "targets": {
     "package": {
-      "executor": "@nrwl/web:rollup",
+      "executor": "@nrwl/rollup:rollup",
       "outputs": [
         "{options.outputPath}"
       ],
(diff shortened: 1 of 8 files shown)


libs/hooks/open-telemetry/project.json
```diff
@@ -13,7 +13,9 @@
         "outputPath": "dist/libs/hooks/open-telemetry",
         "entryFile": "libs/hooks/open-telemetry/src/index.ts",
         "tsConfig": "libs/hooks/open-telemetry/tsconfig.lib.json",
-        "compiler": "babel",
+        "compiler": "tsc",
+        "skipTypeField": true,
```
toddbaert, line 17: This removes the generated `types` field in package JSON.

libs/hooks/open-telemetry/project.json
```diff
@@ -13,7 +13,9 @@
...
+        "generateExportsField": true,
```
toddbaert, line 18: explicitly generate the exports field.

beeme1mr (approved):

PR #212, opened 2023-01-18
fix: module issues with types

toddbaert:
Fixes an issue with imports in ES module projects. Specifically, in some TS configurations, compilation would fail because TS esm support requires file extensions on all imports. Though our code is bundled, our declaration files are not, and they are also subject to this requirement. To avoid having to add `.js` extensions in our rendered type files, this adds a package.json with `type: commonjs` only in the type file roots, making them behave as commonjs modules, (which don't require file extensions in imports).

I attempted first to bundle all our types into a single d.ts (to avoid imports entirely), but was unable to find a decent solution to that.

Fixes: https://github.com/open-feature/js-sdk-contrib/issues/198

I've tested this fix with the example repo in the linked issue, as well as another es-module project, and a commonjs project. Everything works as expected.

### DIFF
--- a/tools/generators/open-feature/index.ts
+++ b/tools/generators/open-feature/index.ts
@@ -135,7 +135,7 @@ function updateProject(tree: Tree, projectRoot: string, umdName: string) {
         format: ['cjs', 'esm'],
         assets: [
           // Move a "commonjs" package.json to the types root (js is bundled).
-          // This prevents is from having to add file extensions to all our imports in ESM contexts, which ESM requires.
+          // This prevents us from having to add file extensions to all our imports in ESM contexts, which ESM requires.
           {
             glob: "package.json",
             input: "./assets",
--- a/tools/generators/open-feature/index.ts
+++ b/tools/generators/open-feature/index.ts
@@ -129,12 +129,16 @@ function updateProject(tree: Tree, projectRoot: string, umdName: string) {
         tsConfig: `${projectRoot}/tsconfig.lib.json`,
         buildableProjectDepsInPackageJsonType: 'dependencies',
         compiler: 'tsc',
-        skipTypeField: true,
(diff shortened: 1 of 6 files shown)


tools/generators/open-feature/index.ts
```diff
@@ -129,12 +129,16 @@ function updateProject(tree: Tree, projectRoot: string, umdName: string) {
         tsConfig: `${projectRoot}/tsconfig.lib.json`,
         buildableProjectDepsInPackageJsonType: 'dependencies',
         compiler: 'tsc',
-        skipTypeField: true,
         generateExportsField: true,
         umdName,
         external: ['typescript'],
         format: ['cjs', 'esm'],
         assets: [
+          {
```
toddbaert, line 137:
```suggestion
          // Move a "commonjs" package.json to the types root (js is bundled).
          // This prevents is from having to add file extensions to all our imports in ESM contexts, which ESM requires.
          {
```

tcarrio (approved):

tools/generators/open-feature/index.ts
```diff
@@ -129,12 +129,18 @@ function updateProject(tree: Tree, projectRoot: string, umdName: string) {
...
         buildableProjectDepsInPackageJsonType: 'dependencies',
         compiler: 'tsc',
-        skipTypeField: true,
         generateExportsField: true,
         umdName,
         external: ['typescript'],
         format: ['cjs', 'esm'],
         assets: [
+          // Move a "commonjs" package.json to the types root (js is bundled).
+          // This prevents is from having to add file extensions to all our imports in ESM contexts, which ESM requires.
```
toddbaert, line 138:
```suggestion
          // This prevents us from having to add file extensions to all our imports in ESM contexts, which ESM requires.
```

PR #218, opened 2023-01-23
feat: add standard flagd caching

toddbaert:
This PR:

- refactors the underlying gRPC service to use generics to avoid lots of repetitive caching code that would otherwise be duplicated in multiple methods
- adds standard flagd caching
- add standard caching related config options / env vars
- adds stream reconnect logic
- tests for the above


libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -2,111 +2,231 @@ import * as grpc from '@grpc/grpc-js';
...
-              ? grpc.credentials.createSsl()
-              : grpc.credentials.createInsecure(),
+            channelCredentials: tls ? grpc.credentials.createSsl() : grpc.credentials.createInsecure(),
           })
         );
+
+    if (config.cache === 'lru') {
+      this._cacheEnabled = true;
+      this._cache = new LRU({ maxSize: config.maxCacheSize || DEFAULT_MAX_CACHE_SIZE, sizeCalculation: () => 1 });
+      this.streamConnection = this.connectStream();
```
toddbaert, line 99: For now, we only need streaming if the cache is enabled (will need to be pulled out once we support events.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -2,111 +2,231 @@ import * as grpc from '@grpc/grpc-js';
...
+  private _streamConnectAttempt = 0;
+  private _streamConnectBackoff = BASE_EVENT_STREAM_RETRY_BACKOFF_MS;
+  private _maxEventStreamRetries;
+  private get _cacheActive() {
+    // the cache is "active" (able to be used) if the config enabled it, AND the gRPC stream is live
+    return this._cacheEnabled && this._streamAlive;
+  }
 
-  constructor(config: Config, client?: ServiceClient) {
+  readonly streamConnection = Promise.resolve(false);
```
toddbaert, line 82: @skyerus this is the equivalent "blocking" method as the one you implemented in Java.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -2,111 +2,231 @@ import * as grpc from '@grpc/grpc-js';
...
-  ): Promise<ResolutionDetails<number>> {
-    const { response } = await this.client.resolveFloat({
-      flagKey,
-      context: this.convertContext(context, logger),
-    }).then(this.onFulfilled, this.onRejected);
-    return {
-      value: response.value,
-      reason: response.reason,
-      variant: response.variant,
-    };
```
toddbaert, line 89: This logic has been made generic and extracted.

package.json
```diff
@@ -13,10 +13,11 @@
   "dependencies": {
     "@bufbuild/connect-web": "^0.6.0",
     "@grpc/grpc-js": "^1.7.3",
-    "@openfeature/js-sdk": "~1.0.0",
+    "@openfeature/js-sdk": "^1.1.0",
     "@opentelemetry/api": "^1.3.0",
     "@protobuf-ts/grpc-transport": "2.8.2",
     "axios": "1.2.3",
+    "lru-cache": "^7.14.1",
```
toddbaert, line 20: @beeme1mr I've verified this is properly added to the deps of the `package.json` in the flagd dist.

package.json
```diff
@@ -13,10 +13,11 @@
   "dependencies": {
     "@bufbuild/connect-web": "^0.6.0",
     "@grpc/grpc-js": "^1.7.3",
-    "@openfeature/js-sdk": "~1.0.0",
+    "@openfeature/js-sdk": "^1.1.0",
```
toddbaert, line 16: required 1.1 because we need `STATIC` and `CACHED` values.

libs/providers/flagd/src/lib/flagd-provider.spec.ts
```diff
@@ -1,11 +1,12 @@
 jest.mock('@protobuf-ts/grpc-transport');
 
-import { Client, ErrorCode, EvaluationContext, OpenFeature } from '@openfeature/js-sdk';
+import { Client, ErrorCode, EvaluationContext, JsonObject, OpenFeature, StandardResolutionReasons } from '@openfeature/js-sdk';
```
beeme1mr, line 3: Can you make sure this is formatted with Prettier?

libs/providers/flagd/src/lib/service/grpc/service.ts
beeme1mr: Can you add a comment explaining why you're calling `Promise.resolve(false)`? The context will likely be valuable to future contributors.

libs/providers/flagd/src/lib/service/grpc/service.ts
beeme1mr: Can you add this as a comment in the code?

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -2,111 +2,238 @@ import * as grpc from '@grpc/grpc-js';
...
-        value: Struct.toJson(response.value) as T,
-        reason: response.reason,
-        variant: response.variant,
-      };
+    return this.resolve(this._client.resolveObject, flagKey, context, logger, this.objectParser);
+  }
+
+  private connectStream(): Promise<boolean> {
+    return new Promise((resolve, reject) => {
+      this.logger?.info(`${FlagdProvider.name}: connecting stream, attempt ${this._streamConnectAttempt}...`);
```
beeme1mr, line 130: Should this be info level?

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -2,111 +2,238 @@ import * as grpc from '@grpc/grpc-js';
...
+    this.logger?.info(`${FlagdProvider.name}: streaming connection established with flagd`);
+    this._streamAlive = true;
+    this._streamConnectAttempt = 0;
+    this._streamConnectBackoff = BASE_EVENT_STREAM_RETRY_BACKOFF_MS;
+    resolve(true);
+  }
+
+  private handleFlagsChanged(message: EventStreamResponse) {
+    if (message.data) {
+      const parsedMessage = Struct.toJson(message.data) as FlagChangeMessage;
```
beeme1mr, line 158: Is this a safe case? Should there be a type guard?

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -141,5 +263,5 @@ export class GRPCService implements Service {
       default:
         throw new GeneralError(err.message);
     }
-  }
+  };
```
beeme1mr, line 266: Is a semicolon necessary?

beeme1mr (approved):
Overall, this looks great! I added a few comments and questions.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: Good idea, will do.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: Yep, good idea.

libs/providers/flagd/src/lib/configuration.ts
```diff
@@ -53,6 +90,15 @@ const getEnvVarConfig = (): Partial<Config> => ({
   ...(process.env[ENV_VAR.FLAGD_SOCKET_PATH] && {
     socketPath: process.env[ENV_VAR.FLAGD_SOCKET_PATH],
   }),
+  ...(process.env[ENV_VAR.FLAGD_CACHE] && {
+    cache: process.env[ENV_VAR.FLAGD_CACHE] as CacheOption,
```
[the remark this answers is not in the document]
toddbaert: yes.

libs/providers/flagd/src/lib/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const BASE_EVENT_STREAM_RETRY_BACKOFF_MS = 1000;
+export const DEFAULT_MAX_EVENT_STREAM_RETRIES = 5;
+export const EVENT_CONFIGURATION_CHANGE = 'configuration_change';
+export const EVENT_PROVIDER_READY = 'provider_ready';
```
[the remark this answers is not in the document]
toddbaert: They will eventually, and are there in the experimental web SDK for example. For now, all these events are used entirely within the provider so they can just be here.

libs/providers/flagd/src/lib/flagd-provider.spec.ts
toddbaert: I will check that.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: I debated this - I though yes since it's something that would happen infrequently. I could be convinced otherwise.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: I doubt it.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: We have all the null checks we need later, but this does at least _look_ suspect. I will improve it.

libs/providers/flagd/src/lib/configuration.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { FlagdProviderOptions, getConfig } from './configuration';
+import { DEFAULT_MAX_CACHE_SIZE, FlagdProviderOptions, getConfig } from './configuration';
```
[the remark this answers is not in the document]
toddbaert: I think so.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -2,111 +2,238 @@ import * as grpc from '@grpc/grpc-js';
...
+    const resolved = {
+      // invoke the parser method if passed
+      value: parser ? parser.call(this, response.value as Struct) : response.value,
+      reason: response.reason,
+      variant: response.variant,
+    } as ResolutionDetails<T>;
+
+    if (this._cacheActive && response.reason === StandardResolutionReasons.STATIC) {
+      // cache this static value
+      this._cache?.set(flagKey, { ...resolved, reason: StandardResolutionReasons.CACHED });
```
skyerus, line 234: This differs from the other implementations in that it sets the reason as CACHED when storing rather than setting it on the retrieved value from cache. This may be fine but thought it was worth pointing out

toddbaert: Interesting thought. I will think about this a bit.

beeme1mr: It may be slightly better to set `CACHED` after the value is retrieved. However, I can't think of a scenario where the behavior would be different between the two different implementations.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: I got rid of a lot of `info`s, I think you're right.

I only kept the connection establishment to info.

libs/providers/flagd/src/lib/configuration.spec.ts
```diff
@@ -21,17 +25,26 @@ describe('Configuration', () => {
...
+    const maxEventStreamRetries = 10;
+    const cache = 'other';
 
     process.env['FLAGD_HOST'] = host;
     process.env['FLAGD_PORT'] = `${port}`;
     process.env['FLAGD_TLS'] = `${tls}`;
     process.env['FLAGD_SOCKET_PATH'] = socketPath;
+    process.env['FLAGD_CACHE'] = cache;
+    process.env['FLAGD_MAX_CACHE_SIZE'] = `${maxCacheSize}`;
+    process.env['FLAGD_MAX_EVENT_STREAM_RETRIES'] = `${maxEventStreamRetries}`;
```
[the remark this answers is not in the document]
toddbaert: It looks like we're already doing that.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: fixed.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: updated for consistency.

libs/providers/flagd/src/lib/configuration.spec.ts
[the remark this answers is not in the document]
beeme1mr: Yep, you're right. I guess we're good.

package.json
```diff
@@ -13,10 +13,10 @@
   "dependencies": {
     "@bufbuild/connect-web": "^0.6.0",
     "@grpc/grpc-js": "^1.7.3",
-    "@openfeature/js-sdk": "~1.0.0",
```
toddbaert, line 16: This doesn't really make sense here.

Having it seems to be no problem, since the package.jsons all include this is a peer dep with whatever version they need, but I wanted to take it out of the prod deps.

libs/providers/flagd/src/lib/configuration.ts
```diff
@@ -53,6 +88,15 @@ const getEnvVarConfig = (): Partial<Config> => ({
   ...(process.env[ENV_VAR.FLAGD_SOCKET_PATH] && {
     socketPath: process.env[ENV_VAR.FLAGD_SOCKET_PATH],
   }),
+  ...(process.env[ENV_VAR.FLAGD_CACHE] && (process.env[ENV_VAR.FLAGD_CACHE] === 'lru' || process.env[ENV_VAR.FLAGD_CACHE] === 'disabled') && {
```
beeme1mr, line 91: Do you need this part? `process.env[ENV_VAR.FLAGD_CACHE] &&`

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -2,111 +2,246 @@ import * as grpc from '@grpc/grpc-js';
...
-              ? grpc.credentials.createSsl()
-              : grpc.credentials.createInsecure(),
+            channelCredentials: tls ? grpc.credentials.createSsl() : grpc.credentials.createInsecure(),
           })
         );
+
+    // for now, we only need streaming if the cache is enabled (will need to be pulled out once we support events)
+    if (config.cache === 'lru') {
+      this._cacheEnabled = true;
+      this._cache = new LRU({ maxSize: config.maxCacheSize || DEFAULT_MAX_CACHE_SIZE, sizeCalculation: () => 1 });
```
beeme1mr, line 107: It looks like if a user sets the `maxCacheSize` to zero it will use the `DEFAULT_MAX_CACHE_SIZE`. That might be slightly confusing. Perhaps we should disable the cache in the situation. What do you think?

toddbaert: I think we have a clear way to disable cache with `'disabled'`.

libs/providers/flagd/src/lib/configuration.ts
toddbaert: no

libs/providers/flagd/src/lib/service/grpc/service.ts
beeme1mr: Yeah, that's fair. I'm fine with this then.

libs/providers/flagd/src/lib/configuration.ts
toddbaert: fixed

PR #238, opened 2023-02-24
Feat/flagd web

toddbaert:


libs/providers/flagd-web/src/lib/flagd-web-provider.ts
```diff
@@ -0,0 +1,194 @@
...
+        this._logger?.error(`${FlagdWebProvider.name}: could not establish connection to flagd`);
+        setTimeout(() => this.retryConnect(context), this._delayMs);
+      }
+    );
+    this._retry++;
+  }
+
+  private async fetchAll(context: EvaluationContext) {
+    const transformedContext = this.transformContext(context);
+    const allResolved = await this._promiseClient.resolveAll({ context: transformedContext });
```
beeme1mr, line 165: Should anything happen if the call to `resolveAll` fails? Perhaps we should retry a few times? Is there a timeout?

beeme1mr: Should the existing flags be marked as stale until flagd responds with the updated values?

libs/providers/flagd-web/src/lib/options.ts
```diff
@@ -0,0 +1,53 @@
+export interface Options {
+  /**
+   * The domain name or IP address of flagd.
+   *
+   * @default window
+   */
+  host: string;
+
+  /**
+   * The port flagd is listen on.
```
beeme1mr, line 10:
```suggestion
   * The port flagd is listening on.
```

libs/providers/flagd-web/src/lib/options.ts
```diff
@@ -0,0 +1,53 @@
+export interface Options {
```
beeme1mr, line 1: What about the path? We may need to support at least a path prefix.

libs/providers/flagd-web/src/lib/options.ts
```diff
@@ -0,0 +1,53 @@
...
+   */
+  maxDelay: number;
+
+  /**
+   * Sets the maximum number of retries for a connection to be made to the flagd instance
+   * 0 means unlimited.
+   *
+   * @default 0
+   */
+  maxRetries: number;
```
beeme1mr, line 36: Can retries be disabled? Perhaps by using something like `-1`?

package.json
```diff
@@ -12,13 +12,14 @@
   "private": true,
   "dependencies": {
```
beeme1mr, line 13: Are you using `@bufbuild/connect`? It's mentioned as a runtime dependency in the [Buf docs](https://connect.build/docs/web/generating-code#local-generation).

package.json
```diff
@@ -12,13 +12,14 @@
...
     "@bufbuild/connect-web": "^0.6.0",
```
beeme1mr, line 14: Is looks like this is now called `@bufbuild/connect-es`.
https://github.com/bufbuild/connect-es/releases/tag/v0.8.0

beeme1mr: Please review the release notes because there are a few breaking changes documented there.

libs/providers/flagd-web/README.md
```diff
@@ -0,0 +1,73 @@
...
+![Experimental](https://img.shields.io/badge/experimental-breaking%20changes%20allowed-yellow)
+
+A feature flag daemon with a Unix philosophy.
+
+## Installation
+
+:warning: This provider requires the use of the experimental @openfeature/web-sdk:
+
+```
+npm i @openfeature/web-sdk
```
skyerus, line 12:
```suggestion
npm install @openfeature/web-sdk
```

libs/providers/flagd-web/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The `FlagdWebProvider` communicates with flagd via the [connect protocol](https://buf.build/blog/connect-a-better-grpc).
+
+### Reconnection
+
+If the connection to the flagd instance fails, the provider will attempt to reconnect with an exponential back-off. The `maxDelay` and `maxRetries` can be specified to customize reconnect behavior.
+
+### Event streaming
+
+The `FlagdWebProvider` receives events from flag with changes. Combined with the event API in the web SDK, this allows for subscription to flag value changes in clients.
```
skyerus, line 29:
```suggestion
The `FlagdWebProvider` receives events from flagd with changes. Combined with the event API in the web SDK, this allows for subscription to flag value changes in clients.
```

package.json
```diff
@@ -12,13 +12,14 @@
...
+    "@bufbuild/protobuf": "^1.0.0",
     "@grpc/grpc-js": "^1.7.3",
+    "@openfeature/web-sdk": "^0.1.0-experimental",
```
skyerus, line 17: I'm confused, the web-sdk repo is empty, where is this coming from?

toddbaert: Sorry about that, I initially created a repo for it, but ended up instead converting the `js-sdk` to a monorepo because a lot of the code/interfaces were similar.

libs/providers/flagd-web/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@openfeature/flagd-web-provider",
+  "version": "0.1.33",
+  "type": "commonjs",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/web-sdk": "*"
```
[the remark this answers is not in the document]
toddbaert: Yes, the problem is that `0.1.0-xxx` is a very difficult thing to define minimums around.

package.json
toddbaert: We don't directly import `@bufbuild/connect` anywhere. If this is a dep, it's a transitive dep of `@bufbuild/connect-web`. I don't think we need to worry about it.

libs/providers/flagd-web/src/lib/options.ts
toddbaert: -1 works to disable retries. I've added documentation and a test for it.

libs/providers/flagd-web/src/lib/options.ts
toddbaert: I will look into this.

libs/providers/flagd-web/README.md
toddbaert: Always a good call.

libs/providers/flagd-web/src/lib/flagd-web-provider.ts
```diff
@@ -0,0 +1,194 @@
...
+    const transport = createConnectTransport({
+      baseUrl: `${tls ? 'https' : 'http'}://${host}:${port}`,
+    });
+    this._promiseClient = promiseClient ? promiseClient : createPromiseClient(Service, transport);
+    this._callbackClient = callbackClient ? callbackClient : createCallbackClient(Service, transport);
+    this._maxRetries = maxRetries === 0 ? Infinity: maxRetries;
+    this._maxDelay = maxDelay;
+    this._logger = logger;
+
+    // TODO: when web-sdk has initialization, put this there.
```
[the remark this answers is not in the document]
toddbaert: Not as part of this PR, we need change to the web SDK.

package.json
toddbaert: @beeme1mr `https://www.npmjs.com/package/@bufbuild/connect-es` doesn't exist. I think the resources you are linking are just about protoc generation. As the link you supplied says:

>  In the future, @bufbuild/connect-web **will only export the Connect and gRPC-web transports.**

That's what we're concern about here - the exported transports.

libs/providers/flagd-web/src/lib/options.ts
toddbaert: I've added `pathPrefix` and tested it!

libs/providers/flagd-web/src/lib/options.ts
```diff
@@ -0,0 +1,61 @@
...
+   * The path at which the flagd gRPC service is available, for example: /flagd-api (optional).
+   * 
+   * @default ""
+   */
+  pathPrefix: string;
+
+  /**
+   * Determines if TLS should be used.
+   *
+   * @default false
```
beeme1mr, line 26: TLS defaults to true

libs/providers/flagd-web/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The `FlagdWebProvider` communicates with flagd via the [connect protocol](https://buf.build/blog/connect-a-better-grpc).
+
+### Available options
+
+| Option name | Type    | Default   | Description                                                                                                                                            |
+| ----------- | ------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| host        | string  | localhost | sets the host used to connect to the flagd instance                                                                                                    |
+| port        | number  | 8013      | sets the port used to connect to the flagd instance                                                                                                    |
+| tls         | boolean | false     | when set to true the provider will attempt to connect to flagd via https                                                                               |
```
beeme1mr, line 29: Defaults to true

libs/providers/flagd-web/README.md
```diff
@@ -0,0 +1,73 @@
...
+| maxRetries  | number  | 0         | sets the maximum number of retries for a connection to be made to the flagd instance, a value of 0 means unlimited. A negative value means no retries. |
```
beeme1mr, line 30: Sets the maximum number of retries for a connection to be made to the flagd instance 0 means unlimited. A negative number means no retries.

libs/providers/flagd-web/README.md
```diff
@@ -0,0 +1,73 @@
...
+| maxDelay    | number  | 60000     | sets the maximum time in ms to wait between reconnect intervals                                                                                        |
```
beeme1mr, line 31: Add pathPrefix please

libs/providers/flagd-web/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+```typescript
+client.addHandler(ProviderEvents.Ready, () => {
+  // do something when the configuration has changed.
+});
+```
+
+### Caching
+
+The `FlagdWebProvider` evaluates flags in bulk, taking into account the evaluation context, and then caches them for local evaluation. The cache is invalidated when flag configuration change events are received.
```
beeme1mr, line 49: Maybe explicitly say it's cached in memory.

libs/providers/flagd-web/src/lib/options.ts
```diff
@@ -0,0 +1,61 @@
...
+  maxRetries: number;
+}
+
+export type FlagdProviderOptions = Partial<Options> & Pick<Options, 'host'>;
+
+export const DEFAULT_MAX_DELAY = 60000;
+
+export function getOptions(options: FlagdProviderOptions): Options {
+  return {
+    ...{
```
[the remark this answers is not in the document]
toddbaert: It's not, it has no default, it's required.

beeme1mr (approved):

PR #252, opened 2023-03-07
feat: add in memory provider

moredip:
adds a super simple in-memory provider, for use in automated tests, and for demos.

### DIFF
--- a/libs/providers/in-memory/src/lib/in-memory-provider.ts
+++ b/libs/providers/in-memory/src/lib/in-memory-provider.ts
@@ -18,11 +18,11 @@ export class InMemoryProvider implements Provider {
   private _flagConfiguration: FlagConfiguration;
 
   constructor(flagConfiguration: FlagConfiguration = {}) {
-    this._flagConfiguration = flagConfiguration;
+    this._flagConfiguration = { ...flagConfiguration };
   }
 
   replaceConfiguration(flagConfiguration: FlagConfiguration) {
-    this._flagConfiguration = flagConfiguration;
+    this._flagConfiguration = { ...flagConfiguration };
   }
 
   async resolveBooleanEvaluation(flagKey: string): Promise<ResolutionDetails<boolean>> {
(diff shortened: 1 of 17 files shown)


beeme1mr (approved):
Looks like a good start to me. Thanks for putting it together.

libs/providers/in-memory/src/lib/in-memory-provider.ts
```diff
@@ -0,0 +1,75 @@
...
+import { FlagConfiguration } from './flag-configuration';
+
+export class InMemoryProvider implements Provider {
+  readonly metadata = {
+    name: 'In-Memory Provider',
+  } as const;
+  private _flagConfiguration: FlagConfiguration;
+
+  constructor(flagConfiguration: FlagConfiguration = {}) {
+    this._flagConfiguration = flagConfiguration;
```
toddbaert, line 21: It really depends on how you want this to work, but JS is pass-by-ref for objects, meaning that since you're not cloning this in any way, the flag definitions can certainly be changed at runtime by changing the value of the object pass here. This might be surprising given what's in the readme.

Here's a branch off this one demoing what I mean: https://github.com/moredip/js-sdk-contrib/pull/1/files

toddbaert (approved):
Approved, but I recommend you reconcile this behavior with the doc or specify what happening here: https://github.com/open-feature/js-sdk-contrib/pull/252/files#r1129595932

If you want to get really fancy, you could even make the cloning an option.

libs/providers/in-memory/src/lib/in-memory-provider.ts
moredip: yeah, the `replaceConfiguration(...)` method is a defined way for a client to do that type of runtime change (intended for testing purposes). 

You're right that the README is a bit misleading here. You're also right that it would be a bit safer if we were to clone the config, but TBH I'm not sure it's that likely to cause issues. 

I'm happy to leave this as it is for now, and if someone wants to fix it later we can do so!

moredip: OK fine I couldn't resist fixing it 🤣 I updated the README and added cloning of the configuration/

moredip:
@toddbaert @beeme1mr I think this is ready to merge, and I think one of you needs to do the honors?

Once it's merged will the package automagically appear in the npm registry?

beeme1mr:
Hey @moredip, your provider should be available via npm.
`@openfeature/in-memory-provider`

PR #280, opened 2023-03-22
fix: reconnect, tests

toddbaert:
Fixes a regression in reconnecting.


libs/providers/flagd-web/src/lib/flagd-web-provider.ts
```diff
@@ -129,13 +129,13 @@ export class FlagdWebProvider implements Provider {
           }
         },
         () => {
-          reject();
```
toddbaert, line 132: @beeme1mr This was the main problem. I added this reject unconditionally no matter what retry we were on. This was the main issue.

Instead, I've moved an error event down below, only after we've exhausted all the retries.

toddbaert: I've also tested this in the browser and it works as expected.

beeme1mr (approved):

PR #282, opened 2023-03-23
fix: issue where context updates somtimes missed

toddbaert:
Always use the most up-to-date context, not one cached by the function closure.

### DIFF
--- a/libs/providers/flagd-web/src/lib/flagd-web-provider.spec.ts
+++ b/libs/providers/flagd-web/src/lib/flagd-web-provider.spec.ts
@@ -1,6 +1,6 @@
 import { CallbackClient, Code, ConnectError, PromiseClient } from '@bufbuild/connect-web';
 import { Struct } from '@bufbuild/protobuf';
-import { Client, ErrorCode, OpenFeature, ProviderEvents, StandardResolutionReasons } from '@openfeature/web-sdk';
+import { Client, ErrorCode, JsonValue, OpenFeature, ProviderEvents, StandardResolutionReasons } from '@openfeature/web-sdk';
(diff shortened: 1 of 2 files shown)


libs/providers/flagd-web/src/lib/flagd-web-provider.spec.ts
```diff
@@ -207,6 +208,23 @@ describe(FlagdWebProvider.name, () => {
...
+          } catch(err) {
+            done(err);
+          }
+        });
+        OpenFeature.setContext(context).then(() => {
+          mockCallbackClient.mockMessage({
+            type: EVENT_CONFIGURATION_CHANGE,
+          });
+        });
+      });
```
toddbaert, line 227: @beeme1mr this test fails if my fix is reverted (the context is not up-to-date).

libs/providers/flagd-web/src/lib/flagd-web-provider.ts
```diff
@@ -105,23 +106,25 @@ export class FlagdWebProvider implements Provider {
...
     return new Promise<void>((resolve) => {
       this._callbackClient.eventStream(
         {},
         (message) => {
+          // get the context at the time of the message
+          const currentContext = OpenFeature.getContext();
           this._logger?.debug(`${FlagdWebProvider.name}: event received: ${message.type}`);
           switch (message.type) {
             case EVENT_PROVIDER_READY:
-              this.fetchAll(context).then(() => {
```
toddbaert, line 118: This context was basically cached in this closure, which would cause `resolveAll` to resolve with an out-of-date context in some cases.

beeme1mr (approved):

PR #319, opened 2023-04-08
fix: go-feature-flag provider this.apiKey is undefined

thomaspoignant:
## This PR
I am struggling here and adding more bugs at every commit 🤦 
This one should be the good one 🤞

I was using `this.apiKey` instead of `options.apiKey`.


beeme1mr (approved):

toddbaert:
It's always easy to miss little things like this. One thing I do to manually tests these packages is to use [yalc](https://github.com/wclr/yalc).

You can cd into the dist directory and publish/consume your package locally.

thomaspoignant:
Oh thanks for sharing this tool this definitely something that can help me in the future.

PR #325, opened 2023-04-12
fix: add shutdown to flagd-web provider

toddbaert:
This PR adds a close handler to the Flagd Web Provider. This fixes some issues I discovered while working on adding e2e tests with @mowies (Jest complained about unclosed handles, which were indeed leaking because we never closed the connection).

### DIFF
--- a/libs/providers/flagd-web/src/lib/flagd-web-provider.ts
+++ b/libs/providers/flagd-web/src/lib/flagd-web-provider.ts
@@ -88,7 +88,7 @@ export class FlagdWebProvider implements Provider {
   }
 
   onClose(): Promise<void> {
-    // close the steam using the saved cancel function
+    // close the stream using the saved cancel function
     return Promise.resolve(this._cancelFunction?.());
   }
 


libs/providers/flagd-web/src/lib/flagd-web-provider.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { CallbackClient, Code, ConnectError, PromiseClient } from '@bufbuild/connect-web';
```
toddbaert, line 1: These imports are deprecated, the inline comments suggested moving to `@bufbuild/connect`. I tested this locally with the playground and it caused no issues.

libs/providers/flagd-web/src/lib/flagd-web-provider.spec.ts
```diff
@@ -239,6 +244,34 @@ describe(FlagdWebProvider.name, () => {
...
+        )
+      );
+      client = OpenFeature.getClient('events-test');
+    });
+
+    describe('API close', () => {
+      it('should call cancel function on provider', () => {
+        expect(mockCallbackClient.cancelFunction).not.toHaveBeenCalled();
+        OpenFeature.close();
+        expect(mockCallbackClient.cancelFunction).toHaveBeenCalled();
```
toddbaert, line 270: Not called => called.

libs/providers/flagd-web/src/lib/flagd-web-provider.ts
```diff
@@ -110,7 +112,7 @@ export class FlagdWebProvider implements Provider {
     this._delayMs = Math.min(this._delayMs * BACK_OFF_MULTIPLIER, this._maxDelay);
 
     return new Promise<void>((resolve) => {
-      this._callbackClient.eventStream(
+      this._cancelFunction = this._callbackClient.eventStream(
```
toddbaert, line 115: The eventStream returns a reference to a function to cancel the connection.

libs/providers/flagd-web/src/lib/flagd-web-provider.ts
```diff
@@ -131,8 +133,9 @@ export class FlagdWebProvider implements Provider {
             }
           }
         },
-        () => {
-          this._logger?.error(`${FlagdWebProvider.name}: could not establish connection to flagd`);
+        (err) => {
+          this._logger?.error(`${FlagdWebProvider.name}: could not establish connection to flagd, ${err?.message}`);
+          this._logger?.debug(err?.stack);
```
toddbaert, line 138: This is unrelated, but I realized we didn't have error context here so I added some.

libs/providers/flagd-web/src/lib/flagd-web-provider.ts
```diff
@@ -91,6 +87,11 @@ export class FlagdWebProvider implements Provider {
     return this.evaluate(flagKey, 'objectValue');
   }
 
+  onClose(): Promise<void> {
+    // close the steam using the saved cancel function
```
toddbaert, line 91:
```suggestion
    // close the stream using the saved cancel function
```

PR #330, opened 2023-04-19
fix: handling zero value responses (previously undefined)

### LINKED ISSUE #165: [flagd-provider] add unit test for zero-value resolutions from flagd
Flagd sends empty values in flag [response payloads](https://github.com/open-feature/schemas/blob/main/protobuf/schema/v1/schema.proto#L54) when the value is equivalent to the zero-value of the associated go type. We should test that this works as expected and still returns an equivalent zero-value (ie: `false`, `0`).

DoD:
- manual test confirming zero values are returned from the flagd provider when they are missing from the gRPC message (boolean set to false or numeric set to 0 in flagd config)
- new tests that assert correct values for zero-value flags

see: https://github.com/open-feature/flagd/issues/218

odubajDT:
## This PR

- introduce logic for zero value handling
- unit tests

### Related Issues

Fixes #165

### DIFF
--- a/libs/providers/flagd/src/lib/service/grpc/service.ts
+++ b/libs/providers/flagd/src/lib/service/grpc/service.ts
@@ -204,28 +204,28 @@ export class GRPCService implements Service {
   }
 
   private objectParser = (struct: Struct) => {
-    if (struct !== undefined) {
+    if (struct) {
       return Struct.toJson(struct);
     }
     return {}
   };
 
   private booleanParser = (value: boolean) => {
-    if (value !== undefined) {
+    if (value) {
       return value;
     }
     return false
   };
 
   private stringParser = (value: string) => {
-    if (value !== undefined) {
+    if (value) {
       return value;
     }
     return ''
   };
 
   private numberParser = (value: number) => {
-    if (value !== undefined) {
+    if (value) {
       return value;
     }
     return 0
--- a/libs/providers/flagd/src/lib/service/grpc/service.ts
+++ b/libs/providers/flagd/src/lib/service/grpc/service.ts
@@ -114,15 +114,15 @@ export class GRPCService implements Service {
     context: EvaluationContext,
     logger: Logger
   ): Promise<ResolutionDetails<boolean>> {
-    return this.resolve(this._client.resolveBoolean, flagKey, context, logger);
+    return this.resolve(this._client.resolveBoolean, flagKey, context, logger, this.booleanParser);
   }
 
   async resolveString(flagKey: string, context: EvaluationContext, logger: Logger): Promise<ResolutionDetails<string>> {
-    return this.resolve(this._client.resolveString, flagKey, context, logger);
+    return this.resolve(this._client.resolveString, flagKey, context, logger, this.stringParser);
   }
 
   async resolveNumber(flagKey: string, context: EvaluationContext, logger: Logger): Promise<ResolutionDetails<number>> {
-    return this.resolve(this._client.resolveFloat, flagKey, context, logger);
+    return this.resolve(this._client.resolveFloat, flagKey, context, logger, this.numberParser);
   }
 
   async resolveObject<T extends JsonValue>(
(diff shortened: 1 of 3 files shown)


Kavindu-Dodan (approved):
lgtm

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -206,17 +206,37 @@ export class GRPCService implements Service {
   private objectParser = (struct: Struct) => {
     if (struct !== undefined) {
```
thisthat, line 207: maybe it is worth to also nullcheck here
```suggestion
    if (struct) {
```

james-milligan (approved):
LGTM

thisthat (approved):

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -204,19 +204,39 @@ export class GRPCService implements Service {
...
+    return 0
   };
 
   private async resolve<T extends FlagValue, Rq extends AnyRequest, Rs extends AnyResponse>(
     resolver: (request: AnyRequest) => UnaryCall<Rq, Rs>,
     flagKey: string,
     context: EvaluationContext,
     logger: Logger,
-    parser?: (struct: Struct) => PbJsonValue
+    parser?: (value: any) => any
```
toddbaert, line 239: Can we restrict these types a bit more?

It can be a union of types: `Struct | number | string...`. Generally, we shouldn't see explicit `any`s (`unknown` is safer if you really don't know the type)

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -114,15 +114,15 @@ export class GRPCService implements Service {
...
   }
 
   async resolveString(flagKey: string, context: EvaluationContext, logger: Logger): Promise<ResolutionDetails<string>> {
-    return this.resolve(this._client.resolveString, flagKey, context, logger);
+    return this.resolve(this._client.resolveString, flagKey, context, logger, this.stringParser);
   }
 
   async resolveNumber(flagKey: string, context: EvaluationContext, logger: Logger): Promise<ResolutionDetails<number>> {
-    return this.resolve(this._client.resolveFloat, flagKey, context, logger);
+    return this.resolve(this._client.resolveFloat, flagKey, context, logger, this.numberParser);
```
toddbaert, line 125: So it looks like this wasn't working as expected? When a zero-value flag was set in flagd config, `undefined` was returned from the provider? Did you see that in a manual test?

odubajDT: yes exactly, this logic was not in place

libs/providers/flagd/src/lib/service/grpc/service.ts
odubajDT: it was quite problematic here, but will give it another try. If it won't go, I will use `undefined`

odubajDT: I am trying desperately but cannot manage to make it work without the use of `any`. The problem is the conversion of the object resolution, which in this case can be anything. Would you maybe have time to sync regarding this problem? I do not have that deep knowledge regarding JS/Typescript, so maybe I am missing some good manners how to resolve this here...

thisthat: The problem here is that `Struct` and `FlagValue` are not compatible types.

odubajDT: we were working on this together with @thisthat but didn't manage to find a solution. For now we prefer to stay with the current solution

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: Perfect! This was a productive issue then.

Thanks @odubajDT and @james-milligan

PR #343, opened 2023-04-27
feat(launchdarkly-provider): Add new provider

Mateoc:
## This PR
Implements a provider for LaunchDarkly [launchdarkly-js-client-sdk](https://github.com/launchdarkly/js-client-sdk).
Is based on the [provider for the node sdk](https://github.com/launchdarkly/openfeature-node-server)

- LaunchDarklyClientProvider


toddbaert:
Hey @Mateoc ! Thanks for your contribution!

We're currently working on updates to the OpenFeature specification to support its usage in web-clients, as well as other client contexts. See [here](https://github.com/open-feature/spec/issues/167) for the progress.

I don't think that the current js-sdk has the required APIs to support a proper LaunchDarkly provider for use in browser contexts. We do however have an experimental [web-sdk](https://www.npmjs.com/package/@openfeature/web-sdk). I won't speak for LaunchDarkly, but I'd feel more comfortable releasing an experimental provider based on that, than something based on the server-side SDK. You can check out the playground, which features a [demo quality LaunchDarkly provider](https://github.com/open-feature/playground/blob/main/packages/js-launchdarkly-provider/src/lib/js-launchdarkly-provider.ts).

toddbaert:
> @toddbaert Thanks for the quick response. I didn't notice this was exclusively server side specially since there are other web [libraries](https://openfeature.dev/docs/reference/technologies/client/web) using it (probably not meant for client usage I did not check them too much), I will check the experimental library, at a first glance the interface seems similar enough for it to be a simple change
> 
> If I end up fixing it, should I open a new pr here ? or do you have a different repo for contributions on web providers ?

This is a fine location! The [flagd-web](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/flagd-web) provider is another client-side example based on the experimental web-sdk.

Just out of curiosity, are you intending on using this in production somewhere? Or is this more of an experiment?

c4milo:
@toddbaert, production! @Mateoc works with me at @redpanda-data :D

Mateoc:
@toddbaert I made the refactor to use '@openfeature/web-sdk' there is a question I could not find the answer:
though the context in this SDK  is not a parameter to the get functions (example: `getBooleanDetails`) , it still passes it on the evaluation methods, I assumed it always passed the same context until `OpenFeature.setContext` is called (assumption seems to be right as far as I test), so I just ignored that parameter and just handled the context change on `onContextChange` which is called when `OpenFeature.setContext` is called. 
Regarding you previous question, we are still doing a couple of POC but yes the aim is to use this on prod.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  constructor(private client: LDClient, options: LaunchDarklyProviderOptions = {}) {
+    if (options.logger) {
+      this.logger = options.logger;
+    } else {
+      this.logger = basicLogger({ level: 'info' });
+    }
+  }
+
+  initialize(context: EvaluationContext): Promise<void> {
```
kinyoklion, line 46: This is something I have not determined how to reconcile yet. For the LD providers we have allowed construction of the client before the provider. This has many benefits, it acts as an escape hatch for unsupported features, it allows for a gradual transition to an OF implementation using the same client, it minimizes the complexity of configuration, and allows for substantial documentation re-use.

So far we have only implemented server side SDKs, which don't require a context during initialization, so everything has worked well.

The new init methods do pose somewhat of a challenge. The client itself was already initialized with a context, and then initialize here is called with a context. 

Here you have chosen to ignore the context, and fulfill the promise. Which will work, but could produce an unexpected result. Maybe also a burden in the developer having to work with two context formats.

I am not sure how to reconcile that yet. I prefer construction of the client ahead of time, but it may not be feasible, or may need to be optional, to allow for the initialization context to be consistent.

toddbaert: > it may not be feasible, or may need to be optional, to allow for the initialization context to be consistent.

I think I agree with this. Though I see why construction of the client ahead of time could be desirable, doing so somewhat defeats the purpose of the initialization paradigm - it's working against the API, to a degree.

toddbaert:
@Mateoc I'll give this a thorough review today or tomorrow.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
Mateoc: I could also format the context (`translateContext`) and see if it matches the one that is already in the client (`client.getContext`) if it is the same then leave it alone if it isn't update it, that way the user may init it on client creation or with OpenFeature. No unexpected behaviour compared to other providers

toddbaert: That's a possibility, or you could merge the contexts (either shallowly or with something like [lodash's merge](https://lodash.com/docs/#merge)), with the one in `initialize` overwriting conflicting properties. Either way you should clearly document this behavior.

I'd like @kinyoklion's take on this proposal.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+
+      const res =  ofClient.getStringDetails(testFlagKey, 'default');
+      expect(res).toEqual({
+        flagKey: testFlagKey,
+        value: 'default',
+        reason: 'ERROR',
+        errorCode: 'TYPE_MISMATCH',
+      });
+    });
+
```
toddbaert, line 237:
```suggestion
```

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+      const res = ofClient.getNumberDetails(testFlagKey, 0);
+      expect(res).toEqual({
+        flagKey: testFlagKey,
+        value: 0,
+        reason: 'ERROR',
+        errorCode: 'TYPE_MISMATCH',
+      });
+    });
+
```
toddbaert, line 131:
```suggestion
```

libs/providers/launchdarkly-client/src/lib/launchdarkly-provider-options.ts
```diff
@@ -0,0 +1,23 @@
...
+   * the Winston logging package.
+   *
+   * If you do not set this property, the provider uses `basicLogger` with a
+   * minimum level of `info`.
+   *
+   * By default this will not be the same logger instance used by the SDK.
+   * If the provider should share a logger with the SDK, then you will need to
+   * provide the logger in the SDK options as well.
+   */
+  logger?: LDLogger;
```
toddbaert, line 22: The JS SDK also has it's own [logger interface](https://github.com/open-feature/js-sdk/blob/main/packages/client/README.md#logging). I think it makes sense to accept either that or the `LDLogger`, maybe with a type union. I think the types basically overlap.

libs/providers/launchdarkly-client/src/lib/translate-context.spec.ts
```diff
@@ -0,0 +1,230 @@
+import translateContext from "./translate-context";
```
toddbaert, line 1: Same as [here](https://github.com/open-feature/js-sdk-contrib/pull/343/files#r1184951465) if this is lifted from another repo.

libs/providers/launchdarkly-client/README.md
```diff
@@ -0,0 +1,27 @@
...
+``` ts
+import {initialize} from 'launchdarkly-js-client-sdk'; 
+
+const initialContext = {};
+const ldClient =  initialize('LDId', initialContext);
+await ldClient.waitForInitialization()
+const ldOpenFeatureProvider = new LaunchDarklyClientProvider(ldClient);
+
+OpenFeature.setProvider(ldOpenFeatureProvider);
+```
```
toddbaert, line 27: I think we should to rigorously document some of the context-related stuff here, both initialization and updates. For example one thing to note is that updating the context will cause the `identify` function to run, which results in network traffic (correct me if I'm wrong @kinyoklion ).

toddbaert:
> I assumed it always passed the same context until `OpenFeature.setContext` is called (assumption seems to be right as far as I test), so I just ignored that parameter and just handled the context change on `onContextChange` which is called when `OpenFeature.setContext` is called.

@Mateoc This is correct. The reason it's passed is some other solutions' SDKs don't have the equivalent of an `identify` and can accept context in their evaluation methods without such a method being called. Usually this is because those SDKs to all their evaluation locally on a cached ruleset. In the case of LD, what you've done is correct.

libs/providers/launchdarkly-client/README.md
```diff
@@ -0,0 +1,27 @@
+# launchdarkly-client Provider
+
```
toddbaert, line 2: Please add an `experimental` badge, as seen [here](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flagd-web/README.md).

libs/providers/launchdarkly-client/README.md
kinyoklion: That is correct. The nature of the JS client SDK is such.

1. Do a `GET` or a `REPORT` to get the current flags for the identified context.
2. If the stream option is set, or someone registers a flag listener, or they call the `stream` method, then a stream is initiated.
3. The stream remains open getting updates to flag values for that context.

Whenever you identify this process restarts. So we only want to identify if the context is actually different.

Versus our server SDKs which download entire rulesets, and evaluate all contexts against that ruleset. On those SDKs identify also exists, but all it does is send an analytics event so the context can be presented in our UI.

(Currently OF doesn't have listeners, so that is one of the reasons to have direct client access.)

libs/providers/launchdarkly-client/src/lib/launchdarkly-provider-options.ts
Mateoc: The interface for LDLogger is a subtype of Logger from the sdk, so I'll make it Logger instead

Mateoc:
@toddbaert thanks for all you feedback!
Let me know if the license part is ok or if I should use a different format.
Also let me know if I miss something or if you find something new.

libs/providers/launchdarkly-client/README.md
```diff
@@ -0,0 +1,41 @@
...
+  //set open feature provider and get client
+  OpenFeature.setProvider(ldOpenFeatureProvider);
+  const client = OpenFeature.getClient('my-client');
+
+  //use client
+  const boolValue = client.getBooleanValue('boolFlag', false);
+```
+## Update Context
+For context update always use ``OpenFeature.setContext(myNewContext);`` instead of ``ldClient.identify(myNewContext);``, as this will always be handled internally.
```
toddbaert, line 35:
```suggestion
For context update always use ``OpenFeature.setContext(myNewContext);`` instead of ``ldClient.identify(myNewContext);``, as this will always be handled internally.

Please note that context changes result in network traffic, so changes should be made sparingly in accordance to relevant user behavior.
```

libs/providers/launchdarkly-client/README.md
```diff
@@ -0,0 +1,41 @@
+# launchdarkly-client Provider
+
+![Experimental](https://img.shields.io/badge/experimental-breaking%20changes%20allowed-yellow)
+
```
toddbaert, line 4: Another small thing - you may want to re-order the readme to put the usage/initialization stuff first and the development stuff toward the end, as more people will likely be interested in how to use this than how to add to it (the README is also published to NPM).

toddbaert, line 4:
```suggestion

An unofficial browser provider for LaunchDarkly.
```

kinyoklion:
@Mateoc looks like you need to sign-off on some of your commits.

libs/providers/launchdarkly-client/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Installation
+
+```
+$ npm install @openfeature/launchdarkly-client-provider
+```
+
+## Sample initialization
+``` ts
+  // init launchdarkly-js-client-sdk
+  const initialContext = {};
```
kinyoklion, line 16: Being as this context is an LDContext, it should contain either a `key` or `anonymous` should be true (in which case it will generate a key.)

Without something the SDK isn't going to initialize and will log
```
[LaunchDarkly] Invalid context specified. Please see https://docs.launchdarkly.com/sdk/client-side/javascript#initializing-the-client for instructions on SDK initialization.
```

libs/providers/launchdarkly-client/README.md
```diff
@@ -0,0 +1,45 @@
...
+For context update always use ``OpenFeature.setContext(myNewContext);`` instead of ``ldClient.identify(myNewContext);``, as this will always be handled internally.
+
+Please note that context changes result in network traffic, so changes should be made sparingly in accordance to relevant user behavior.
+``` ts
+  await OpenFeature.setContext({ targetingKey: 'my-key' })
+  //Laundarkly uses key but this provider tranlates targetingKey to key; 
+  //So the above is the same as doing
+  await OpenFeature.setContext({ key: 'my-key' });
+```
+
```
kinyoklion, line 38: Potentially it would be good to include either a section similar to this: https://github.com/launchdarkly/openfeature-node-server#openfeature-specific-considerations

Or event a link to that. 

For a very basic context OF is reasonably interoperable, but for a complex context it is useful to understand that several attributes have specialized behavior.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
kinyoklion: I think we need to circle back and get some closure on this. I think most everything else looks good from a code perspective. But initialization is still a concern. I would approve the comparison being discussed earlier, though I think there would need to be readme docs explaining caveats.

Or I would approve a provider that takes the LDOptions interface, and then initializes the review internally.

toddbaert:
> @toddbaert I just fix latest comments. Thanks! There is no rush lets wait to see what they say. Also we will need to create a module that integrates `@openfeature/web-sdk` with react. If we end up creating something worth sharing where could I share it ?

Absolutely! I know @moredip has expressed interest in this, and of course @beeme1mr and myself are also interested. Much of our company's UI is react, we'd certainly need such a module.

I'd encourage you to bring this up in the slack channel, and perhaps create an issue the in the [js-contrib repo](https://github.com/open-feature/js-sdk-contrib).

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
kinyoklion: @Mateoc We still need to conclude this item.

kinyoklion:
> @kinyoklion thanks for the feedback let me know if the changes are ok @toddbaert I had to force push to sign the commits not sure if that is ok

Everything looks good aside from closing the loop on initialization. Thank you for helping work through this.

toddbaert:
> @kinyoklion thanks for the feedback let me know if the changes are ok @toddbaert I had to force push to sign the commits not sure if that is ok

Force push whatever on your own forks! My policy is to try to avoid it to an extent b/c it can be a bit confusing for reviewers, but you gotta do it sometimes.

toddbaert:
@Mateoc you have some lint errors. Mostly around `@ts-ignore`. I would just add whatever cases you need to solve the same problem.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
Mateoc: @kinyoklion done, I also added validation for the anonymous case you mentioned

kinyoklion (approved):
I think we can improve the initialization, but I think this is probably fine considering this will be experimental at this point in time.

toddbaert:
@Mateoc I was able to test this locally and it worked as I'd expect, with the exception of the build (which also failed in the CI). You'll need to add your dependencies to the root `package.json`. Our NX tooling will automatically add the correct dependencies to your module's package.json (you should be able to run `npm run package` and confirm that the `package.json` has your deps in it.

The only think you should have to do manually is specify peer deps. I would specify the LD provider and the current openfeature web-sdk and greater versions as your peer deps. I would recommend you pin the current minor version of the web SDK in peer deps, since below 0, we treat this as the major version.

libs/providers/launchdarkly-client/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "peerDependencies": {
+    "@openfeature/web-sdk": "^0.3.2-experimental",
+    "launchdarkly-js-client-sdk": "^3.1.3"
+  },
+  "dependencies": {
+    "lodash.isequal": "^4.5.0"
+  },
+  "devDependencies": {
+    "@types/lodash.isequal": "^4.5.6"
+  }
```
toddbaert, line 18:
```suggestion

```

toddbaert, line 18:
```suggestion
```

libs/providers/launchdarkly-client/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@openfeature/launchdarkly-client-provider",
+  "version": "0.0.1",
+  "type": "commonjs",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/web-sdk": "^0.3.2-experimental",
```
toddbaert, line 10:
```suggestion
    "@openfeature/web-sdk": "~0.3.2-experimental",
```

libs/providers/launchdarkly-client/package.json
toddbaert, line 18:
```suggestion
  }
```

toddbaert: These should compute automatically if you add them to the root package.json

Mateoc: Done
I did not think this was going to work, is not even in the peer, how ?

toddbaert (approved):

libs/providers/launchdarkly-client/package.json
toddbaert: nx _should_ automatically compute these and put them in the distributed modules package.json. That is how it worked until a recent bug in nx. Upgrading to nx v16 fixes it. I'll do that in a separate PR and make sure everything works again before releasing. I've rebased you branch on main and resolved some conflicts. I'm OK to merge now unless you have any objections.

package.json
```diff
@@ -39,19 +41,21 @@
...
     "@types/object-hash": "^3.0.0",
     "@typescript-eslint/eslint-plugin": "~5.59.0",
     "@typescript-eslint/parser": "~5.59.0",
     "axios-mock-adapter": "1.21.4",
     "babel-preset-minify": "0.5.2",
     "eslint": "~8.40.0",
     "eslint-config-prettier": "8.8.0",
-    "jest": "^29.3.1",
-    "jest-environment-jsdom": "^29.3.1",
+    "jest": "^29.4.1",
```
toddbaert, line 52:
```suggestion
    "jest": "^29.4.1",
```

package.json
```diff
@@ -39,19 +41,21 @@
...
+    "jest-environment-jsdom": "^29.4.1",
     "jest-fetch-mock": "^3.0.3",
+    "lodash.isequal": "^4.5.0",
```
toddbaert, line 55:
```suggestion
```

libs/providers/launchdarkly-client/package.json
toddbaert: @Mateoc FYI, this is how things SHOULD work: https://github.com/open-feature/js-sdk-contrib/pull/362

It generates the following deps for you: 
image

toddbaert:
@Mateoc when [this](https://github.com/open-feature/js-sdk-contrib/pull/360) is merged, your provider will be available on npm

Mateoc:
@toddbaert awesome, thanks

c4milo:
Thanks @toddbaert for all the assistance!

PR #361, opened 2023-05-13
fix: update nx to fix shipped packages

toddbaert:
This updates the repo to nx v16, which fixes a recent regression that caused no dependencies to be computed in the library modules' dist. For example, here is the before and after in the dist/package.json for flagd:

before:
image

after: 
image

The only breaking change I found was that this broke our generators. I had to update the generator script/command.


toddbaert:
Closing, found some bugs with nx/rollout.

PR #362, opened 2023-05-13
chore: fix missing computed deps

toddbaert:
This reverts nx to `v15.9.2`, which fixes a recent regression that caused no dependencies to be computed in the library modules' dist. For example, here is the before and after in the dist/package.json for flagd:

before:
image

after: 
image

Luckily, nothing has been shipped since this regression.

I also tried an upgrade to nx `v16`, which has some breaking changes but also fixed this issue. I nearly got that working, but there appears to be some odd bugs with `nx/rollup` there. I'll try the `v16` upgrade again later.


package.json
```diff
@@ -18,30 +18,31 @@
     "@openfeature/web-sdk": "^0.3.1-experimental",
     "@opentelemetry/api": "^1.3.0",
     "@protobuf-ts/grpc-transport": "2.9.0",
+    "@protobuf-ts/runtime-rpc": "^2.9.0",
```
toddbaert, line 21: This dependency was simply missing, it should have been added with a recent deprecation to protobuf.

thomaspoignant (approved):

thomaspoignant (approved):

thomaspoignant (approved):

beeme1mr (approved):
This is definitely one of the downsides to using nx automatic dependency detection. We should consider adding a build output check to make sure dependencies are added to the built artifact correctly. It's concerning that a patch update to nx would be this impactful. We may also want to disable nx updates in renovate.

toddbaert:
> This is definitely one of the downsides to using nx automatic dependency detection. We should consider adding a build output check to make sure dependencies are added to the built artifact correctly. It's concerning that a patch update to nx would be this impactful. We may also want to disable nx updates in renovate.

Ya, it gets worse. We can talk next week. :weary: 

I think checking the deps is a good idea, but it might be even better to run a full e2e suite. I'm thinking we could use the flagd built on each CI run in the same integration suite that we run in the JS-SDK... Then we'll know at least the packaging works for flagd, so it should be working generally.

PR #366, opened 2023-05-13
chore: migrate to nx 16

toddbaert:
This upgrade `nx` to 16.

The only changes that weren't automatically done by the nx migration automation were:

- changes to the generator scrips: https://github.com/open-feature/js-sdk-contrib/pull/366#discussion_r1193025576
- changes to deal with the new behavior in `nx/rollup`: https://github.com/open-feature/js-sdk-contrib/pull/366#discussion_r1193025957 (see https://github.com/nrwl/nx/pull/17005 and https://github.com/nrwl/nx/pull/16713 for background)


package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.0.0",
   "license": "Apache-2.0",
   "scripts": {
-    "generate-hook": "nx workspace-generator open-feature hook",
-    "generate-provider": "nx workspace-generator open-feature provider",
+    "generate-hook": "nx generate open-feature hook",
+    "generate-provider": "nx generate open-feature provider",
```
toddbaert, line 7: This was a breaking change. I had to change the commands to make them work in 16.

libs/providers/env-var/project.json
```diff
@@ -50,7 +49,7 @@
         "compiler": "tsc",
         "generateExportsField": true,
         "umdName": "Env Var",
-        "external": ["typescript"],
+        "external": "all",
```
toddbaert, line 52: There was a change in nx:rollup that caused anything not specified here to auto-bundle (which admittedly makes sense). That's not what we want here, or at least not what we've been doing. The bundling also caused a bunch of breakage in flagd, some of the gRPC code didn't bundle well.

I create a doc PR in nx: https://github.com/nrwl/nx/pull/17005

tools/workspace-plugin/src/generators/open-feature/index.ts
```diff
@@ -126,7 +126,7 @@ function updateProject(tree: Tree, projectRoot: string, umdName: string) {
         compiler: 'tsc',
         generateExportsField: true,
         umdName,
-        external: ['typescript'],
+        external: 'all',
```
toddbaert, line 129: updated the generator accordingly.

beeme1mr (approved):

PR #418, opened 2023-06-16
feat!: support streaming connection and initialization fixes 

sago2k8:
## This PR

- Updates the use of launch darkly client in the browser.
- Fixes couple of bugs when changing the context and fetching the variations.
- Updates the dependecies  so the initalizaiton of the LD client. happens in the provider.
- expose events for the client to be consumed

### Notes

This is used for our team at [Redpanda](https://redpanda.com/), we use open feature across the stack and we want  to use it in our frontend apps too.

### Follow-up Tasks

- update client references so the web-sdk can listen to events to propagate real time changes:  I might have missed it but I didn't find a way to make the web-sdk react to real time events from the provider.

### Related issues:
- #444


libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -29,78 +32,114 @@ function wrongTypeResult<T>(value: T): ResolutionDetails<T> {
...
+ * Provider Status
+ * defined here https://github.com/open-feature/js-sdk/blob/d26c4eeedde9cc321c69375594f1ccbfaff41a28/packages/shared/src/types.ts#L265
+ * ideally will be exported by '@openfeature/web-sdk'
+ * */
+enum ProviderStatus {
+  NOT_READY = 'NOT_READY',
+  READY = 'READY',
+  ERROR = 'ERROR',
+}
+
```
toddbaert, line 45: This is exported in the last few versions of the web-sdk. You can feel free to update to the latest web-sdk in the root package json of this project.

libs/providers/launchdarkly-client/package.json
```diff
@@ -7,7 +7,7 @@
     "current-version": "echo $npm_package_version"
   },
   "peerDependencies": {
-    "@openfeature/web-sdk": "~0.3.2-experimental",
-    "launchdarkly-js-client-sdk": "^3.1.3"
+    "@openfeature/web-sdk": "*",
```
toddbaert, line 10: Once we release a non-experimental version, we should properly set this. I'll make an issue to do this for all components in the contribs

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -29,78 +32,114 @@ function wrongTypeResult<T>(value: T): ResolutionDetails<T> {
...
+  public events = new OpenFeatureEventEmitter();
 
-  constructor(private client: LDClient, options: LaunchDarklyProviderOptions = {}) {
-    if (options.logger) {
-      this.logger = options.logger;
+  /*
+   * implement status field/accessor
+   * https://openfeature.dev/specification/sections/providers#requirement-242
+   * */
+  public status: ProviderStatus = ProviderStatus.NOT_READY;
```
toddbaert, line 61: An [accessor](https://www.typescriptlang.org/docs/handbook/classes.html#accessors) might be better there. There's no need to allow modification of these field from the outside.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -29,78 +32,114 @@ function wrongTypeResult<T>(value: T): ResolutionDetails<T> {
...
+
+  constructor(private readonly envKey: string, { providerLogger, ...ldOptions }: LaunchDarklyProviderOptions) {
```
toddbaert, line 63: Though it's a breaking change, this is a much more idiomatic API. The previous one made sense before the events spec had fully matured, but now its ergonomics are not ideal

I suppose it would be possible to support the old one as well, but I question the value of doing so.

libs/providers/launchdarkly-client/src/lib/translate-result.ts
```diff
@@ -1,5 +1,5 @@
-import {ErrorCode, ResolutionDetails} from "@openfeature/js-sdk";
-import {LDEvaluationDetail} from "launchdarkly-js-client-sdk";
+import { ErrorCode, ResolutionDetails } from '@openfeature/web-sdk';
```
toddbaert, line 1: nice catch here.

libs/providers/launchdarkly-client/.eslintrc.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": ["../../../.eslintrc.json"],
-  "ignorePatterns": ["!**/*"],
+  "ignorePatterns": ["!**/*", "node_modules"],
```
toddbaert, line 3: Did you get lint errors in the node_modules?

If so, maybe we should move this to the root config?

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -29,78 +32,114 @@ function wrongTypeResult<T>(value: T): ResolutionDetails<T> {
...
+
+  async initialize(): Promise<void> {
+    /* for simplicity always initialize with anonymous by default*/
+
+    this._client = initialize(this.envKey, { anonymous: true }, this.ldOptions);
+
+    /*
+     * set event listeners to propagate ld events to open feature provider
+     * */
+    this.setListeners();
```
toddbaert, line 87: Perhaps this should be optional, based on a configuration?

I'm not sure if all users would make use of this or not, or if there's performance/traffic costs associated with setting up such listeners in LD.

toddbaert (approved):
These look like solid improvements to me. I have a few small concerns, particularly [this](https://github.com/open-feature/js-sdk-contrib/pull/418/files#r1252214568).

@sago2k8 Based on [these policies](https://github.com/open-feature/.github/blob/main/CONTRIBUTING.md#contrib-repositories), we will soon be adding a CODEOWNERs file for each sub component in these monorepos. It's clear you understand the SDK as well as this provider. Would you like to be added to the CODEOWNERS for this component?

I'm awaiting additional reviews, particularly from @kinyoklion  and @Mateoc

sago2k8:
> These look like solid improvements to me. I have a few small concerns, particularly [this](https://github.com/open-feature/js-sdk-contrib/pull/418/files#r1252214568).
> 
> @sago2k8 Based on [these policies](https://github.com/open-feature/.github/blob/main/CONTRIBUTING.md#contrib-repositories), we will soon be adding a CODEOWNERs file for each sub component in these monorepos. It's clear you understand the SDK as well as this provider. Would you like to be added to the CODEOWNERS for this component?
> 
> I'm awaiting additional reviews, particularly from @kinyoklion and @Mateoc

Sure, Happy to be part of the CODEOWERS

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
sago2k8: I updated the package version to `@openfeature/web-sdk`, and I still get a compilation error. 
```Error during bundle: Error: libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts:13:3 - error TS2305: Module '"@openfeature/web-sdk"' has no exported member 'ProviderStatus'.

13   ProviderStatus,```

wondering if I am missing something, I see it in the `common` package.

File: libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts (hunk 3 above)
Comment (sago2k8) [reply libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts]: Done !!

File: libs/providers/launchdarkly-client/.eslintrc.json (hunk 6 above)
Comment (sago2k8) [reply libs/providers/launchdarkly-client/.eslintrc.json:3]: this is the exact error I get: 
```
  Failed to load parser 'babel-eslint' declared in 'libs/providers/launchdarkly-client/node_modules/launchdarkly-js-sdk-common/.eslintrc.yaml': Cannot find module 'babel-eslint'
```
seems that eslint is trying to pick up some modules configuration

File: libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts (hunk 1 above)
Comment (toddbaert) [reply libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts:45]: Perhaps try deleting your node modules, and trying again? Locally, when I install this, I can import `ProviderStatus` without issue. When I `go to definition` in my IDE, I see:

image

Which seems right to me.

File: libs/providers/launchdarkly-client/.eslintrc.json (hunk 6 above)
Comment (toddbaert) [reply libs/providers/launchdarkly-client/.eslintrc.json]: I think running `npm install` in the library folder might be related to your other issue [here](https://github.com/open-feature/js-sdk-contrib/pull/418#discussion_r1252274079).

File: libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts [hunk 8]
```diff
@@ -29,78 +32,125 @@ function wrongTypeResult<T>(value: T): ResolutionDetails<T> {
...
     return this.client.close();
   }
 
+  /** set listeners to LD client and event the correspodent event in the Provider
+   * necessary for LD streaming changes
+   * */
+  private setListeners() {
+    this.client.on('change', (changeset: LDFlagChangeset) => {
+      this.events.emit(ProviderEvents.ConfigurationChanged, {
+        flagKeysChanged: Object.keys(changeset),
```
toddbaert, line 118:
```suggestion
        flagsChanged: Object.keys(changeset),
```

This resolves a compilation issue for me (and the CI) which I think is related to the other issues mentioned in this PR. I _think_ an old version of the web-sdk is being cached _somewhere_ in your workspace :grimacing:

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
toddbaert: You also might have to clear any node modules in the library folder. I generally don't keep these around and try to avoid installing them at all.

libs/providers/launchdarkly-client/src/lib/launchdarkly-provider-options.ts
```diff
@@ -36,5 +36,5 @@ export interface LaunchDarklyProviderOptions {
    * If the provider should share a logger with the SDK, then you will need to
    * provide the logger in the SDK options as well.
    */
-  logger?: LDLogger  | Logger;
+  providerLogger?: LDLogger  | Logger;
```
toddbaert, line 39: personally, I think this is clear enough and we can just keep "logger" but that's a nit.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
sago2k8: Indeed, thank you !!

libs/providers/launchdarkly-client/src/lib/launchdarkly-provider-options.ts
sago2k8: Agree!!

sago2k8:
Thanks for the review @toddbaert, I addressed your comments. Waiting for the extra reviewer

I can safely say @Mateoc will be ok with the constructor changes, but better if we get confirmation.

toddbaert (approved):
Looks great.

I'm not sure if you have any strong opinion on this, but though this does fix some behavior, it also adds significant new features. I wonder if `feat!` would be a better commit type?

I have removed "launch darkly client" from the title since our release flow automatically updates the readmes/release in such a way it's clear which component the commit applies to.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -34,73 +38,110 @@ export class LaunchDarklyClientProvider implements Provider {
...
-  async initialize(context: EvaluationContext): Promise<void> {
-    const newContext = this.translateContext(context);
-    const oldContext = this.client.getContext();
-    const ignoreAnonymous = newContext.anonymous && oldContext.anonymous && !newContext.key
-    if(!ignoreAnonymous && !isEqual(newContext, oldContext)) {
-       await this.client.identify(newContext);
+  async initialize(): Promise<void> {
+    /* for simplicity always initialize with anonymous by default*/
+
+    this._client = initialize(this.envKey, { anonymous: true }, this.ldOptions);
```
kinyoklion, line 81: I don't find initializing with an anonymous context, unconditionally, to be a good idea. It generates less than optimal traffic (possibly an extra stream, and always an extra poll). Additionally it may cause substantial confusion with bootstrapping, as you ideally want that data to be associated with the correct context.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -34,73 +38,110 @@ export class LaunchDarklyClientProvider implements Provider {
...
+
+    /*
+     * set event listeners to propagate ld events to open feature provider,
+     * use LD property streaming, since when enabled you are opting in for straming updates
+     * */
+    if (this.ldOptions?.streaming) {
+      this.setListeners();
```
kinyoklion, line 88: This is probably fine, but the SDK itself would typically also start streaming if anyone listened for updates. So, this is kind of an indirect way of opting in.

Because of that I think we should add a note to the readme indicating that events will only be emitted if streaming is enabled.

kinyoklion (changes requested):

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
toddbaert: > This is probably fine, but the SDK itself would typically also start streaming if anyone listened for updates. So, this is kind of an indirect way of opting in.
> 
> Because of that I think we should add a note to the readme indicating that events will only be emitted if streaming is enabled.

In that case, we should probably add a note to the readme as you say, AND extend/override the jsdoc on the configuration object (I think you can add a link to the overridden jsdoc as well).

cc @sago2k8

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
toddbaert: Would your recommended solution be to make this a required part of the configuration then?

kinyoklion: So, before, it allowed init with a context, and then needed to do a comparison. This was because the client existed before the provider.

Here I am confused though, I think this should just use the context that should be passed to `initialize`. Then the user would just need to set a global context before setting the provider? In which case it would just get initialized with that context?

sago2k8: Yeap I think I focus on my use case which is a error obviously, fixing it now, thanks @kinyoklion

sago2k8:
Thanks for the review @kinyoklion, would you review it again ?

sago2k8:
> Looks great.
> 
> I'm not sure if you have any strong opinion on this, but though this does fix some behavior, it also adds significant new features. I wonder if `feat!` would be a better commit type?
> 
> I have removed "launch darkly client" from the title since our release flow automatically updates the readmes/release in such a way it's clear which component the commit applies to.

Agree, set title to feat!

libs/providers/launchdarkly-client/README.md
```diff
@@ -11,34 +11,58 @@ $ npm install @openfeature/launchdarkly-client-provider
...
+ * @see https://launchdarkly.github.io/js-client-sdk/interfaces/LDOptions.html
+ */
+const ldOptions = {
+  straming: true,
+};
+
+/*
+ * You can set the initial context, this would be the context you pass to the LD initialize function
+ */
+const inialLDContext = {
```
kinyoklion, line 30: `inialLDContext` -> `initialLDContext`.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -34,73 +38,113 @@ export class LaunchDarklyClientProvider implements Provider {
...
-  async initialize(context: EvaluationContext): Promise<void> {
-    const newContext = this.translateContext(context);
-    const oldContext = this.client.getContext();
-    const ignoreAnonymous = newContext.anonymous && oldContext.anonymous && !newContext.key
-    if(!ignoreAnonymous && !isEqual(newContext, oldContext)) {
-       await this.client.identify(newContext);
+  async initialize(): Promise<void> {
+    /* for simplicity always initialize with anonymous by default*/
+
+    this._client = initialize(this.envKey, this.initiaContext, this.ldOptions);
```
kinyoklion, line 84: `initiaContext` -> `initialContext`.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
```diff
@@ -34,73 +38,113 @@ export class LaunchDarklyClientProvider implements Provider {
...
+    return this._client;
   }
 
-  async initialize(context: EvaluationContext): Promise<void> {
-    const newContext = this.translateContext(context);
-    const oldContext = this.client.getContext();
-    const ignoreAnonymous = newContext.anonymous && oldContext.anonymous && !newContext.key
-    if(!ignoreAnonymous && !isEqual(newContext, oldContext)) {
-       await this.client.identify(newContext);
+  async initialize(): Promise<void> {
```
kinyoklion, line 81: I am potentially still missing something, but I think this takes a context, and my expectation is that context is used.


`  async initialize(context: EvaluationContext): Promise<void> {
	this._client = initialize(this.envKey, this.translateContext(evaluationContext), this.ldOptions);
`

kinyoklion (changes requested):

toddbaert:
@sago2k8 @Mateoc related PR for ownership of this component: https://github.com/open-feature/js-sdk-contrib/pull/446/files#diff-0829329780a5428e2182662b54354b425c340a08b0b4468048abce4bd9d0312bR23-R26

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
sago2k8: You are completely right I was overlooking an important sdk behaviour, I think it's alright now, could you check it out @kinyoklion

toddbaert:
@sago2k8 I'm merging main into this to fix the CI issue.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
sago2k8: I did added some more context in the README but nothing really in the jsdoc, do you happen to have an example on how to override the jsdoc properly?

kinyoklion (approved):
Minor comment, but I think things look good.

toddbaert:
I'll give @sago2k8 some time to address any nits or make final changes. Will merge tomorrow EOD unless I hear otherwise.

sago2k8:
Hey @toddbaert I don't think I have received and invitation to an org yet.

toddbaert:
> Hey @toddbaert I don't think I have received and invitation to an org yet.

Very sorry @sago2k8 ! I added you and then removed you again when I alphabetized :facepalm: .

I've opened another PR to add you. When [this](https://github.com/open-feature/community/pull/228) is merged, you will get an invite.

PR #422, opened 2023-06-20
feat!: implement events and shutdown for spec 0.6.0

### LINKED ISSUE #421: [configcat-provider] add spec 0.6.0 support
Implement the changes to support the new features of spec 0.6.0.
Initialization is not needed by ConfigCat.

```[tasklist]
### Tasks
- [x] Shutdown ConfigCat on provider close
- [x] Ready, ConfigChanged and Error from ConfigCat are emitted by the provider
```

lukas-reining:
## This PR

Implements events and shutdown from https://github.com/open-feature/js-sdk/issues/424.

### Related Issues

Fixes #421


libs/providers/config-cat/README.md
beeme1mr, line 1: Can you mention the peer dependency requirement in the readme? Please include the minimum OpenFeature SDK version as well.

libs/providers/config-cat/src/lib/config-cat-provider.ts
```diff
@@ -1,36 +1,52 @@
...
+    client.on('clientReady', () => this.events.emit(ProviderEvents.Ready));
+    client.on('configChanged', (projectConfig) =>
+      this.events.emit(ProviderEvents.ConfigurationChanged, { metadata: { ...projectConfig } })
+    );
+    client.on('clientError', (message: string, error) =>
+      this.events.emit(ProviderEvents.Error, {
+        message: message,
+        metadata: error,
+      })
+    );
```
beeme1mr, line 36: Could you move this to the initialize method?

https://open-feature.github.io/js-sdk/interfaces/OpenFeature_JS_SDK.Provider.html#initialize

beeme1mr (commented):
Overall, it looks good. Just added a few comments.

libs/providers/config-cat/README.md
lukas-reining: @beeme1mr Added it.

libs/providers/config-cat/README.md
```diff
@@ -8,6 +8,15 @@ This provider is an implementation for [ConfigCat](https://configcat.com) a mana
 $ npm install @openfeature/config-cat-provider
 ```
 
+#### Required peer dependencies
+
+The OpenFeature SDK is required as peer dependency.
+The minimum required version of `@openfeature/js-sdk` currently is `1.3.0`.
+
+```
+$ npm install @openfeature/js-sdk
```
beeme1mr, line 17: It looks like `configcat-js` is also a peer dependency.

libs/providers/config-cat/src/lib/config-cat-provider.ts
```diff
@@ -1,36 +1,54 @@
...
+  };
 
   private constructor(client: IConfigCatClient) {
     this.client = client;
   }
 
-  public static create(...params: Parameters<typeof configcat.getClient>) {
-    return new ConfigCatProvider(configcat.getClient(...params));
+  public async initialize(context?: EvaluationContext): Promise<void> {
+    this.client.on('clientReady', () => this.events.emit(ProviderEvents.Ready));
```
toddbaert, line 29: I think you want to change this so that you create a new Promise in this func, which is resolved with your client ready event, and return that.

Then the SDK will await it, and it will return ready itself.

lukas-reining: Oh totally overlooked that.
Committed a change.
I had to remove the option for a provider to be created by injecting a ConfigCat Client because there I could not safely assume to get the ClientReady event emitted. Only when subscribing to that event as a setupHook I can safely listen to that and await it.

libs/providers/config-cat/src/lib/config-cat-provider.ts
```diff
@@ -1,42 +1,88 @@
...
+          hooks.on('clientError', (message: string, error) =>
+            this.events.emit(ProviderEvents.Error, {
+              message: message,
+              metadata: error,
+            })
+          );
+        };
+      }
+
+      this.client = getClient(...originalParameters);
```
beeme1mr, line 65: This will cause the promise returned in the initialization to hang indefinitely if `options` is falsy.

lukas-reining: Oh yes, sure it does🤦‍♂️
Did not pay enough attention, fixed it thanks @beeme1mr.

beeme1mr (approved):

libs/providers/config-cat/src/lib/config-cat-provider.ts
```diff
@@ -1,42 +1,87 @@
...
+        );
+
+        hooks.on('clientError', (message: string, error) =>
+          this.events.emit(ProviderEvents.Error, {
+            message: message,
+            metadata: error,
+          })
+        );
+      };
+
```
toddbaert, line 63: This looks great to me now.

toddbaert:
@lukas-reining I'll leave it up to you, since you're the owner of this provider, but I think I would consider marking this as a breaking change for your provider, because it requires `1.3.0` and won't work properly without it. I expect providers to have breaking changes _fairly_ often, and usually they should be easy to absorb since it's just usually some configuration code or deps that need updating.

You can mark this breaking by changing the PR title to `feat!: implement events and shutdown for spec 0.6.0` (note the `!`)

toddbaert (approved):

odubajDT (approved):
I am not a JS/TS expert, but LGTM :)

PR #425, opened 2023-06-27
feat: Return CACHED reason when using cache

thomaspoignant:
## This PR
- Return `CACHED` as a reason when using the cache in the provider.


thomaspoignant:
@toddbaert no worries I missed it too 🙈

PR #429, opened 2023-06-28
fix: use copy-anything for deepClone to be compatible with nodeJS before v17

### LINKED ISSUE #428: Newest npm go-feature-flag-provider version incompatible with older NodeJS versions
@thomaspoignant just a heads up that the most recent change to the the JS `go-feature-flag-provider` [here](https://github.com/open-feature/js-sdk-contrib/commit/1a76aec01350cfab7c3324dcc5e15ea4e90e1462#diff-ee3cb7dbd197df1db128ec8577ed4071880f3019afd227280bbb438518db574eR123) used a feature (`structuredClone()`) that is [only available in NodeJS 17](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) and above. This change will break builds for anyone who has the default value from `npm install` in their `package.json` (which is `"@openfeature/go-feature-flag-provider": "^0.5.8"`) and using a NodeJS version either 16 or below.

This should probably be considered a breaking change if not a in the NodeJS world.

For anyone who uses this and finds their build is randomly failing, you can just remove the `^` from the version in `package.json` to pin the version so it works for older NodeJS versions.

`"@openfeature/go-feature-flag-provider": "^0.5.8"` -> `"@openfeature/go-feature-flag-provider": "0.5.8"`

thomaspoignant:
## This PR
- use `copy-anything` for deepClone to be compatible with nodeJS before v17

Fixes #428

### DIFF
--- a/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
+++ b/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
@@ -10,6 +10,7 @@ import {
   TypeMismatchError,
 } from '@openfeature/js-sdk';
 import axios from 'axios';
+import _ from 'lodash';
 import {transformContext} from './context-transformer';
 import {ProxyNotReady} from './errors/proxyNotReady';
 import {ProxyTimeout} from './errors/proxyTimeout';
@@ -117,11 +118,11 @@ export class GoFeatureFlagProvider implements Provider {
    */
   async callGoffDataCollection() {
     if (this.dataCollectorBuffer?.length === 0) {
-      return
+      return;
     }
 
-    const dataToSend = structuredClone(this.dataCollectorBuffer);
-    this.dataCollectorBuffer = []
+    const dataToSend = _.cloneDeep<FeatureEvent<any>[] | undefined>(this.dataCollectorBuffer) || [];
+    this.dataCollectorBuffer = [];
 
     const request: DataCollectorRequest<boolean> = {events: dataToSend, meta: this.dataCollectorMetadata,}
     const endpointURL = new URL(this.endpoint);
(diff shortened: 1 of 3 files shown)


libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
```diff
@@ -10,6 +10,7 @@ import {
   TypeMismatchError,
 } from '@openfeature/js-sdk';
 import axios from 'axios';
+import _ from 'lodash';
```
toddbaert, line 13: I _think_ it's pure luck that this works, getting this transitively though another dep.

We don't have `lodash` as a dependency (though we do have the modularized [lodash.isEqual](https://www.npmjs.com/package/lodash.isequal) package) to keep bundle size small.

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
```diff
@@ -117,11 +118,11 @@ export class GoFeatureFlagProvider implements Provider {
    */
   async callGoffDataCollection() {
     if (this.dataCollectorBuffer?.length === 0) {
-      return
+      return;
     }
 
-    const dataToSend = structuredClone(this.dataCollectorBuffer);
-    this.dataCollectorBuffer = []
+    const dataToSend = _.cloneDeep<FeatureEvent<any>[] | undefined>(this.dataCollectorBuffer) || [];
```
beeme1mr, line 124: Could you use `JSON.parse(JSON.stringify())`? That should work if `FeatureEvent` doesn't include circular references, maps, or sets.

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
thomaspoignant: oh good catch, my IDE worked on this one so I trust it _(I should be more careful the next time)_.

toddbaert (approved):

beeme1mr (approved):

Issue #447, opened 2023-07-06
Add OTel metrics hook


toddbaert:
# Add OTel metrics hook

## Reasoning

I want to add a hook which captures OTel metrics, similar to https://github.com/open-feature/go-sdk-contrib/blob/main/hooks/open-telemetry/pkg/metrics.go

## Requirements:

- [ ] implement `Hook` interface
- [ ] add tests
- [ ] complete README

cc @Kavindu-Dodan

toddbaert:
@beeme1mr @Kavindu-Dodan @lukas-reining 

Currently, I think there are two problems with this package:

- it's (unfortunately) on version 6.0 (we probably released the 1.0 of that one too soon).
- it, and it's exports, are poorly named. The `OpenTelemetryHook` export should probably be called `TraceHook` or `TracingHook`. The package is called `@openfeature/open-telemetry-hook` which is singular.

I think we should deprecate this package, create a new one called `@openfeature/open-telemetry-hooks` (note the 's'), include both a metrics and tracing hook, and release it as sub `1.0`.

How do you guys feel about this?

toddbaert:
> That sounds reasonable. We'll just need to make sure we depreciate the old one in the repo and on npm. We'll also want to update the playground once the new package is available.

I was just going to change the name and version of the package in the repo, then deprecate the old. We could also create a brand new module, but I think that would be more trouble than it's worth, personally. We could put a note on the readme that the package name was changed from `@openfeature/open-telemetry-hook` to `@openfeature/open-telemetry-hooks`. What do you think?

Kavindu-Dodan:
Another option is to deprecate open-telemetry-hook.ts [1] and extract it into new implementation class. 

I think it's not good to deprecate the package as there seems to be some usage [1] for the package. Rather, deprecating the constructor and pointing to a newer one can be better 🤔 


[1] - https://github.com/open-feature/js-sdk-contrib/blob/main/libs/hooks/open-telemetry/src/lib/open-telemetry-hook.ts
[1] - https://www.npmjs.com/package/@openfeature/open-telemetry-hook

Kavindu-Dodan:
May be this is a good time to introduce a new package 🤔 This will help maintainers and allow us to add more features. It should be an easy migration for adopters as long as it's a change in package naming and constructor. So +1 from my end for a package naming if other agree

PR #449, opened 2023-07-07
fix: deprecate otel hook

toddbaert:
Giving some warning and redirecting people to the new package.

See discussion [here](https://github.com/open-feature/js-sdk-contrib/issues/447#issuecomment-1625550744).


beeme1mr (approved):
Approved, but we'll want to time this release with the release of the new hooks. We'll also want to mark the package as deprecated in NPM.

PR #460, opened 2023-07-12
chore: fix ci permissions

toddbaert:


.github/workflows/component-owners.yml
```diff
@@ -3,9 +3,9 @@ on:
   pull_request_target:
 
 permissions:
-  contents: read              # to read changed files
-  issues: read|write          # to read/write issue assignees
-  pull-requests: read|write   # to read/write PR reviewers
+  contents: read          # to read changed files
+  issues: write           # to read/write issue assignees
+  pull-requests: write    # to read/write PR reviewers
```
toddbaert, line 8: I was specifically wondering if write implied read, but found zero doc on this so took a guess. It turns out it does, and `read|write` is invalid.

PR #474, opened 2023-07-19
Init web-sdk go-feature-flag

thomaspoignant:
## This PR
- 1st version of the web-sdk for GO Feature Flag

### Related Issues

Fixes https://github.com/thomaspoignant/go-feature-flag/issues/662

### DIFF
--- a/libs/providers/go-feature-flag-web/README.md
+++ b/libs/providers/go-feature-flag-web/README.md
@@ -1,0 +1,15 @@
+# go-feature-flag-web Provider
+
+## Installation
+
+```
+$ npm install @openfeature/go-feature-flag-web-provider
+```
+
+## Building
+
+Run `nx package providers-go-feature-flag-web` to build the library.
+
+## Running unit tests
+
+Run `nx test providers-go-feature-flag-web` to execute the unit tests via [Jest](https://jestjs.io).
--- a/libs/providers/go-feature-flag-web/package.json
+++ b/libs/providers/go-feature-flag-web/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "@openfeature/go-feature-flag-web-provider",
+  "version": "0.0.1",
+  "type": "commonjs",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/web-sdk": "*"
+  }
+}
--- a/libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
+++ b/libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
@@ -1,0 +1,233 @@
+import {
+  EvaluationContext, FlagMetadata,
+  FlagNotFoundError,
+  FlagValue,
+  Logger,
+  OpenFeature,
+  OpenFeatureEventEmitter,
+  Provider,
+  ProviderEvents,
+  ResolutionDetails, ResolutionReason,
+  StandardResolutionReasons,
+  TypeMismatchError,
+} from "@openfeature/web-sdk";
+import {GoFeatureFlagAllFlagRequest, GOFeatureFlagAllFlagsResponse, GoFeatureFlagWebProviderOptions,} from "./model";
+import axios from "axios";
+import {transformContext} from "./context-transformer";
+
+export class GoFeatureFlagWebProvider implements Provider {
+  private readonly _websocketPath = "ws/v1/flag/change"
+
+
+  metadata = {
+    name: GoFeatureFlagWebProvider.name,
+  };
+
+  // logger is the Open Feature logger to use
+  private _logger?: Logger;
+  // endpoint of your go-feature-flag relay proxy instance
+  private readonly _endpoint: string;
+  // timeout in millisecond before we consider the http request as a failure
+  private readonly _apiTimeout: number;
+
+  // initial delay in millisecond to wait before retrying to connect the websocket
+  private readonly _websocketRetryInitialDelay;
+  // multiplier of _websocketRetryInitialDelay after each failure
+  private readonly _websocketRetryDelayMultiplier;
+  // maximum number of retries
+  private readonly _websocketMaxRetries;
+  private _websocket?: WebSocket;
+  private _flags: { [key: string]: ResolutionDetails<FlagValue> } = {};
+
+  events = new OpenFeatureEventEmitter();
+
+  constructor(options: GoFeatureFlagWebProviderOptions, logger?: Logger) {
+    this._logger = logger;
+    this._apiTimeout = options.apiTimeout || 0; // default is 0 = no timeout
+    this._endpoint = options.endpoint;
+    this._websocketRetryInitialDelay = options.websocketRetryInitialDelay || 100;
+    this._websocketRetryDelayMultiplier = options.websocketRetryDelayMultiplier || 2;
+    this._websocketMaxRetries = options.websocketMaxRetries || 10;
+    // Add API key to the headers
+    if (options.apiKey) {
+      axios.defaults.headers.common['Authorization'] = `Bearer ${options.apiKey}`;
+    }
+  }
+
+  async initialize(context: EvaluationContext): Promise<void> {
+    await this.fetchAll(context);
+    await this.connectWebsocket();
+    this._logger?.debug("go-feature-flag provider initialized")
+  }
+
+  /**
+   * connectWebsocket is starting the websocket and associate some handler
+   * to react if the state of the websocket change.
+   */
+  async connectWebsocket(): Promise<void> {
+    const wsURL = new URL(this._endpoint);
+    wsURL.pathname =
+      wsURL.pathname.endsWith('/') ? wsURL.pathname + this._websocketPath : wsURL.pathname + '/' + this._websocketPath;
+    wsURL.protocol = "ws"
+    this._logger?.debug(`Trying to connect the websocket at ${wsURL}`)
+
+    this._websocket = new WebSocket(wsURL, ["ws", "http", "https"]);
+    await this.waitWebsocketFinalStatus(this._websocket);
+
+    this._websocket.onopen = (event) => {
+      this._logger?.info(`Websocket to go-feature-flag open: ${event}`);
+    };
+    this._websocket.onmessage = async () => {
+      this._logger?.info(`Change in your configuration flag`);
+      await this.fetchAll(OpenFeature.getContext());
+    }
+    this._websocket.onclose = async () => {
+      this._logger?.info(`Websocket closed, trying to reconnect`);
+      await this.reconnectWebsocket();
+    };
+    this._websocket.onerror = async (event: Event) => {
+      this._logger?.error(`Error while connecting the websocket: ${event}`);
+      await this.reconnectWebsocket();
+    };
+  }
+
+  /**
+   * waitWebsocketFinalStatus is waiting synchronously for the websocket to be in a stable
+   * state (CLOSED or OPEN).
+   * @param socket - the websocket you are waiting for
+   */
+  waitWebsocketFinalStatus(socket: WebSocket): Promise<void> {
+    return new Promise((resolve) => {
+      const checkConnection = () => {
+        if (socket.readyState === WebSocket.OPEN || socket.readyState === WebSocket.CLOSED) {
+          return resolve();
+        }
+        // Wait 5 milliseconds before checking again
+        setTimeout(checkConnection, 5);
+      };
+      checkConnection();
+    });
+  }
+
+
+  /**
+   * reconnectWebsocket is using an exponential backoff pattern to try to restart the connection
+   * to the websocket.
+   */
+  async reconnectWebsocket() {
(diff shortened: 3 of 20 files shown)


libs/providers/go-feature-flag-web/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@openfeature/go-feature-flag-web-provider",
+  "version": "0.0.1",
+  "type": "commonjs",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/web-sdk": "*"
```
toddbaert, line 10: @thomaspoignant we could consider now releasing a non-experimental web-sdk version now that the spec is merged, which would allow you to actually version this more appropriately.

thomaspoignant: @toddbaert Yes, it would be even better if we have a final release.

libs/providers/go-feature-flag-web/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Reconnection
+If the connection to the GO Feature Flag instance fails, the provider will attempt to reconnect with an exponential back-off.   
+The `websocketMaxRetries` can be specified to customize reconnect behavior.
+
+### Event streaming
+The `FlagdWebProvider` receives events from flagd with changes. Combined with the event API in the web SDK, this allows for subscription to flag value changes in clients.
+
+```typescript
+client.addHandler(ProviderEvents.Ready, () => {
```
toddbaert, line 72: You may want to switch this to a flag value changed example to match the text above, plus it's a more interesting feature.

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,269 @@
...
+    this._websocketMaxRetries = options.websocketMaxRetries || 10;
+    // Add API key to the headers
+    if (options.apiKey) {
+      axios.defaults.headers.common['Authorization'] = `Bearer ${options.apiKey}`;
+    }
+  }
+
+  async initialize(context: EvaluationContext): Promise<void> {
+    await this.fetchAll(context);
+    await this.connectWebsocket();
```
toddbaert, line 66: Could you do these at the same time with a Promise.all? Do they need to go in order?

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,269 @@
...
+    this._websocketRetryInitialDelay = options.websocketRetryInitialDelay || 100;
+    this._websocketRetryDelayMultiplier = options.websocketRetryDelayMultiplier || 2;
+    this._websocketMaxRetries = options.websocketMaxRetries || 10;
+    // Add API key to the headers
+    if (options.apiKey) {
+      axios.defaults.headers.common['Authorization'] = `Bearer ${options.apiKey}`;
+    }
+  }
+
+  async initialize(context: EvaluationContext): Promise<void> {
```
toddbaert, line 64: I think you should start your provider in a `NOT_READY` state, and update it once the init is done. This is actually _needed_ in other SDKs, but we have a small bug in the one (IMO) where we initialize every time. This can lead to (in some situations) multiple inits of the same provider.

See [here](https://github.com/open-feature/js-sdk/issues/505).

libs/providers/go-feature-flag-web/src/lib/model.ts
```diff
@@ -0,0 +1,91 @@
+import {
+  ErrorCode,
+  EvaluationContextValue,
+} from '@openfeature/js-sdk';
+import {FlagValue, ResolutionDetails} from "@openfeature/web-sdk";
```
toddbaert, line 5:
```suggestion
import {FlagValue} from "@openfeature/web-sdk";
```

libs/providers/go-feature-flag-web/tsconfig.lib.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "lib": ["ES2015", "DOM"],
```
toddbaert, line 4: I wonder if we need new generators for client components. I guess you added `DOM` manually?

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,269 @@
...
+      this._logger?.info(`error while reconnecting the websocket, next try in ${delay} ms (${attempt}/${this._websocketMaxRetries}).`)
+    }
+    this.events.emit(ProviderEvents.Stale, {
+      message: 'impossible to get status from GO Feature Flag (websocket connection stoppe'
+    });
+  }
+
+  onClose(): Promise<void> {
+    this._websocket?.close(1000, "Closing GO Feature Flag provider");
+    return Promise.resolve(undefined);
```
toddbaert, line 164:
```suggestion
    return Promise.resolve();
```

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  async onContextChange(_: EvaluationContext, newContext: EvaluationContext): Promise<void> {
+    this._logger?.debug(`new context provided: ${newContext}`);
+    await this.fetchAll(newContext);
+  }
```
toddbaert, line 170:
```suggestion
  onContextChange(_: EvaluationContext, newContext: EvaluationContext): Promise<void> {
    this._logger?.debug(`new context provided: ${newContext}`);
    return this.fetchAll(newContext);
  }
```

I think you can just return this since it's returning a void Promise.

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,269 @@
...
+  StandardResolutionReasons,
+  TypeMismatchError,
+} from "@openfeature/web-sdk";
+import {
+  GoFeatureFlagAllFlagRequest,
+  GOFeatureFlagAllFlagsResponse,
+  GoFeatureFlagWebProviderOptions,
+  GOFeatureFlagWebsocketResponse,
+} from "./model";
+import axios from "axios";
```
toddbaert, line 20: TOTALLY up to you, but I would recommend just using `fetch`. With a simple wrapper it will work pretty well. The only gotcha is I think by default it doesn't throw on non-200, but you can change that without much effort. It will keep you package size down at least, which is nice for client projects where every byte of JS counts.

toddbaert: You could also totally do this as a new issue :shrug:

toddbaert (approved):
You are the go-sdk expert, so I of course approve! I did leave a lot of optional comments. The only one that really worries me is [this one](https://github.com/open-feature/js-sdk-contrib/pull/474/files#r1274012627) (which is also related to an SDK bug).

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,269 @@
...
+    await this.fetchAll(context);
+    await this.connectWebsocket();
+    this._logger?.debug("go-feature-flag provider initialized")
+  }
+
+  /**
+   * connectWebsocket is starting the websocket and associate some handler
+   * to react if the state of the websocket change.
+   */
+  async connectWebsocket(): Promise<void> {
```
toddbaert, line 74: Very very cool to have WS support!

libs/providers/go-feature-flag-web/package.json
toddbaert: I think we can release a sub 1.0 web-sdk tomorrow if you'd like, so you can actually put a proper version here. cc @beeme1mr @lukas-reining @kinyoklion

lukas-reining: @toddbaert That would be really good.

thomaspoignant: @toddbaert Perfect, I will wait for the release anyway before releasing the provider.

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,269 @@
...
+      this._logger?.info(`Websocket to go-feature-flag open: ${event}`);
+    };
+    this._websocket.onmessage = async ({data}) => {
+      this._logger?.info(`Change in your configuration flag`);
+      const t: GOFeatureFlagWebsocketResponse = JSON.parse(data);
+      const flagsChanged = this.extractFlagNamesFromWebsocket(t);
+      await this.fetchAll(OpenFeature.getContext(), flagsChanged);
+    }
+    this._websocket.onclose = async () => {
+      this._logger?.info(`Websocket closed, trying to reconnect`);
```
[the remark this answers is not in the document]
thomaspoignant: It could be a lot of things here, such as the connection is closed from the upstream, the server does not exist anymore because we have downscale etc ...

But I agree using `warn` here is better.

libs/providers/go-feature-flag-web/tsconfig.lib.json
thomaspoignant: Yes, I copy/paste it from the `flagd-web` provider, but yes I agree we need a generator for web provider.

libs/providers/go-feature-flag-web/package.json
toddbaert: https://github.com/open-feature/js-sdk/pull/506#issuecomment-1652158329

thomaspoignant:
@toddbaert I've addressed all your comments, including using `fetch` rather than `axios`.
If you don't mind doing another review it would be awesome.

I think we should wait the release of a stable SDK to be available before merging this PR.

toddbaert:
Sorry @thomaspoignant I will re-review this first thing tomorrow

thomaspoignant:
> Sorry @thomaspoignant I will re-review this first thing tomorrow

No rush, no worries

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,341 @@
...
+      try {
+        await this.fetchAll(ctx, flagsChanged);
+        this._status = ProviderStatus.READY;
+        return
+      } catch (err) {
+        this._status = ProviderStatus.ERROR;
+        this.handleFetchErrors(err)
+        await new Promise((resolve) => setTimeout(resolve, delay));
+        delay *= this._retryDelayMultiplier;
+        this._logger?.info(`xxx ${delay} ms (${attempt}/${this._maxRetries}).`)
```
toddbaert, line 246: Is this "xxx" supposed to be here?

Something I like to do is use the classname as a prefix to my log messages: `${GoFeatureFlagWebProvider.name}: ....`

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.ts
```diff
@@ -0,0 +1,341 @@
...
+      Accept: 'application/json',
+    });
+    if(this._apiKey){
+      headers.set('Authorization', `Bearer ${this._apiKey}`);
+    }
+
+    const init: RequestInit = {
+      method: "POST",
+      headers: {
+        'Content-Type': 'application/json',
```
toddbaert, line 276: You could extract the common headers to a template object, or even add them into a convenience method.

toddbaert (approved):
@thomaspoignant looks great to me! Only minor nits that are opinions. Please be careful resolving conflicts in the `.release-please-manifest` :sweat_smile: 

I plan on releasing our first non-experimental web-sdk on Monday, with this change: https://github.com/open-feature/js-sdk/pull/507#issuecomment-1654678449. I'm waiting until then to give people a few days to update the providers impacted by the change.

You can wait until then if you'd like to set a proper peer for this.

libs/providers/go-feature-flag-web/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@openfeature/go-feature-flag-web-provider",
+  "version": "0.0.1",
+  "type": "commonjs",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/web-sdk": "^0.4.0"
```
thomaspoignant, line 10: I've stick the version to the official release to come.

beeme1mr:
@toddbaert, it looks like the CI is failing because of flagd. I'm assuming the NX affecting command detected so many components because the root package json was modified, right? Either way, we should unblock Thomas.

thomaspoignant:
> Awesome, nice job @thomaspoignant! We should extend the playground to support client-side GOFF once this has been published.

Sure I will look at it as soon as we merge it.

PR #475, opened 2023-07-19
fix: otel metric semantic convs

toddbaert:
The keys were wrong :grimacing:


beeme1mr:
Perhaps you should move these to the const file since it's the same between metrics and spans.

https://github.com/open-feature/js-sdk-contrib/blob/main/libs/hooks/open-telemetry/src/lib/constants.ts

toddbaert:
> Perhaps you should move these to the const file since it's the same between metrics and spans.
> 
> https://github.com/open-feature/js-sdk-contrib/blob/main/libs/hooks/open-telemetry/src/lib/constants.ts

yep

toddbaert:
@beeme1mr updated.

Kavindu-Dodan (approved):

libs/hooks/open-telemetry/src/lib/attributes.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const ACTIVE_COUNT_NAME = `${FEATURE_FLAG}.evaluation_active_count`;
+export const REQUESTS_TOTAL_NAME = `${FEATURE_FLAG}.evaluation_requests_total`;
+export const SUCCESS_TOTAL_NAME = `${FEATURE_FLAG}.evaluation_success_total`;
+export const ERROR_TOTAL_NAME = `${FEATURE_FLAG}.evaluation_error_total`;
+
+export type EvaluationAttributes = {[key: `${typeof FEATURE_FLAG}.${string}`]: string | undefined };
+export type ExceptionAttribute = { [key in `${typeof FEATURE_FLAG}.exception`]: string };
+
+export const KEY_ATTR: keyof EvaluationAttributes = `${FEATURE_FLAG}.key`;
```
Kavindu-Dodan, line 11: nit - can refer to FF sem conv documentation for first three attributes : https://opentelemetry.io/docs/specs/otel/logs/semantic_conventions/feature-flags/

toddbaert: Added this link in a comment.

beeme1mr (approved):

Issue #382, opened 2023-05-16
[flagd-provider] add events, init support


toddbaert:
Once the js-sdk has `events` and `init/shutdown` support, we need to update the flagd-provider to emit events and do proper init/shutdown. We already have a stream implemented for cache-busing, so this should be pretty simple.

Definition of done:
- flagd does does startup logic (like initializing it's stream) in `initialize` function
- flagd shuts down stream in `close/shutdown` function
- flagd emits appropriate events on flag changes, and errors
- associated testing

Blocked by:
- js-sdk implementation: https://github.com/open-feature/js-sdk/issues/424

toddbaert:
I'm starting on this.

PR #484, opened 2023-07-24
feat!: events, init, shutdown

toddbaert:
This PR:

- absorbs changes related to updating our proto generation plugin
- adds change events when change messages received from flagd
- adds shutdown

:warning: depends on https://github.com/open-feature/schemas/pull/99 (must be merged first). Will keep this as draft until then.

Fixes: https://github.com/open-feature/js-sdk-contrib/issues/382

### DIFF
--- a/libs/providers/flagd/package.json
+++ b/libs/providers/flagd/package.json
@@ -7,6 +7,6 @@
   },
   "peerDependencies": {
     "@grpc/grpc-js": "^1.6.0",
-    "@openfeature/js-sdk": "^1.1.0"
+    "@openfeature/js-sdk": ">=1.3.0"
   }
 }
--- a/libs/providers/flagd/src/lib/flagd-provider.ts
+++ b/libs/providers/flagd/src/lib/flagd-provider.ts
@@ -29,10 +29,17 @@ export class FlagdProvider implements Provider {
   private _status = ProviderStatus.NOT_READY;
   private _events = new OpenFeatureEventEmitter();
 
+  /**
+   * Construct a new flagd provider.
+   * 
+   * @param options options, see {@link FlagdProviderOptions}
+   * @param logger optional logger, see {@link Logger}
+   * @param service optional internal service implementation, should not be needed for production
+   */
   constructor(
     options?: FlagdProviderOptions,
-    service?: Service,
     private logger?: Logger,
+    service?: Service,
   ) {
     this._service = service ? service : new GRPCService(getConfig(options), undefined, logger);
   }
--- a/libs/providers/flagd/src/lib/flagd-provider.ts
+++ b/libs/providers/flagd/src/lib/flagd-provider.ts
@@ -1,4 +1,13 @@
-import { EvaluationContext, JsonValue, Logger, OpenFeatureEventEmitter, Provider, ProviderEvents, ProviderStatus, ResolutionDetails } from '@openfeature/js-sdk';
+import {
+  EvaluationContext,
+  JsonValue,
+  Logger,
+  OpenFeatureEventEmitter,
+  Provider,
+  ProviderEvents,
+  ProviderStatus,
+  ResolutionDetails,
+} from '@openfeature/js-sdk';
 import { FlagdProviderOptions, getConfig } from './configuration';
 import { GRPCService } from './service/grpc/service';
 import { Service } from './service/service';
@@ -17,34 +26,41 @@ export class FlagdProvider implements Provider {
   }
 
   private readonly _service: Service;
-  private _status = ProviderStatus.NOT_READY; 
+  private _status = ProviderStatus.NOT_READY;
   private _events = new OpenFeatureEventEmitter();
 
-  constructor(options?: FlagdProviderOptions, service?: Service, private logger?: Logger) {
+  constructor(
+    options?: FlagdProviderOptions,
+    service?: Service,
+    private logger?: Logger,
+  ) {
     this._service = service ? service : new GRPCService(getConfig(options), undefined, logger);
   }
 
   initialize(): Promise<void> {
-    return this._service.connect(this.setReady.bind(this), this.emitChanged.bind(this), this.setError.bind(this)).then(() => {
-      this.logger?.debug(`${this.metadata.name}: ready`);
-      this._status = ProviderStatus.READY;
-    }).catch((err) => {
-      this._status = ProviderStatus.ERROR;
-      this.logger?.error(`${this.metadata.name}: error during initialization: ${err.message}, ${err.stack}`);
-      throw err;
-    });
+    return this._service
+      .connect(this.setReady.bind(this), this.emitChanged.bind(this), this.setError.bind(this))
+      .then(() => {
+        this.logger?.debug(`${this.metadata.name}: ready`);
+        this._status = ProviderStatus.READY;
+      })
+      .catch((err) => {
+        this._status = ProviderStatus.ERROR;
+        this.logger?.error(`${this.metadata.name}: error during initialization: ${err.message}, ${err.stack}`);
+        throw err;
+      });
   }
 
   onClose(): Promise<void> {
-    this.logger?.debug(`${this.metadata.name}: shutting down`)
-    return this._service.disconnect();  
+    this.logger?.debug(`${this.metadata.name}: shutting down`);
+    return this._service.disconnect();
   }
 
   resolveBooleanEvaluation(
     flagKey: string,
     _: boolean,
     transformedContext: EvaluationContext,
-    logger: Logger
+    logger: Logger,
   ): Promise<ResolutionDetails<boolean>> {
     return this._service
       .resolveBoolean(flagKey, transformedContext, logger)
@@ -55,7 +71,7 @@ export class FlagdProvider implements Provider {
     flagKey: string,
     _: string,
     transformedContext: EvaluationContext,
-    logger: Logger
+    logger: Logger,
   ): Promise<ResolutionDetails<string>> {
     return this._service
       .resolveString(flagKey, transformedContext, logger)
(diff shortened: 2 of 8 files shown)


libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -245,50 +272,37 @@ export class GRPCService implements Service {
...
-      return Struct.fromJsonString(JSON.stringify(context));
-    } catch (e) {
-      const message = 'Error serializing context.';
-      const error = e as Error;
-      logger.error(`${message}: ${error?.message}`);
-      logger.error(error?.stack);
-      throw new ParseError(message);
-    }
-  }
-
```
toddbaert, line 278: we dont need to do any conext parsing with this new proto plugin :sunglasses:

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -78,40 +75,52 @@ export class GRPCService implements Service {
...
+  connect(
+    connectCallback: () => void,
+    changedCallback: (flagsChanged: string[]) => void,
+    disconnectCallback: () => void,
+  ): Promise<void> {
+    return this.connectStream(connectCallback, changedCallback, disconnectCallback);
+  }
+
+  disconnect(): Promise<void> {
+    this._stream?.destroy();
```
toddbaert, line 116: The old generated protobuf package had no `cancel` or `destroy`. I literally couldn't figure out how to kill the stream.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -127,56 +136,70 @@ export class GRPCService implements Service {
...
       const stream = this._client.eventStream({});
-      stream.responses.onError(() => {
-        this.handleError(reject);
+      stream.on('error', () => {
+        this.handleError(reject, connectCallback, changedCallback, disconnectCallback);
       });
-      stream.responses.onComplete(() => {
-        this.handleComplete();
+      stream.on('close', () => {
+        this.handleClose();
```
toddbaert, line 156: I renamed this function to match the handler name.

libs/providers/flagd/package.json
```diff
@@ -7,6 +7,6 @@
   },
   "peerDependencies": {
     "@grpc/grpc-js": "^1.6.0",
-    "@openfeature/js-sdk": "^1.1.0"
+    "@openfeature/js-sdk": ">=1.3.0"
```
toddbaert, line 10: Require 1.3.0+. @beeme1mr we could actually specify the recent 1.4.0... but technically, 1.3.0 will work.

libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -8,24 +8,38 @@ export class FlagdProvider implements Provider {
...
 
+  private readonly _service: Service;
+  private _status = ProviderStatus.NOT_READY; 
+  private _events = new OpenFeatureEventEmitter();
+
   constructor(options?: FlagdProviderOptions, service?: Service, private logger?: Logger) {
     this._service = service ? service : new GRPCService(getConfig(options), undefined, logger);
   }
 
+  initialize(): Promise<void> {
```
toddbaert, line 27: @beeme1mr the `service` argument here is only used for testing. I think we should move this to the last arg, and move the logger up, since this is already a breaking change.

I will also add some in-line doc.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -186,7 +209,7 @@ export class GRPCService implements Service {
       this._streamConnectAttempt++;
       setTimeout(() => {
         this._streamConnectBackoff = this._streamConnectBackoff * 2;
-        this.connectStream();
+        this.connectStream(connectCallback, changedCallback, disconnectCallback);
```
beeme1mr, line 212: Should this be awaited?

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -196,47 +219,51 @@ export class GRPCService implements Service {
...
-  private async resolve<T extends FlagValue, Rq extends AnyRequest, Rs extends AnyResponse>(
-    resolver: (request: AnyRequest) => UnaryCall<Rq, Rs>,
+  private async resolve<T extends FlagValue>(
+    promise: (
+      request: AnyRequest,
+      callback: (error: ServiceError | null, response: AnyResponse) => void,
+    ) => ClientUnaryCall,
     flagKey: string,
     context: EvaluationContext,
     logger: Logger,
```
beeme1mr, line 263: Logger isn't used

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -196,47 +219,51 @@ export class GRPCService implements Service {
...
-    parser?: (value: any) => any
+    parser?: (value: any) => any,
```
beeme1mr, line 264: Can we avoid using `any` here?

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: I don't see why, the function just has side effects, and it's set on a timeout.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -78,40 +75,52 @@ export class GRPCService implements Service {
...
+    connectCallback: () => void,
+    changedCallback: (flagsChanged: string[]) => void,
+    disconnectCallback: () => void,
+  ): Promise<void> {
+    return this.connectStream(connectCallback, changedCallback, disconnectCallback);
+  }
+
+  disconnect(): Promise<void> {
+    this._stream?.destroy();
+    return Promise.resolve();
```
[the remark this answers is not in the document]
toddbaert: Personally I don't like seeing `async` markers in functions that don't `await`, but I don't really care much. `destroy` is synchronous, so there's nothing to await.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: Probably.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -1,4 +1,4 @@
-import * as grpc from '@grpc/grpc-js';
+import { ClientReadableStream, ClientUnaryCall, ServiceError, credentials, status } from '@grpc/grpc-js';
```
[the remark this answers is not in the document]
toddbaert: Format has no impact here, I think the line needs to be longer for it to break. If I add more then it breaks.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: I added a debug statement. I could also remove the logger. Interested in your opinion on this once, since I guess you could always add a logger hook.

libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -1,4 +1,4 @@
-import { EvaluationContext, JsonValue, Logger, Provider, ResolutionDetails } from '@openfeature/js-sdk';
+import { EvaluationContext, JsonValue, Logger, OpenFeatureEventEmitter, Provider, ProviderEvents, ProviderStatus, ResolutionDetails } from '@openfeature/js-sdk';
```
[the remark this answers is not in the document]
toddbaert: done.

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -78,40 +75,52 @@ export class GRPCService implements Service {
...
       : new ServiceClient(
-          new GrpcTransport({
-            host: socketPath ? `unix://${socketPath}` : `${host}:${port}`,
-            channelCredentials: tls ? grpc.credentials.createSsl() : grpc.credentials.createInsecure(),
-          })
+          socketPath ? `unix://${socketPath}` : `${host}:${port}`,
+          tls ? credentials.createSsl() : credentials.createInsecure(),
         );
 
     // for now, we only need streaming if the cache is enabled (will need to be pulled out once we support events)
```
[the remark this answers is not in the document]
toddbaert: Good point.

libs/providers/flagd/src/lib/service/grpc/service.ts
toddbaert: I've done this, though a bit of casting was needed, I think it improves type-safety. It only goes so far though b/c the underlying generated gRPC code returns an `any`.

beeme1mr (approved):
Looks good! Please update the readme to include the events that this provider supports before you merge.

toddbaert:
I'm going to do one final smoke test of this (by running the js-sdk test suite with it) and then merge.

libs/providers/flagd/src/lib/flagd-provider.ts
```diff
@@ -8,29 +17,57 @@ export class FlagdProvider implements Provider {
...
+   * 
+   * @param options options, see {@link FlagdProviderOptions}
+   * @param logger optional logger, see {@link Logger}
+   * @param service optional internal service implementation, should not be needed for production
    */
-  get streamConnection() {
-    return this._service.streamConnection;
+  constructor(
+    options?: FlagdProviderOptions,
+    private logger?: Logger,
```
beeme1mr, line 41: Can this be read-only?

toddbaert: Probably, I'll take a look.

toddbaert: I've marked all of these `readonly`.

PR #493, opened 2023-07-27
fix: [GO Feature Flag server] Implement ProviderStatus

### LINKED ISSUE #489: [go-feature-flag] doesn't implement 'status'
The Go-Feature-Flag provider doesn't implement status: https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts

In order for it to stay functional after [this](https://github.com/open-feature/js-sdk/pull/507) is released, it needs to be updated to set its `status` field (it should start with it `NOT_READY` and update it to `READY` once its `bgScheduler` interval is started.

thomaspoignant:
## This PR
- Implement ProviderStatus

Fixes #489

### DIFF
--- a/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
+++ b/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
@@ -69,7 +69,6 @@ export class GoFeatureFlagProvider implements Provider {
   // logger is the Open Feature logger to use
   private logger?: Logger;
 
-  // status of the provider
   private _status: ProviderStatus = ProviderStatus.NOT_READY;
 
   constructor(options: GoFeatureFlagProviderOptions, logger?: Logger) {
--- a/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
+++ b/libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
@@ -5,6 +5,7 @@ import {
   JsonValue,
   Logger,
   Provider,
+  ProviderStatus,
   ResolutionDetails,
   StandardResolutionReasons,
   TypeMismatchError,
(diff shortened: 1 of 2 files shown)


libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
```diff
@@ -68,6 +69,9 @@ export class GoFeatureFlagProvider implements Provider {
   // logger is the Open Feature logger to use
   private logger?: Logger;
 
+  // status of the provider
```
beeme1mr, line 72: It's up to you, but I would remove this comment. I think the code is self-explanatory.

beeme1mr (approved):
Approved with some minor feedback.

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
thomaspoignant, line 72:
```suggestion
```

toddbaert (commented):
lgtm as well

liran2000:
Minor comment, according to the [spec](https://openfeature.dev/specification/sections/providers#requirement-245), I think this indication is missing:
```
Requirement 2.4.5
The provider SHOULD indicate an error if flag resolution is attempted before the provider is ready.
```

thomaspoignant:
@liran2000 you are totally right, I've open this PR https://github.com/open-feature/js-sdk-contrib/pull/505 to fix that.

toddbaert:
> Minor comment, according to the [spec](https://openfeature.dev/specification/sections/providers#requirement-245), I think this indication is missing:
> 
> ```
> Requirement 2.4.5
> The provider SHOULD indicate an error if flag resolution is attempted before the provider is ready.
> ```

Nice catch @liran2000 . We actually missed this in flagd-js as well. I created and issue: https://github.com/open-feature/js-sdk-contrib/issues/517

PR #495, opened 2023-07-27
fix: issue with flagd not disconnecting

toddbaert:


libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -112,7 +112,9 @@ export class GRPCService implements Service {
   }
 
   async disconnect(): Promise<void> {
-    this._stream?.destroy();
+    // cancel the stream and close the connection
+    this._stream?.cancel();
+    this._client.close();
```
toddbaert, line 117: cancel AND close

libs/providers/flagd/src/lib/service/grpc/service.ts
```diff
@@ -146,9 +148,13 @@ export class GRPCService implements Service {
   ): Promise<void> {
     return new Promise((resolve, reject) => {
       this.logger?.debug(`${FlagdProvider.name}: connecting stream, attempt ${this._streamConnectAttempt}...`);
-      const stream = this._client.eventStream({});
-      stream.on('error', () => {
-        this.handleError(reject, connectCallback, changedCallback, disconnectCallback);
+      const stream = this._client.eventStream({}, {});
+      stream.on('error', (err: ServiceError | undefined) => {
+        if (err?.code === status.CANCELLED) {
+          this.logger?.debug(`${FlagdProvider.name}: stream cancelled, will not be re-established`);
```
toddbaert, line 154: Don't start a reconnect in the cancelled case, which puts us in a loop and hangs things because `close()` waits for the stream to be done.

package.json
```diff
@@ -40,7 +40,7 @@
     "@nx/rollup": "16.5.5",
     "@nx/web": "16.5.5",
     "@nx/workspace": "16.5.5",
-    "@openfeature/js-sdk": "^1.3.1",
+    "@openfeature/js-sdk": "^1.4.1",
```
beeme1mr, line 43: Did you intend to bump this version? If so, make sure to bump the minimum peer version [here](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flagd/package.json#L10).

beeme1mr (approved):

package.json
toddbaert: This is just for local dev, as discussed.

PR #505, opened 2023-07-31
fix: [GO Feature Flag server] handle PROVIDER_NOT_READY from the spec

thomaspoignant:
## This PR
- Return an error if we try to evaluate before the provider is ready.
- Rewrite the tests to use the SDK and not test only the provider.


thomaspoignant:
> I've found this testing style to be better for providers, though maybe some people prefer "unit"-level tests.

I'll prefer it too, that's what I've done for other providers but this one the 1st one and I haven't done it properly the 1st time :D

PR #580, opened 2023-09-12
chore: GO Feature Flag move datacollector to hook + change cache key

thomaspoignant:
## This PR
- Move the data collector call into a hook
- Change the cache key with the hash of the entire context.


toddbaert (approved):
Ya this is nice. It's a better separation of concerns IMHO!

PR #585, opened 2023-09-17
fix: GOFF fix issues in hook

thomaspoignant:
## This PR
- Missing check in the web provider
- default field had a wrong value for errors.


libs/providers/go-feature-flag/src/lib/data-collector-hook.ts
```diff
@@ -100,7 +103,7 @@ export class GoFeatureFlagDataCollectorHook implements Hook {
     evaluationDetails: EvaluationDetails<FlagValue>,
     hookHints?: HookHints
   ) {
-    if (evaluationDetails.reason !== StandardResolutionReasons.CACHED){
+    if (!this.collectUnCachedEvaluation && evaluationDetails.reason !== StandardResolutionReasons.CACHED){
```
beeme1mr, line 106: I'm having trouble understanding the intent here. Should this be an or? What's the expected behavior?

beeme1mr: Ok, I think I understand.

beeme1mr (approved):

PR #596, opened 2023-09-28
fix: packaging issues impacting babel/react

toddbaert:
* run nx migrations we missed
* update nx again (with new migrations)
* remove `"type": "module"` from all package.jsons (let nx handle it, this was the bug)
* ADDS ability to generate CLIENT packages (with new prompt)

---

I've tested this in:

- create-react app
- an esm node project
- a commonjs node project

Fixes: https://github.com/open-feature/js-sdk-contrib/issues/597


thomaspoignant (approved):

libs/providers/go-feature-flag-web/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@openfeature/go-feature-flag-web-provider",
   "version": "0.1.3",
-  "type": "commonjs",
```
toddbaert, line 4: This was the primary issue. I've removed it from all packages.

The latest NX automation seems to work great if we don't interfere with it in this way.

thomaspoignant (approved):

beeme1mr (approved):

toddbaert:
@beeme1mr updated generators/templates look good, thanks for fixing that.

Merging. I will release the goff web provider ASAP because it's broken.

PR #607, opened 2023-10-09
feat: add in memory provider with specs and context support

luizgribeiro:
## This PR

Implements the following features for `inMemoryProvider`:
- default value for flags
- reason for flag evaluation
- Context based evaluation

### Related Issues
It's part of [this issue](https://github.com/open-feature/js-sdk/issues/565) from js-sdk repo.

### Follow-up Tasks

- Move this provider to the `js-sdk` repo 

### How to test

Unit tests were also implemented to assure


beeme1mr:
Thanks! I'll check it out tomorrow.

libs/providers/in-memory/src/lib/in-memory-provider.spec.ts
```diff
@@ -1,158 +1,526 @@
...
+    it('resolves to variant value with reason target match if context is provided and flag spec has context evaluator', async () => {
+      const booleanFlagCtxSpec = {
+        'a-boolean-flag': {
+          variants: {
+            on: true,
+            off: false
+          },
+          disabled: false,
+          defaultVariant: "on",
+          contextEvaluator: (_context: unknown) => "off"
```
beeme1mr, line 92:
```suggestion
          contextEvaluator: (_: unknown) => "off"
```

libs/providers/in-memory/src/lib/in-memory-provider.spec.ts
```diff
@@ -1,158 +1,526 @@
...
+    it('resolves to variant value with reason target match if context is provided and flag spec has context evaluator', async () => {
+      const StringFlagCtxSpec = {
+        'a-string-flag': {
+          variants: {
+            on: itsOn,
+            off: itsOff
+          },
+          disabled: false,
+          defaultVariant: "on",
+          contextEvaluator: (_context: unknown) => "off"
```
beeme1mr, line 191:
```suggestion
          contextEvaluator: (_: unknown) => "off"
```

beeme1mr: You should be able to use the `EvaluationContext` type instead of unknown.

libs/providers/in-memory/src/lib/in-memory-provider.spec.ts
```diff
@@ -1,158 +1,526 @@
...
+    it('resolves to variant value with reason target match if context is provided and flag spec has context evaluator', async () => {
+      const numberFlagCtxSpec = {
+        'a-number-flag': {
+          variants: {
+            on: onNumber,
+            off: offNumber
+          },
+          disabled: false,
+          defaultVariant: "on",
+          contextEvaluator: (_context: unknown) => "off"
```
beeme1mr, line 290:
```suggestion
          contextEvaluator: (_: unknown) => "off"
```

libs/providers/in-memory/src/lib/in-memory-provider.spec.ts
```diff
@@ -1,158 +1,526 @@
...
+    it('resolves to variant value with reason target match if context is provided and flag spec has context evaluator', async () => {
+      const ObjectFlagCtxSpec = {
+        'a-object-flag': {
+          variants: {
+            on: onObject,
+            off: offObject
+          },
+          disabled: false,
+          defaultVariant: "on",
+          contextEvaluator: (_context: unknown) => "off"
```
beeme1mr, line 389:
```suggestion
          contextEvaluator: (_: unknown) => "off"
```

libs/providers/in-memory/src/lib/in-memory-provider.ts
```diff
@@ -32,51 +34,70 @@ export class InMemoryProvider implements Provider {
...
-    return this._flagConfiguration[flagKey];
   }
 }
 
-function staticResolution<U>(value: U): ResolutionDetails<U> {
-  return {
-    value,
-    reason: StandardResolutionReasons.STATIC,
-  };
+class DisabledError extends Error {
```
beeme1mr, line 98: Perhaps we should add this as a standard error type. @toddbaert What do you think?

libs/providers/in-memory/src/lib/in-memory-provider.ts
```diff
@@ -1,16 +1,18 @@
...
+  Logger,
+  EvaluationContext,
+  TargetingKeyMissingError,
+  FlagValueType,
 } from '@openfeature/js-sdk';
-import { FlagConfiguration } from './flag-configuration';
+import { FlagConfiguration, Flag } from './flag-configuration';
 
 export class InMemoryProvider implements Provider {
   public readonly events = new OpenFeatureEventEmitter();
```
beeme1mr, line 18:
```suggestion
  public readonly events = new OpenFeatureEventEmitter();
  public readonly runsOn = 'server';
```

This will add a runtime compatibility check once [v1.5.0](https://github.com/open-feature/js-sdk/pull/560) is released.

beeme1mr (commented):
This looks good but it's definitely a breaking change in its current state. We can mark this a breaking change by updating the PR title to start with `feat!:` and make sure to update the getting started docs right after this is released. Another option would be to maintain backward compatibility. Finally, we could move this directly into the SDK and deprive the provider.

libs/providers/in-memory/src/lib/flag-configuration.ts
```diff
@@ -1,2 +1,12 @@
...
+type Variants<T> = Record<string, T>;
+
+export type Flag = {
+  variants: Variants<boolean> | Variants<string> | Variants <number> | Variants<JsonValue>;
+  defaultVariant: string;
+  disabled: boolean;
+  contextEvaluator?: (ctx: EvaluationContext) => string;
+} ;
+
+export type FlagConfiguration = Record<string, Flag>;
```
toddbaert, line 12: makes sense to me!

toddbaert, line 12: Makes sense to me!

libs/providers/in-memory/src/lib/in-memory-provider.ts
```diff
@@ -32,51 +34,70 @@ export class InMemoryProvider implements Provider {
...
-
-    return staticResolution(flagValue);
   }
 
-  async resolveNumberEvaluation(flagKey: string): Promise<ResolutionDetails<number>> {
-    const flagValue = this.lookupFlagValueOrThrow(flagKey);
-
-    if (typeof flagValue !== 'number') {
-      throw new TypeMismatchError();
+  private lookupFlagValueOrThrow(flagKey: string, ctx?: EvaluationContext, logger?: unknown): ResolutionDetails<any>{
```
toddbaert, line 71:
```suggestion
  private lookupFlagValue(flagKey: string, ctx?: EvaluationContext, logger?: unknown): ResolutionDetails<any>{
```

I know you didn't add this method, but I think `orThrow` is a bit gratuitous here. I think it's something that need not be in the method name.

libs/providers/in-memory/src/lib/in-memory-provider.ts
```diff
@@ -32,51 +34,70 @@ export class InMemoryProvider implements Provider {
...
-  }
+    this.throwIfFlagDisabled(flagSpec);
+
+    const isContextEval = ctx && flagSpec?.contextEvaluator;
+    const variant = isContextEval ? flagSpec.contextEvaluator?.(ctx) : flagSpec.defaultVariant;
+
+    const value = variant && flagSpec?.variants[variant];
+
+    if (value === undefined) {
+      throw new TargetingKeyMissingError();
```
toddbaert, line 85: This is not a `TargetingKeyMissingError`. `TargetingKeyMissingError` is specifically to be used when a provider requires a targeting key but one is not supplied in the context (see https://openfeature.dev/specification/types#error-code).

If the specified variant is missing, this basically means that the flag spec is invalid (the resolved variant doesn't map to a value). I would say this is a `GeneralError` (you could subclass GeneralError and throw a `VariantMissingError` you made up, if you want). You could also make the argument this is a `ParseError`. I'd be fine with either, but `TargetingKeyMissingError` seems wrong.

toddbaert (changes requested):
@luizgribeiro This looks great overall! Thanks so much! I would love to see something like this in the SDK.

I have 2 concerns about error handling:

- https://github.com/open-feature/js-sdk-contrib/pull/607/files#r1350721608
- https://github.com/open-feature/js-sdk-contrib/pull/607/files#r1350716854

Lastly, if you could add lots of comments (especially in the `lookupFlagValue`) that would probably be very educational for other provider authors.

toddbaert:
> Finally, we could move this directly into the SDK and deprive the provider.

@luizgribeiro @beeme1mr I think this is the best path. We have important doc that references this provider that we'd have to update with these breaking changes. I think it would be easier to simply implement all of this in the js-sdk directly and then deprecate this package

Let me know if you disagree or there's some reason this is a bad idea.

luizgribeiro:
Alright. I'll make the ajustments pointed out and move it to the js-sdk repo openning a new PR and closing this one.

luizgribeiro:
I've just pushed the newer version. 

If that's ok, let me know where to check this package out (maybe [this path](https://github.com/open-feature/js-sdk/tree/main/packages/server)) and I'll open a new PR on the other repo.

beeme1mr:
Yeah, that's what I had in mind. There's a provider folder where you could add the in-memory provider.

PR #620, opened 2023-10-24
feat: Initial version of the flagd js core

### LINKED ISSUE #615: POC shared package in monorepo
See parent issue.

Kavindu-Dodan:
## This PR

Fixes #615 and introduce the initial version of the flagd js core.

JS core wraps flagd flag parsing and evaluation logic into a single package. It expose a simple contract to store flags and resolve them. This is intended to be use by concrete provider implementations with their own flag obtaining mechanisms (ex:- gRPC sync, http fetch, file source etc...)


libs/shared/flagd-js-core/README.md
```diff
@@ -0,0 +1,21 @@
+# flagd-js-core
+
+`flagd-js-core` contains the core logic of flagd in-process evaluation provider.
```
toddbaert, line 3:
```suggestion
`flagd-js-core` contains the core logic of a flagd [in-process evaluation](https://flagd.dev/architecture/#in-process-evaluation) provider.
```

libs/shared/flagd-js-core/README.md
```diff
@@ -0,0 +1,21 @@
...
+This package is intended to be used by concrete implementations of flagd in-process provider.
+
+## Usage
+
+`flagd-js-core` wraps a simple flagd feature flag storage and flag evaluation logic.
+
+To use this implementation, instantiate a `FlagdJSCore` and provider valid flagd flag configurations.
```
toddbaert, line 10:
```suggestion
To use this implementation, instantiate a `FlagdJSCore` and provide valid flagd flag configurations.
```

libs/shared/flagd-js-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,202 @@
...
+    return this._targetingString;
+  }
+
+  get variants(): Map<string, FlagValue> {
+    return this._variants;
+  }
+}
+
+// flagd flag definition schema
+export const flagdSchema = {
```
toddbaert, line 39: Can we pull this as a submodule from schemas? We can pull based on a tag, for versioning.

libs/shared/flagd-js-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,202 @@
+import {FlagValue} from '@openfeature/server-sdk';
```
toddbaert, line 1: We can't use any specific imports from `@openfeature/server-sdk` or `@openfeature/web-sdk`, since this would mean we can't use it in the other runtime.

Hypothetically, all the stuff you're using is actually coming from @openfeature/shared. This is the first use-case we've had of a thing external to the js-sdk that could use that as a dep directly. We could potentially start publishing it to address this...

cc @beeme1mr

beeme1mr: Yes, that makes sense to me but we should give the package a better name.

Kavindu-Dodan: Makes sense. Proposing packaging name `@openfeature/sdk-common`

toddbaert: Will open a new issue discussing this in the js-sdk.

libs/shared/flagd-js-core/src/lib/storage.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * The simple contract of the core logics storage layer.
+ */
+export interface Storage {
+  setConfigurations(cfg: string): void;
+
+  getFlag(key: string): FeatureFlag | undefined;
+}
+
+export class StorageImpl implements Storage {
```
toddbaert, line 13:
```suggestion
export class MemoryStorage implements Storage {
```

A bit of a nit, but I tend to dislike `Impl` names. I think `MemoryStorage` is a fair name, but there's probably others too, like `MapStorage` or something.

libs/shared/flagd-js-core/src/lib/storage.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor() {
+    this._flags = new Map<string, FeatureFlag>();
+  }
+
+  getFlag(key: string): FeatureFlag | undefined {
+    return this._flags.get(key);
+  }
+
+  setConfigurations(cfg: string): void {
```
toddbaert, line 24: I think we could just as easily make this object immutable, and instantiate a new instance anytime we get a new flagconfig, but I don't have a strong preference here.

libs/shared/flagd-js-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Flagd flag configuration structure.
+ */
+export class FeatureFlag {
+  private readonly _state: string;
+  private readonly _defaultVariant: string;
+  private readonly _variants: Map<string, FlagValue>;
+  private readonly _targetingString: string;
+
+  constructor(flag: any) {
```
toddbaert, line 14:
```suggestion
  constructor(flag: unknown) {
```
`unknown` and `any` seem similar on the surface, but they are in many respects the opposite. `any` lets you do anything without type assertions, `unknown` lets you do nothing without type assertions. The latter is generally safer.

toddbaert (approved):
This is a good first-pass to me. I've left lots of things to think about but don't see blockers.

I don't see anything that is wrong, or couldn't be refactored if we decide it's not optimal as we add features.

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: This is done. I had to enable `resolveJsonModule` for this and it works. Will update with the next commit.

libs/shared/flagd-js-core/src/lib/storage.ts
Kavindu-Dodan: Done

libs/shared/flagd-js-core/src/lib/storage.ts
Kavindu-Dodan: Given that I have a Java background, my intention here was to reuse the same `FlagdJSCore` instance. but I know things are different in Js world, so I don't have any strong preference. We can simplify and expose set of functions here if we need

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: Thank you for the explanation. I don't see any logic that goes into this class, so I think I will go ahead with the interface suggestion

Kavindu-Dodan: We need both the interface (to by type safe) and the class. The class is needed to parse variants and preserve targeting string

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: Thank you :)

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: Consider checking the new implementation - https://github.com/open-feature/js-sdk-contrib/pull/620/files#diff-6dad1c1a73438bc494e8c021c542754462aa2977828aaf2bd270927a8426ca41

libs/shared/flagd-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,44 @@
+import {FlagValue} from '@openfeature/server-sdk';
+
+/**
+ * Flagd flag configuration structure mapping to schema definition.
+ */
+export interface Flag {
+  state: string,
+  defaultVariant: string,
+  variants: any,
```
toddbaert, line 9:
```suggestion
  variants: {[key: string]: FlagValue},
```

I think this is a precisely correct type for this field. See [FlagValue](https://github.com/open-feature/js-sdk/blob/8e873a3d2b6b2ada0ce7913a00ee1988226a7e05/packages/shared/src/evaluation/evaluation.ts#L15) from the SDK.

libs/shared/flagd-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,44 @@
...
+  targeting: string
+}
+
+/**
+ * Flagd flag configuration structure for internal reference.
+ */
+export class FeatureFlag {
```
toddbaert, line 16: Ya I personally like this class and think it will make things read nicely.

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // flag exist check
+    const flag = this._storage.getFlag(flagKey);
+    if (flag === undefined) {
+      throw new FlagNotFoundError(`flag: ${flagKey} not found`);
+    }
+
+    // flag status check
+    if (flag.state === 'DISABLED') {
+      throw new GeneralError(`flag: ${flagKey} is disabled`);
```
toddbaert, line 88: IMO, disabled is not an error - we expect disabled flags to be evaluated, so an exception based control flow is an anti-pattern. This should just return the default with the `Reason` = `DISABLED`.

libs/shared/flagd-js-core/src/lib/feature-flag.ts
toddbaert: Checked it and I like it.

toddbaert (approved):
Love it so far, approving again. Watch out for [this](https://github.com/open-feature/js-sdk-contrib/pull/620/files#r1373124737).

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: Discussed separately at https://github.com/open-feature/js-sdk/issues/616

libs/shared/flagd-core/src/lib/feature-flag.ts
Kavindu-Dodan: yup, this works ❤️

libs/shared/flagd-core/src/lib/flagd-core.ts
Kavindu-Dodan: Good point, I will update this

PR #621, opened 2023-10-24
feat: WIP flagd in-process provider

Kavindu-Dodan:
## This PR

WIP and based on https://github.com/open-feature/js-sdk-contrib/pull/620


libs/shared/flagd-js-core/src/lib/flagd-js-core.ts
beeme1mr, line 1: I don't think we need `js` in the name. Perhaps something we could call it something like `flagd-common`, `flagd-shared`, or `flagd-evaluation-engine`.

libs/shared/flagd-js-core/package-lock.json
beeme1mr, line 1: You shouldn't have a package lock file here. Perhaps you installed the dependencies from this folder accidently.

libs/shared/flagd-js-core/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@openfeature/flagd-js-core",
+  "version": "0.0.1-alpha",
+  "dependencies": {
+    "@openfeature/server-sdk": ">=1.6.0",
```
beeme1mr, line 5: This could be a peer dependency.

libs/shared/flagd-js-core/src/lib/flagd-js-core.ts
```diff
@@ -0,0 +1,133 @@
...
+} from '@openfeature/server-sdk';
+
+/**
+ * Expose flag configuration setter and flag resolving methods.
+ */
+export class FlagdJSCore {
+  private _storage: Storage;
+
+  /**
+   * Construct with optional your own storage layer.
```
beeme1mr, line 18:
```suggestion
   * Optionally construct with your own storage layer.
```

libs/shared/flagd-js-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,202 @@
...
+    return this._targetingString;
+  }
+
+  get variants(): Map<string, FlagValue> {
+    return this._variants;
+  }
+}
+
+// flagd flag definition schema
+export const flagdSchema = {
```
beeme1mr, line 39: This should be its own file.

libs/shared/flagd-js-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Flagd flag configuration structure.
+ */
+export class FeatureFlag {
+  private readonly _state: string;
+  private readonly _defaultVariant: string;
+  private readonly _variants: Map<string, FlagValue>;
+  private readonly _targetingString: string;
+
+  constructor(flag: any) {
```
beeme1mr, line 14: Please replace `any` with a type that represents a flagd configuration.

libs/shared/flagd-js-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,202 @@
...
+    this._state = flag['state'];
+    this._defaultVariant = flag['defaultVariant'];
+    this._variants = new Map<string, FlagValue>(Object.entries(flag['variants']));
+    this._targetingString = JSON.stringify(flag['targeting']);
```
beeme1mr, line 18: Why are you storing this as a string?

libs/shared/flagd-js-core/src/lib/feature-flag.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Flagd flag configuration structure.
+ */
+export class FeatureFlag {
+  private readonly _state: string;
+  private readonly _defaultVariant: string;
+  private readonly _variants: Map<string, FlagValue>;
+  private readonly _targetingString: string;
+
+  constructor(flag: any) {
+    this._state = flag['state'];
```
beeme1mr, line 15:
```suggestion
    this._state = flag.state;
```

This will be safer once it's not an `any` type.

libs/shared/flagd-js-core/src/lib/flagd-js-core.ts
```diff
@@ -0,0 +1,133 @@
...
+        errorMessage: `flag: ${flagKey} not found`,
+        reason: StandardResolutionReasons.ERROR,
+      };
+    }
+
+    // flag status
+    if (flag.state == 'DISABLED') {
+      return {
+        value: defaultValue,
+        errorCode: ErrorCode.GENERAL,
```
beeme1mr, line 89: Should this be a `FLAG_NOT_FOUND` error?

libs/shared/flagd-js-core/src/lib/flagd-js-core.ts
```diff
@@ -0,0 +1,133 @@
...
+    if (!guard(variant)) {
+      return {
+        value: defaultValue,
+        errorCode: ErrorCode.TYPE_MISMATCH,
+        errorMessage: `returning default variant for flagKey: ${flagKey}, type not valid`,
+        reason: StandardResolutionReasons.ERROR,
+      };
+    }
+
+    return {
```
beeme1mr, line 128: Variant should be included in the response.

toddbaert:
I've left all my feedback on the dependency PR (looks like most of @beeme1mr 's relevant points are related to that as well) so please check there.

I think for the server case, we might be better off simply adding this functionality to the existing server provider as we did with Java, but I don't really feel to strongly about that since this isn't for the web and therefor bundle size isn't quite as important.

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: Changed to obtain this from flagd-schema (as proposed by @toddbaert )

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: Targeting handling is not implemented yet and string here is for simplicity

libs/shared/flagd-js-core/src/lib/flagd-js-core.ts
Kavindu-Dodan: Done, I updated package name to `flagd-core`. This is available in the upstream (base) PR of this change set - https://github.com/open-feature/js-sdk-contrib/pull/620

libs/shared/flagd-js-core/package-lock.json
Kavindu-Dodan: Thanks, removed upstream

libs/shared/flagd-js-core/src/lib/flagd-js-core.ts
Kavindu-Dodan: Applied upstream, will update this draft PR once upstream is ready

libs/shared/flagd-js-core/src/lib/feature-flag.ts
Kavindu-Dodan: Changed upstream, please check PR - https://github.com/open-feature/js-sdk-contrib/pull/620

libs/shared/flagd-js-core/src/lib/flagd-js-core.ts
Kavindu-Dodan: Introrudced errors. Updated in the upstream PR -  https://github.com/open-feature/js-sdk-contrib/pull/620

Issue #614, opened 2023-10-18
[flagd] in-process evaluation mode, and shared evaluation engine


toddbaert:
As mentioned in flagd documentation of  **Creating an in-process flagd provider** [^1], the in-process provider comes with unique advantages. This issue focuses on creating a TS/JS implementation. The implementation will be server-side, but shareable components (namely the evaluation engine, see tasks below) should be packaged and released separately, so they can be used in a hypothetical client side implementation, browser UI validation, etc.

**Requirements**

- Fulfill requirements outlined by flagd in-process guideline [^1]
- Minimal external dependencies 
- Updated/New ReadMe & usage guide
- Unit & e2e tests (e2e test already exist [here](https://github.com/open-feature/flagd-testbed/tree/main/gherkin)) 

**Technical requirements**

- Build both remote and in-process providers into a single dependency (To be decided after the POC)
- Reuse existing interfaces and implementations where possible 
- Shareable flagd evaluation engine lib published as independent module

**Tasks**

**_NOTE: each of these should be it's own issue. Please ping @toddbaert if you're interested in taking one of these tasks and I will create an issue on-demand._**

- [x] #615
- [x] evaluation engine package
- [x] #626
    - [x] flag definition validation using [the schema](https://github.com/open-feature/schemas/blob/main/json/flagd-definitions.json)
      - [x] [Shared evaluators](https://flagd.dev/reference/flag-definitions/#shared-evaluators)
      - [x] [injected properties](https://flagd.dev/reference/flag-definitions/#flagd-properties-in-the-evaluation-context)
    - [x] Custom JsonLogic evaluators 
      - [x] [Fractional](https://flagd.dev/reference/specifications/custom-operations/fractional-operation-spec/)
      - [x] [Semver](https://flagd.dev/reference/specifications/custom-operations/semver-operation-spec/)
      - [x] [String Comp](https://flagd.dev/reference/specifications/custom-operations/string-comparison-operation-spec/)
- [x] Serverside provider
  - [x] uses evaluation engine package above
  - [x] connects via gRPC sync to get flag definition

[^1]: https://flagd.dev/reference/specifications/in-process-providers/

Kavindu-Dodan:
`evaluation engine package` is also completed with PR https://github.com/open-feature/js-sdk-contrib/pull/620

I am now actively working on `Basic JSONLogic engine shared library`. This should complete the flagd-core part of the implementation

Kavindu-Dodan:
[...]

PR #623, opened 2023-10-27
feat: flagd-core json logic evaluator

Kavindu-Dodan:
## This PR

Completes the `Basic JSONLogic engine shared library` part of issue #614

Features matching the original issue. 

- [x] flag definition validation using [the schema](https://github.com/open-feature/schemas/blob/main/json/flagd-definitions.json)
- [x]  [Shared evaluators](https://flagd.dev/reference/flag-definitions/#shared-evaluators)
- [x] [injected properties](https://flagd.dev/reference/flag-definitions/#flagd-properties-in-the-evaluation-context)
- [x] Custom JsonLogic evaluators (string comp, sem ver & fractional)

Closes: https://github.com/open-feature/js-sdk-contrib/issues/626


libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,43 @@ export class FlagdCore {
...
 
-    const defaultVariant = flag.defaultVariant;
-
-    let resolvedVariant;
+    let variant;
     let reason;
 
-    if (flag.targetingString == undefined) {
-      resolvedVariant = flag.variants.get(defaultVariant);
+    if (flag.targeting == undefined) {
```
toddbaert, line 98:
```suggestion
    if (!flag.targeting) {
```

Mostly this is how people do this sort of check from a style perspective, but also keep in mind `null !== undefined`. Note that`null == undefined` though, so this probably works as you expect... you just tend to see the style above much more.

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,43 @@ export class FlagdCore {
...
-      reason = StandardResolutionReasons.STATIC;
+      let targetingResolution;
+      try {
+        targetingResolution = this._targeting.applyTargeting(flagKey, flag.targeting, evalCtx);
+      } catch (e) {
+        console.error(`Error evaluating targeting rule for flag ${flagKey}, falling back to default`, e);
+        targetingResolution = null;
+      }
+
+      if (targetingResolution == null) {
```
toddbaert, line 110: Same as above.

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,43 @@ export class FlagdCore {
...
+        variant = flag.defaultVariant;
+        reason = StandardResolutionReasons.DEFAULT;
+      } else {
+        variant = targetingResolution;
+        reason = StandardResolutionReasons.TARGETING_MATCH;
+      }
     }
 
-    // todo improve this error condition when targeting evaluation is ready
+    const resolvedVariant = flag.variants.get(variant)
```
toddbaert, line 119: recommend you change the check below this too.

libs/shared/flagd-core/src/lib/feature-flag.ts
```diff
@@ -17,13 +17,13 @@ export class FeatureFlag {
   private readonly _state: string;
   private readonly _defaultVariant: string;
   private readonly _variants: Map<string, FlagValue>;
-  private readonly _targetingString: string;
+  private readonly _targeting: any;
```
toddbaert, line 20:
```suggestion
  private readonly _targeting: unknown;
```

This is safer and since you're not accessing props on it, it should be easy to use this way (might need to add an assertion or two).

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,43 @@ export class FlagdCore {
...
-      throw new TypeMismatchError('evaluated type does not match the flag type');
+      throw new TypeMismatchError('Evaluated type does not match the flag type');
     }
 
-    // todo check variant updates from targeting
     return {
       value: resolvedVariant as T,
       reason: reason,
-      variant: defaultVariant
+      variant: variant
```
beeme1mr, line 131:
```suggestion
      variant,
```

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -0,0 +1,97 @@
...
+    const entry = from[i]
+    if (!Array.isArray(entry)) {
+      throw new Error("Invalid bucket entries");
+    }
+
+    if (entry.length != 2) {
+      throw new Error("Invalid bucketing entry. Require two values - variant and percentage");
+    }
+
+    if (typeof entry[0] != 'string') {
```
beeme1mr, line 80:
```suggestion
    if (typeof entry[0] !== 'string') {
```

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -0,0 +1,97 @@
...
+      throw new Error("Bucketing require variant to be present in string format");
+    }
+
+    if (typeof entry[1] != 'number') {
```
beeme1mr, line 84:
```suggestion
    if (typeof entry[1] !== 'number') {
```

libs/shared/flagd-core/src/lib/targeting/sem-ver.ts
```diff
@@ -0,0 +1,46 @@
...
+  const args = Array.from(data)
+
+  if (args.length != 3) {
+    return false;
+  }
+
+  const semVer1 = parse(args[0]);
+  const semVer2 = parse(args[2]);
+
+  if (semVer1 == null || semVer2 == null) {
```
beeme1mr, line 19:
```suggestion
  if (!semVer1 || !semVer2) {
```

libs/shared/flagd-core/src/lib/targeting/sem-ver.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const args = Array.from(data)
+
+  if (args.length != 3) {
+    return false;
+  }
+
+  const semVer1 = parse(args[0]);
+  const semVer2 = parse(args[2]);
```
beeme1mr, line 17: Can this parsing fail?

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -3,18 +3,19 @@ import {
...
   TypeMismatchError,
 } from '@openfeature/server-sdk';
+import {Targeting} from "./targeting/targeting";
 
 /**
  * Expose flag configuration setter and flag resolving methods.
  */
 export class FlagdCore {
   private _storage: Storage;
+  private _targeting = new Targeting()
```
toddbaert, line 18:
```suggestion
  private _targeting = new Targeting();
```

package.json
```diff
@@ -21,10 +21,13 @@
     "@protobuf-ts/grpc-transport": "^2.9.0",
     "@protobuf-ts/runtime-rpc": "^2.9.0",
     "@swc/helpers": "0.5.3",
+    "@types/imurmurhash": "^0.1.3",
```
beeme1mr, line 24: Should `imurmurhash` also be a dependency?

libs/shared/flagd-core/src/lib/targeting/targeting.ts
```diff
@@ -0,0 +1,31 @@
...
+    engine.addMethod(fractionalRule, fractional);
+
+    this._logicEngine = engine;
+  }
+
+  applyTargeting(flagKey: string, logic: any, data: Record<any, any>): any {
+    const ctxData = {
+      [flagdPropertyKey]: {
+        [flagKeyPropertyKey]: flagKey,
+        [timestampPropertyKey]: Date.now(),
```
toddbaert, line 24: This is ms, not seconds, I think you'd need to change it to be compliant.

libs/shared/flagd-core/src/lib/targeting/targeting.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    expect(targeting.applyTargeting("flagA", input, {targetingKey: "key"})).toBe(null)
+  })
+
+  it("should validate rule - buckets should be variant and fraction", () => {
+    const input = {
+      fractional: [
+        ["red", 50],
+        [100, 50]
+      ]
+    }
```
toddbaert, line 157: style: you could put all these in a "describe" func that is called "validation" and then just name them with what you've put after the `-`. This nesting is pretty idiomatic in JS tests.

libs/shared/flagd-core/src/lib/targeting/targeting.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+  it("should evaluate valid rule", () => {
+    const input = {
+      fractional: [
+        {"var": "key"},
+        ["red", 50],
+        ["blue", 50]
+      ]
+    }
+
+    expect(targeting.applyTargeting("flagA", input, {key: "bucketKeyA"})).toBe("red")
```
toddbaert, line 116: I think you should add another expect that ends up at "blue".

toddbaert:
Overall looks good. All my comments are minor and mostly style, except [this one](https://github.com/open-feature/js-sdk-contrib/pull/623#discussion_r1377715866) which is non-compliant AFAIK.

package.json
```diff
@@ -21,10 +21,13 @@
...
+    "@types/semver": "^7.5.4",
```
toddbaert, line 25: all `@types...` packages should be DEV dependencies.

libs/shared/flagd-core/src/lib/targeting/targeting.ts
```diff
@@ -0,0 +1,31 @@
...
+      },
+      ...data
+    }
```
beeme1mr, line 27:
```suggestion
      ...data,
      [flagdPropertyKey]: {
        [flagKeyPropertyKey]: flagKey,
        [timestampPropertyKey]: Date.now(),
      },
    }
```

https://flagd.dev/reference/specifications/in-process-providers/#flagd-properties-in-the-evaluation-context

libs/shared/flagd-core/src/lib/targeting/targeting.ts
```diff
@@ -0,0 +1,31 @@
...
+    const engine = new LogicEngine();
+    engine.addMethod(startsWithRule, startsWithHandler);
+    engine.addMethod(endsWithRule, endsWithHandler);
+    engine.addMethod(semVerRule, semVer);
+    engine.addMethod(fractionalRule, fractional);
+
+    this._logicEngine = engine;
+  }
+
+  applyTargeting(flagKey: string, logic: any, data: Record<any, any>): any {
```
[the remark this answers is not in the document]
Kavindu-Dodan: This is possible up to some level. I need to use `data: Record<any, unknown>` as I have to spread the record and combine it with other custom flagd properties.

libs/shared/flagd-core/src/lib/targeting/targeting.ts
beeme1mr: Please also add a test to make sure the `$flagd` properties override anything supplied by the user.

libs/shared/flagd-core/src/lib/targeting/sem-ver.ts
Kavindu-Dodan: A `null` is returned if parsing failed, which we check in our logic

libs/shared/flagd-core/src/lib/targeting/targeting.ts
Kavindu-Dodan: Thank you. this is a good catch 🙏

libs/shared/flagd-core/src/lib/targeting/targeting.spec.ts
Kavindu-Dodan: Good suggestion, will do this

package.json
Kavindu-Dodan: @toddbaert any reason for this ? 🤔

libs/shared/flagd-core/src/lib/targeting/targeting.ts
Kavindu-Dodan: Is it the ordering of properties ? if so does it matter ? 🤔

libs/shared/flagd-core/src/lib/flagd-core.ts
Kavindu-Dodan: This is done for all usages hence will resolve this convo

libs/shared/flagd-core/src/lib/feature-flag.ts
Kavindu-Dodan: Done, changed to unkown

package.json
Kavindu-Dodan: Adde to module as a dependency

package.json
Kavindu-Dodan: This is done anyway

toddbaert: This is because types are only used at compile-time, not runtime. Anything that is only used at compile-time is a devDependency - there's no need to ship them or force consumers to download them.

libs/shared/flagd-core/src/lib/targeting/targeting.ts
toddbaert: The ordering impacts which one "wins" in cases of duplicate keys. In your original implementation, the data would overwrite the injected properties if it had `$flagd` in it. You should check do see if the prop is incoming, and warn as we do in Go that the `$flagd` prop is special and will be overwritten.

toddbaert (approved):
Looks good to me, at least good enough that I think it's worth trying to put it in a provider and trying running the e2e tests at least!

libs/shared/flagd-core/src/lib/targeting/targeting.ts
Kavindu-Dodan: thanks for the details. I will add the check and warning

libs/shared/flagd-core/src/lib/targeting/targeting.spec.ts
Kavindu-Dodan: this is done

libs/shared/flagd-core/src/lib/targeting/targeting.ts
Kavindu-Dodan: Done. Added the warning and validation tests - https://github.com/open-feature/js-sdk-contrib/pull/623/commits/dd9025fabd09be28e0a23c8f4edf0b83316dcba2

toddbaert (approved):

libs/shared/flagd-core/package.json
```diff
@@ -1,12 +1,13 @@
...
   "name": "@openfeature/flagd-core",
-  "version": "0.0.1-alpha",
+  "version": "0.1.0",
   "scripts": {
     "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
     "current-version": "echo $npm_package_version"
   },
   "dependencies": {
-    "@openfeature/server-sdk": ">=1.6.0",
+    "@types/imurmurhash": "^0.1.3",
```
beeme1mr, line 9: I think this should be a dev dependency and `imurmurhash` should be a dependency.

FYI @toddbaert

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,47 @@ export class FlagdCore {
...
     } else {
-      // todo - targeting evaluation
-      // till targeting is handled, return the static result
-      resolvedVariant = flag.variants.get(defaultVariant);
-      reason = StandardResolutionReasons.STATIC;
+      let targetingResolution;
+      try {
+        targetingResolution = this._targeting.applyTargeting(flagKey, flag.targeting, evalCtx);
+      } catch (e) {
+        console.error(`Error evaluating targeting rule for flag ${flagKey}, falling back to default`, e);
```
beeme1mr, line 106: Can we follow the safe logger pattern used in the SDK? That could be handled in a follow up task.

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,47 @@ export class FlagdCore {
...
+        targetingResolution = null;
+      }
+
+      if (!targetingResolution) {
+        variant = flag.defaultVariant;
+        reason = StandardResolutionReasons.DEFAULT;
+      } else {
+        variant = targetingResolution;
+        reason = StandardResolutionReasons.TARGETING_MATCH;
```
beeme1mr, line 115: Do we have any way of detecting if the reason is SPLIT?

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,47 @@ export class FlagdCore {
...
+      }
+    }
+
+    if (typeof variant !== 'string') {
```
beeme1mr, line 119: flagd casts the variant to a string. That allows for the shorthand JSON logic implementation like `{"==" : [1, 1]}` to match the `true` variant.

beeme1mr: We should make sure this is explicitly defined in the spec and tested since it's a common pattern in flagd.

FYI @toddbaert

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,47 @@ export class FlagdCore {
...
+      throw new TypeMismatchError('Variant must be a string, but found ' + typeof variant)
     }
 
-    // todo improve this error condition when targeting evaluation is ready
-    if (resolvedVariant === undefined) {
-      throw new TypeMismatchError(`variant ${defaultVariant} not found in flag with key ${flagKey}`);
+    const resolvedVariant = flag.variants.get(variant)
+    if (!resolvedVariant) {
+      throw new TypeMismatchError(`Variant ${variant} not found in flag with key ${flagKey}`);
```
beeme1mr, line 125: Should this be a type mismatch? Perhaps a general error would be more appropriate.

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,47 @@ export class FlagdCore {
...
     }
 
     if (!guard(resolvedVariant)) {
-      throw new TypeMismatchError('evaluated type does not match the flag type');
+      throw new TypeMismatchError('Evaluated type does not match the flag type');
```
beeme1mr, line 129: It may be challenging to troubleshoot the problem with this message.

Could you please extend it to include information such as:
 - flag key
 - expected type
 - actual type

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -91,35 +92,47 @@ export class FlagdCore {
...
     }
 
-    // todo check variant updates from targeting
     return {
       value: resolvedVariant as T,
       reason: reason,
```
beeme1mr, line 134:
```suggestion
      reason,
```

Because you can

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -0,0 +1,97 @@
+import {flagdPropertyKey, flagKeyPropertyKey, targetingPropertyKey} from "./common";
+import MurmurHash3 from "imurmurhash";
+
+export const fractionalRule = "fractional";
+
+export function fractional(data: unknown, context: Record<any, any>): string | null {
```
beeme1mr, line 6: Can the record be `Record<unknown, unknown>`? If not, can you add a comment to have the linter ignore this?

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -0,0 +1,97 @@
...
+  if (!Array.isArray(data)) {
+    return null;
+  }
+
+  const args = Array.from(data);
+  if (args.length < 2) {
+    console.error('Invalid targeting rule. Require at least two buckets.')
```
beeme1mr, line 13: Same comment about the logger.

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -0,0 +1,97 @@
...
+  let bucketBy: string;
+  let buckets: unknown[];
+
+  if (typeof args[0] == "string") {
+    bucketBy = args[0];
+    buckets = args.slice(1, args.length)
+  } else {
+    bucketBy = context[targetingPropertyKey];
+    if (!bucketBy) {
+      return null;
```
beeme1mr, line 31: Should we log before returning? It seems like this condition would only happen if something very unexpected happened.

package.json
beeme1mr, line 1: This also seems to be missing `imurmurhash` as a dependency.

libs/shared/flagd-core/src/lib/feature-flag.ts
```diff
@@ -1,12 +1,12 @@
...
 
 /**
  * Flagd flag configuration structure mapping to schema definition.
  */
 export interface Flag {
   state: string,
   defaultVariant: string,
-  variants: {[key: string]: FlagValue},
+  variants: { [key: string]: FlagValue },
   targeting: string
```
beeme1mr, line 10:
```suggestion
  targeting?: string;
```

Targeting should be optional.

libs/shared/flagd-core/src/lib/feature-flag.ts
```diff
@@ -1,12 +1,12 @@
-import {FlagValue} from '@openfeature/server-sdk';
+import {FlagValue} from '@openfeature/core';
 
 /**
  * Flagd flag configuration structure mapping to schema definition.
  */
 export interface Flag {
   state: string,
   defaultVariant: string,
```
beeme1mr, line 8:
```suggestion
  defaultVariant: string;
```

libs/shared/flagd-core/src/lib/feature-flag.ts
```diff
@@ -1,12 +1,12 @@
...
-  variants: {[key: string]: FlagValue},
+  variants: { [key: string]: FlagValue },
```
beeme1mr, line 9:
```suggestion
  variants: { [key: string]: FlagValue };
```

libs/shared/flagd-core/src/lib/flagd-core.ts
toddbaert: Do you mean the variant key? Anything but a string would be invalid JSON, and if you mean the `defaultVariant`, the schema doesn't allow it to be anything but a string

libs/shared/flagd-core/src/lib/flagd-core.ts
Kavindu-Dodan: we don't detect this even at flagd. And I don't see a clear cut path to enable this 🤔

beeme1mr: Yeah, that's what I thought. Thanks.

libs/shared/flagd-core/src/lib/flagd-core.ts
toddbaert: I understand now, this is the targeting result. I think this is worth specifying. Could you open an issue in flagd?

package.json
Kavindu-Dodan: It's there as a dev dep  - https://github.com/open-feature/js-sdk-contrib/blob/0dd6165e2e7d286ccfbc195c25839d5cda9b52c4/package.json#L54

toddbaert: That's just the type, you'd need just `imurmurhash` for the runtime components.

libs/shared/flagd-core/src/lib/flagd-core.ts
Kavindu-Dodan: I added special toString() conversion for targeting resolving. Plus added a test to validate the short circuiting rule

libs/shared/flagd-core/src/lib/targeting/fractional.ts
Kavindu-Dodan: No this is not possible as `Record` expect any type. I disabled the lint check on this line

libs/shared/flagd-core/src/lib/targeting/fractional.ts
Kavindu-Dodan: I addressed this in the evaluator end. But adding this to targeting handlers is not that straightforward as we have to deal with method signatures of json logic engine

package.json
Kavindu-Dodan: This is resolved

libs/shared/flagd-core/src/lib/flagd-core.ts
Kavindu-Dodan: this is done for the evaluator contract

libs/shared/flagd-core/src/lib/flagd-core.ts
Kavindu-Dodan: Yes, converted to a general error

libs/shared/flagd-core/src/lib/targeting/fractional.ts
Kavindu-Dodan: Added a log and yes this is if `targetingKey` is missing as a fallback option

libs/shared/flagd-core/src/lib/flagd-core.ts
Kavindu-Dodan: Added suggested details to the error message

libs/shared/flagd-core/package.json
```diff
@@ -1,13 +1,16 @@
...
     "current-version": "echo $npm_package_version"
   },
   "dependencies": {
-    "@openfeature/server-sdk": ">=1.6.0",
+    "@openfeature/core": "^0.0.16",
     "ajv": "^8.12.0",
     "tslib": "^2.3.0"
+  },
+  "devDependencies": {
+    "@types/imurmurhash": "^0.1.3"
```
toddbaert, line 14:
```suggestion
  },
  "devDependencies": {
    "@types/imurmurhash": "^0.1.3"
```

This will be computed automatically by nx

libs/shared/flagd-core/package.json
```diff
@@ -1,13 +1,16 @@
...
   "name": "@openfeature/flagd-core",
-  "version": "0.0.1-alpha",
+  "version": "0.1.0",
   "scripts": {
     "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
     "current-version": "echo $npm_package_version"
   },
   "dependencies": {
-    "@openfeature/server-sdk": ">=1.6.0",
+    "@openfeature/core": "^0.0.16",
```
toddbaert, line 9: I would instead set this as a minimum peer dependency:

```
peerDependencies: {
  "@openfeature/core": ">=0.0.16"
}
```

package.json
```diff
@@ -16,6 +16,7 @@
     "@bufbuild/connect-web": "^0.13.0",
     "@bufbuild/protobuf": "^1.2.0",
     "@grpc/grpc-js": "^1.8.20",
+    "@openfeature/core": "^0.0.16",
     "@openfeature/web-sdk": "^0.4.0",
```
toddbaert, line 20:
```suggestion
```

Please move both of these to dev dependencies. (I realize one was not you, but it's confusing anyway so lets fix it).

libs/shared/flagd-core/package.json
Kavindu-Dodan: removed

libs/shared/flagd-core/package.json
Kavindu-Dodan: done

package.json
Kavindu-Dodan: done.

PR #651, opened 2023-11-14
fix: false positive of a falsy variant value check

beeme1mr:
## This PR

Prevents flagd core from throwing if the variant value is a falsy value.

### DIFF
--- a/libs/shared/flagd-core/src/lib/flagd-core.ts
+++ b/libs/shared/flagd-core/src/lib/flagd-core.ts
@@ -120,7 +120,7 @@ export class FlagdCore {
     }
 
     const resolvedVariant = flag.variants.get(variant);
-    if (!resolvedVariant) {
+    if (resolvedVariant === undefined) {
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan (approved):

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -120,7 +120,7 @@ export class FlagdCore {
     }
 
     const resolvedVariant = flag.variants.get(variant);
-    if (!resolvedVariant) {
+    if (resolvedVariant === undefined) {
```
toddbaert, line 123: ooof, one of the few times `!someVar` isn't the correct choice :sob:

toddbaert: is `null` a valid variant?

beeme1mr: That should be handled by the type checker that's run immediately after this section.

toddbaert (approved):

PR #654, opened 2023-11-16
fix: orphaned grpc connection, semver ~, change events

toddbaert:
This fixes 3 small bugs and adds e2e tests:

Small bugs:

- sem_ver `~` operator matched if major version was different
- change handler was not fired on changes
- grpc connection was reconnected if stream cancelled from client


beeme1mr (approved):

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -62,27 +65,33 @@ export class GrpcFetch implements DataFetch {
...
-      this.reconnectWithBackoff(resolveConnect, dataFillCallback, connectCallback, disconnectCallback);
-    });
-
-    this._syncStream.on('end', () => {
-      this._logger?.error('Stream ended, attempting to reconnect');
-      disconnectCallback();
-      this.reconnectWithBackoff(resolveConnect, dataFillCallback, connectCallback, disconnectCallback);
+    this._syncStream.on('error', (err: ServiceError | undefined) => {
+      if (err?.code === status.CANCELLED) {
+        this._logger?.debug('Stream cancelled, will not be re-established');
```
toddbaert, line 75: When we call stream.cancel in JS, it fires a an error, which is caught by the on error handler, so we get in an infinite reconnect loop event when we try to exit.

libs/shared/flagd-core/src/lib/targeting/sem-ver.ts
```diff
@@ -39,7 +39,7 @@ export function semVer(data: unknown): boolean {
     case '^':
       return semVer1.major == semVer2.major;
     case '~':
-      return semVer1.minor == semVer2.minor;
+      return semVer1.major == semVer2.major && semVer1.minor == semVer2.minor;
```
toddbaert, line 42: This is correct in Java, it just looks like a condition was missing here.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -62,27 +65,33 @@ export class GrpcFetch implements DataFetch {
       this._logger?.debug('Received sync payload');
       dataFillCallback(data.flagConfiguration);
       connectCallback();
+      changedCallback([]); // flags changed list not supported
```
toddbaert, line 68: We weren't running a change handler, so though we got data, we never emitted flags changed.

Kavindu-Dodan (approved):

PR #660, opened 2023-11-20
fix!:  reconnect, missing and duped events, remove max reconnect

toddbaert:
- simplifies reconnect logic by watching the channel state (as recommended by gRPC js maintainers)
- fires error events on errors
- fixes an issue causing duplicated READY events
- removes the max-reconnect option
- adds e2e tests for reconnection


libs/providers/flagd/src/e2e/setup-in-process-provider.ts
```diff
@@ -7,10 +7,13 @@ const FLAGD_NAME = 'flagd Provider';
...
-assert(
-  OpenFeature.providerMetadata.name === FLAGD_NAME,
-  new Error(`Expected ${FLAGD_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`),
-);
+OpenFeature.setProvider('unstable', new FlagdProvider({ resolverType: 'in-process', host: 'localhost', port: 9091 }));
+// TODO: update with correct assertions once we have ability to get providerMetadata for any provider
+// assert(
+//   OpenFeature.providerMetadata.name === FLAGD_NAME,
+//   new Error(`Expected ${FLAGD_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`),
+// );
```
toddbaert, line 18: So we could run these tests in parallel, I started using a dedicated named provider `'e2e'` for this, but when I did that, I realized we can only query the metadata of the default provider. We need to add a feature to get named provider metadata to the js-sdk then fix this.

libs/providers/flagd/src/e2e/setup-rpc-provider.ts
```diff
@@ -6,9 +6,11 @@ const FLAGD_NAME = 'flagd Provider';
...
 // register the flagd provider before the tests.
 console.log('Setting flagd provider...');
-OpenFeature.setProvider(new FlagdProvider({ cache: 'disabled' }));
-assert(
-  OpenFeature.providerMetadata.name === FLAGD_NAME,
-  new Error(`Expected ${FLAGD_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`),
-);
+OpenFeature.setProvider('e2e', new FlagdProvider({ cache: 'disabled' }));
+OpenFeature.setProvider('unstable', new FlagdProvider({ cache: 'disabled', port: 8014 }));
+// TODO: update with correct assertions once we have ability to get providerMetadata for any provider
```
toddbaert, line 11: same as https://github.com/open-feature/js-sdk-contrib/pull/660/files#r1399628600

libs/providers/flagd/src/e2e/step-definitions/flagd-reconnect.unstable.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { OpenFeature, ProviderEvents } from '@openfeature/server-sdk';
+import { defineFeature, loadFeature } from 'jest-cucumber';
+
+jest.setTimeout(30000);
+
+// load the feature file.
+const feature = loadFeature('features/flagd-reconnect.feature');
```
toddbaert, line 7: New reconnect e2e test!

libs/providers/flagd/src/lib/flagd-provider.spec.ts
```diff
@@ -485,73 +494,23 @@ describe(FlagdProvider.name, () => {
     });
 
     describe('connection/re-connection', () => {
-      it('should attempt to inital connection multiple times', (done) => {
+      it('should watch channel for reconnect after error', () => {
```
toddbaert, line 497: This test was simplified with our new connection mechanism

libs/providers/flagd/src/lib/service/grpc/grpc-service.ts
```diff
@@ -153,42 +145,33 @@ export class GRPCService implements Service {
     return this.resolve(this._client.resolveObject, flagKey, context, logger);
   }
 
-  private connectStream(
-    connectCallback: () => void,
+  private listen(
```
toddbaert, line 148: This method was changed to be more like the in-process equivalent.

libs/providers/flagd/src/lib/service/grpc/grpc-service.ts
```diff
@@ -153,42 +145,33 @@ export class GRPCService implements Service {
...
-      });
-      stream.on('close', () => {
-        this.handleClose();
-      });
-      stream.on('data', (message) => {
-        if (message.type === EVENT_PROVIDER_READY) {
-          this.handleProviderReady(resolve, connectCallback);
-        } else if (message.type === EVENT_CONFIGURATION_CHANGE) {
-          this.handleFlagsChanged(message, changedCallback);
+          reconnectCallback();
```
toddbaert, line 168: When listen is called the first time (with a resolve() callback) we don't want to emit our own READY event, when we are reconnecting, we do. This fixes some duplicate ready events.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -35,85 +30,61 @@ export class GrpcFetch implements DataFetch {
...
-      connectCallback();
       changedCallback([]); // flags changed list not supported
-      resolveConnect();
-      this._nextBackoff = initBackOffMs;
+      // if resolveConnect is undefined, this is a reconnection; we only want to fire the reconnect callback in that case
+      if (resolveConnect) {
+        resolveConnect();
+      } else {
+        reconnectCallback();
+      }
```
toddbaert, line 68: Same as https://github.com/open-feature/js-sdk-contrib/pull/660/files#r1399631787

libs/providers/flagd/src/lib/service/grpc/grpc-service.ts
```diff
@@ -209,38 +192,26 @@ export class GRPCService implements Service {
...
+  private reconnect(
+    reconnectCallback: () => void,
+    changedCallback: (flagsChanged: string[]) => void,
+    disconnectCallback: () => void,
+  ) {
+    const channel = this._client.getChannel();
+    channel.watchConnectivityState(channel.getConnectivityState(true), Infinity, () => {
+      this.listen(reconnectCallback, changedCallback, disconnectCallback);
+    });
+  }
```
toddbaert, line 204: This is the entirety of our reconnect logic now.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -35,85 +30,61 @@ export class GrpcFetch implements DataFetch {
...
-      if (this._nextBackoff > maxBackOffMs) {
-        this._nextBackoff = maxBackOffMs;
-      }
-
-      this.listen(resolver, dataFillCallback, reconnectCallback, changedCallback, disconnectCallback);
-    }, this._nextBackoff);
+  ) {
+    const channel = this._syncClient.getChannel();
+    channel.watchConnectivityState(channel.getConnectivityState(true), Infinity, () => {
+      this.listen(dataFillCallback, reconnectCallback, changedCallback, disconnectCallback);
```
toddbaert, line 87: Same as https://github.com/open-feature/js-sdk-contrib/pull/660/files#r1399633017

libs/providers/flagd/project.json
```diff
@@ -82,7 +82,7 @@
           "npx jest --setupFiles './setup-in-process-provider.ts'"
         ],
         "cwd": "libs/providers/flagd/src/e2e",
-        "parallel": false
+        "parallel": true
```
toddbaert, line 85: Now that I used a named provider, we can run these in parallel.

libs/providers/flagd/src/lib/service/in-process/data-fetch.ts
```diff
@@ -4,7 +4,7 @@
 export interface DataFetch {
   connect(
     dataFillCallback: (flags: string) => void,
-    connectCallback: () => void,
+    reconnectCallback: () => void,
```
toddbaert, line 7: @Kavindu-Dodan I've specifically called this a "reconnectCallback" because it's called only on connections AFTER the initial one (see other comments).

libs/providers/flagd/src/lib/service/grpc/grpc-service.ts
```diff
@@ -73,14 +68,10 @@ export class GRPCService implements Service {
   private _client: ServiceClient;
   private _cache: LRUCache<string, ResolutionDetails<FlagValue>> | undefined;
   private _cacheEnabled = false;
-  private _streamAlive = false;
-  private _streamConnectAttempt = 0;
-  private _stream: ClientReadableStream<EventStreamResponse> | undefined = undefined;
-  private _streamConnectBackoff = BASE_EVENT_STREAM_RETRY_BACKOFF_MS;
-  private _maxEventStreamRetries;
```
toddbaert, line 80: We're able to remove all this sort of state around reconnection.

libs/providers/flagd/src/e2e/step-definitions/evaluation.spec.ts
```diff
@@ -29,10 +29,6 @@ defineFeature(feature, (test) => {
     });
   });
 
-  afterAll(async () => {
-    await OpenFeature.close();
-  });
-
```
toddbaert, line 35: All closes in the e2e suite were removed, and instead moved to the `tear-down.ts`

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -35,85 +30,61 @@ export class GrpcFetch implements DataFetch {
...
-      this._nextBackoff = this._nextBackoff * 2;
-      if (this._nextBackoff > maxBackOffMs) {
-        this._nextBackoff = maxBackOffMs;
-      }
-
-      this.listen(resolver, dataFillCallback, reconnectCallback, changedCallback, disconnectCallback);
-    }, this._nextBackoff);
+  ) {
+    const channel = this._syncClient.getChannel();
+    channel.watchConnectivityState(channel.getConnectivityState(true), Infinity, () => {
```
Kavindu-Dodan, line 86: I don't think we need to get connectivity state with connect retry set to true. This is only needed if the connection is in idle state. In reconnect, we should see a transition from `READY` to `TRANSIENT_FAILURE`

Besides, do we need this as a callback from `channel.watchConnectivityState` 🤔  reason I am asking, we only end up in reconnect only for error. Which means we anyway need to call for a reconnect.

Kavindu-Dodan (approved):
Thanks for e2e tests and the refactoring.

I am arpproving, however I left [this](https://github.com/open-feature/js-sdk-contrib/pull/660#discussion_r1399815480) comment

beeme1mr (approved):

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
toddbaert: I took this pattern based on a comment from one of the gRPC-js maintainers here: https://github.com/grpc/grpc-node/issues/2218#issuecomment-1240994000

It seems this is a good method for reacting to channel readiness.

PR #671, opened 2023-12-06
fix: "in" op, by using json-logic-js

toddbaert:


libs/shared/flagd-core/src/lib/targeting/targeting.ts
```diff
@@ -30,6 +24,11 @@ export class Targeting {
       },
     };
 
-    return this._logicEngine.run(logic, ctxData);
+    // we need access to the context/$flag object in the "fractional" rule, so set it at arg zero if it's not there
```
toddbaert, line 27:
```suggestion
    // we need access to the context/$flagd object in the "fractional" rule, so set it at arg zero if it's not there
```

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -1,20 +1,20 @@
...
 
-// eslint-disable-next-line  @typescript-eslint/no-explicit-any
-export function fractional(data: unknown, context: Record<any, any>): string | null {
-  if (!Array.isArray(data)) {
-    return null;
-  }
-
-  const args = Array.from(data);
+export function fractional(...args: unknown[]): string | null {
   if (args.length < 2) {
```
beeme1mr, line 7: This check should be done after you slice the array.

toddbaert: right!

Kavindu-Dodan (approved):

beeme1mr (approved):

PR #683, opened 2023-12-11
Fix hang

toddbaert:


libs/providers/flagd/project.json
```diff
@@ -78,8 +78,8 @@
       "executor": "nx:run-commands",
       "options": {
         "commands": [
-          "npx jest --setupFiles './setup-rpc-provider.ts'",
-          "npx jest --setupFiles './setup-in-process-provider.ts'"
+          "npx jest --setupFiles './setup-rpc-provider.ts' --runInBand --detectOpenHandles",
+          "npx jest --setupFiles './setup-in-process-provider.ts' --runInBand --detectOpenHandles"
```
toddbaert, line 82: These additional flags cause jest to hang (which is what we want) instead of forcing an exit, so we'll detect future regressions with grpc resource hangs.

libs/providers/flagd/src/lib/service/common/grpc-util.ts
```diff
@@ -0,0 +1,16 @@
...
+   */
+  if (stream) {
+    stream.removeAllListeners();
+    stream.on('error', () => {
+      // swallow afters after closed
+    });
+    stream.cancel();
+    stream.destroy();
+  }
+};
```
toddbaert, line 16: This is the actual fix.

libs/providers/flagd/src/lib/service/common/grpc-util.ts
```diff
@@ -1,10 +1,14 @@
...
+export const closeStreamIfDefined = (stream: ClientReadableStream<unknown> | undefined) => {
+  /**
+   * cancel() is necessary to prevent calls from hanging the process, so we need to we need to remove all the
+   * handlers, and add a no-op for 'error' so we can cancel without bubbling up an exception
+   */
+  if (stream) {
+    stream.removeAllListeners();
+    stream.on('error', () => {});
+    stream.cancel();
+    stream.destroy();
```
beeme1mr, line 14: Are either of these operations async?

toddbaert: No, they both just return void.

beeme1mr (approved):

libs/providers/flagd/src/lib/service/common/grpc-util.ts
```diff
@@ -0,0 +1,16 @@
+import { ClientReadableStream } from '@grpc/grpc-js';
+
+export const closeStreamIfDefined = (stream: ClientReadableStream<unknown> | undefined) => {
```
toddbaert, line 3: This is the actual fix

libs/providers/flagd/src/lib/service/common/grpc-util.ts
```diff
@@ -0,0 +1,16 @@
...
+   * cancel() is necessary to prevent calls from hanging the process, so we need to we need to remove all the
+   * handlers, and add a no-op for 'error' so we can cancel without bubbling up an exception
+   */
+  if (stream) {
+    stream.removeAllListeners();
+    stream.on('error', () => {
+      // swallow errors after closed
+    });
+    stream.cancel();
+    stream.destroy();
```
toddbaert, line 14: This is the actual fix, which I've extracted to a util and made use of in both grpc connections.

PR #686, opened 2023-12-13
fix: fixed issue with nested fractional evaluations

### LINKED ISSUE #685: Fractional evaluations in nested targeting rules are not running

beeme1mr:
## This PR

- reverts back to json logic engine
- adds tests for flag definitions found 

### Related Issues

Fixes #685 

### Notes

The JSON Logic engine library passes context to custom operations. This makes it much easier to add the fractional evaluation operation without having to implement fragile workarounds.

The JSON logic engine has released an updated version that resolved the "in" operation. That was the original reason we switched libraries in the first place.


package.json
```diff
@@ -26,7 +26,7 @@
     "configcat-js": "^8.0.0",
     "copy-anything": "^3.0.5",
     "imurmurhash": "^0.1.4",
-    "json-logic-js": "^2.0.2",
+    "json-logic-engine": "^1.3.1",
```
toddbaert, line 29:
```suggestion
    "json-logic-engine": "1.3.1",
```

This is a bit of an unpopular dependency. We don't HAVE TO but we could lock the version here.

toddbaert (approved):

PR #703, opened 2023-12-18
feat(flagd-core): add update config support, returns changed keys

beeme1mr:
## This PR

- adds an update config method that returns all affected flag keys
- adds hash to internal feature flag class

### DIFF
--- a/libs/shared/flagd-core/src/lib/feature-flag.ts
+++ b/libs/shared/flagd-core/src/lib/feature-flag.ts
@@ -21,7 +21,6 @@ export class FeatureFlag {
   private readonly _targeting: unknown;
   private readonly _hash: string;
 
-
   constructor(flag: Flag) {
     this._state = flag['state'];
     this._defaultVariant = flag['defaultVariant'];
--- a/libs/shared/flagd-core/src/lib/feature-flag.ts
+++ b/libs/shared/flagd-core/src/lib/feature-flag.ts
@@ -1,4 +1,5 @@
 import { FlagValue } from '@openfeature/core';
+import { createHash } from 'crypto';
 
 /**
  * Flagd flag configuration structure mapping to schema definition.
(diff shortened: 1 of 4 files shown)


libs/shared/flagd-core/src/lib/storage.ts
```diff
@@ -12,6 +12,14 @@ export interface Storage {
    */
   setConfigurations(cfg: string): void;
 
+  /**
+   * Updates the configurations and returns the list of flags that have changed.
+   * @param cfg The configuration string to be parsed and stored.
+   * @returns The list of flags that have changed.
+   * @throws {Error} If the configuration string is invalid.
+   */
+  updateConfigurations(cfg: string): string[];
```
toddbaert, line 21: Would it be a problem to simply add this functionality to `setConfigurations`?

libs/shared/flagd-core/src/lib/feature-flag.ts
```diff
@@ -18,12 +19,18 @@ export class FeatureFlag {
...
   private readonly _variants: Map<string, FlagValue>;
   private readonly _targeting: unknown;
+  private readonly _hash: string;
 
   constructor(flag: Flag) {
     this._state = flag['state'];
     this._defaultVariant = flag['defaultVariant'];
     this._variants = new Map<string, FlagValue>(Object.entries(flag['variants']));
     this._targeting = flag['targeting'];
+    this._hash = createHash('sha1').update(JSON.stringify(flag)).digest('base64');
```
toddbaert, line 29: I _think_ this may be a bit less deterministic than you hope.

I think the hash would be different based on the order of the props in the `Flag` you pass in. This might not be that important, _but_, you could also generate this has based on a aggregation of the internal state, which I think would be more deterministic:

Something like: 

```ts
  this._hash = createHash('sha1').update(JSON.stringify(this)).digest('base64');
```

libs/shared/flagd-core/src/lib/storage.ts
beeme1mr: I went back and forth on this. I'm happy to update `setConfigurations`.

toddbaert: From a UX standpoint I think it reduces confusion to do that.

libs/shared/flagd-core/src/lib/feature-flag.ts
beeme1mr: This should be deterministic enough. We would emit a change event if someone modified the order of the properties in the object but that seems like an edge case and something we can easily explain.

toddbaert (approved):

PR #708, opened 2023-12-22
feat: add offline mode, fix in-process connection edge cases

beeme1mr:
## This PR

- adds an offline mode to the in-process evaluator
- fixes reconnection issue affecting the in-process gRPC
- extend tests to cover more scenarios.


libs/providers/flagd/src/lib/service/in-process/data-fetch.ts
```diff
@@ -2,12 +2,34 @@
  * Contract of in-process resolver's data fetcher
  */
 export interface DataFetch {
+  /**
+   * Connects the data fetcher
+   */
   connect(
-    dataFillCallback: (flags: string) => void,
+    /**
+     * Callback that runs when data is filled from the source
```
toddbaert, line 10:
```suggestion
     * Callback that runs when data is received from the source
```

libs/providers/flagd/src/lib/service/in-process/file/file-fetch.ts
```diff
@@ -0,0 +1,60 @@
+import { Logger, OpenFeatureError } from '@openfeature/core';
+import { DataFetch } from '../data-fetch';
+import { promises as fsPromises, watch, type FSWatcher } from 'fs';
+import { GeneralError } from '@openfeature/core';
+
+const encoding = 'utf8';
+
+export class FileFetch implements DataFetch {
+  private _filename: string;
+  private _watcher: FSWatcher | null = null;
```
toddbaert, line 10:
```suggestion
  private _watcher: FSWatcher | undefined = undefined;
```

libs/providers/flagd/src/lib/service/in-process/file/file-fetch.ts
```diff
@@ -0,0 +1,60 @@
...
+      });
+    } catch (err) {
+      if (err instanceof OpenFeatureError) {
+        throw err;
+      } else {
+        switch ((err as { code?: string })?.code) {
+          case 'ENOENT':
+            throw new GeneralError(`File not found: ${this._filename}`);
+          case 'EACCES':
+            throw new GeneralError(`File not accessible: ${this._filename}`);
```
toddbaert, line 48: I think we should add constructors that support error options, to specifically support [cause](https://nodejs.org/api/errors.html#errorcause). This allows you to embed causal errors into rethrown errors. It's supported in browsers as well.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -10,10 +10,12 @@ import { closeStreamIfDefined } from '../../common';
...
  */
 export class GrpcFetch implements DataFetch {
-  private _syncClient: FlagSyncServiceClient;
-  private _syncStream: ClientReadableStream<SyncFlagsResponse> | undefined;
+  private readonly _syncClient: FlagSyncServiceClient;
   private readonly _request: SyncFlagsRequest;
+  private _syncStream: ClientReadableStream<SyncFlagsResponse> | undefined;
   private _logger: Logger | undefined;
+  private _isConnected = false;
+  private _initialized = false;
```
toddbaert, line 18: Maybe add some inline comments to explain the subtle differences between these.

libs/providers/flagd/src/e2e/step-definitions/flagd.spec.ts
```diff
@@ -38,22 +38,24 @@ defineFeature(feature, (test) => {
...
+      client.addHandler(ProviderEvents.ConfigurationChanged, async (details) => {
         ran = true;
+        eventDetails = details;
       });
     });
     and(/^a flag with key "(.*)" is modified$/, async () => {
-      // this happens every 1s in the associated container, so wait 2s
-      await new Promise((resolve) => setTimeout(resolve, 2000));
+      // this happens every 1s in the associated container, so wait 3s
+      await new Promise((resolve) => setTimeout(resolve, 3000));
```
toddbaert, line 52: I ran the test suite another 10 times and it seems stable after this change.

libs/providers/flagd/README.md
```diff
@@ -74,6 +73,19 @@ Flag configurations for evaluation are obtained via gRPC protocol using [sync pr
...
+```
+  OpenFeature.setProvider(new FlagdProvider({
+      resolverType: 'in-process',
+      offlineFlagSourcePath: './flags.json',
+  }))
+```
+
+Offline mode uses `fs.watch` to watch for changes to the file, which has (a few caveat)[https://nodejs.org/docs/latest/api/fs.html#caveats] to be aware of.
+This mode is useful for local development, test cases and for offline application.
+
```
toddbaert, line 88: May want to add something about it only been tested on Linux.

toddbaert (approved):
Nice addition, and thanks for the fixes to the reconnect logic.

Test looks good.

libs/providers/flagd/README.md
```diff
@@ -74,6 +73,19 @@ Flag configurations for evaluation are obtained via gRPC protocol using [sync pr
...
+```
+  OpenFeature.setProvider(new FlagdProvider({
+      resolverType: 'in-process',
+      offlineFlagSourcePath: './flags.json',
+  }))
+```
+
+Offline mode uses `fs.watch` to watch for changes to the file, which has (a few caveat)[https://nodejs.org/docs/latest/api/fs.html#caveats] to be aware of.
+This mode is useful for local development, test cases and for offline application.
```
beeme1mr, line 87:
```suggestion
This mode is useful for local development, test cases, and for offline applications.
```

libs/providers/flagd/src/lib/service/in-process/file/file-fetch.ts
beeme1mr: Yeah, that sounds good. 👍

libs/providers/flagd/README.md
```diff
@@ -74,6 +73,19 @@ Flag configurations for evaluation are obtained via gRPC protocol using [sync pr
...
+To enable this mode, you should provide a valid flag configuration file with the option `offlineFlagSourcePath`.
+
+```
+  OpenFeature.setProvider(new FlagdProvider({
+      resolverType: 'in-process',
+      offlineFlagSourcePath: './flags.json',
+  }))
+```
+
+Offline mode uses `fs.watch` to watch for changes to the file, which has (a few caveat)[https://nodejs.org/docs/latest/api/fs.html#caveats] to be aware of.
```
Kavindu-Dodan, line 86:
```suggestion
Offline mode uses `fs.watch` to watch for changes to the file, which has [a few caveat](https://nodejs.org/docs/latest/api/fs.html#caveats) to be aware of.
```
It might be worth highlighting that this won't work on Docker or K8s where network file systems are used.

libs/providers/flagd/src/lib/service/in-process/file/file-fetch.ts
toddbaert: I've created https://github.com/open-feature/js-sdk-contrib/issues/713

libs/providers/flagd/README.md
beeme1mr: I wonder if I should just switch to `fs.watchFile`. It uses polling to overcome some of these limitations. The overhead should be minimal, and users should only enable this feature if they understand the implication.

What do you think @Kavindu-Dodanand @toddbaert?

Kavindu-Dodan: I prefer polling with `fs.watchFile` as this should reliably emit flag changes (per documentation and other sources). I would not worry about performance as we recommend the in-process offline mode as an alternative to other well-established flag sync mechanisms we support.

libs/providers/flagd/src/lib/service/in-process/file/file-fetch.ts
```diff
@@ -0,0 +1,66 @@
...
+      const output = await fsPromises.readFile(this._filename, encoding);
+      // Don't emit the change event for the initial read
+      dataFillCallback(output);
+
+      // Using watchFile instead of watch to support virtualized host file systems.
+      watchFile(
+        this._filename,
+        {
+          // More aggressive polling interval to support faster flag changes.
+          interval: 1007,
```
Kavindu-Dodan, line 32: I think we should be fine with the default interval 5007. In Java PR (https://github.com/open-feature/java-sdk-contrib/pull/614), I am using 5000 value already.

Kavindu-Dodan (approved):

PR #709, opened 2023-12-22
feat: add flag metadata to in-process evaluator

beeme1mr:
## This PR

- adds flag metadata to the in-process evaluator

### Notes

The defined 'selector' will be included as `scope` in the flag metadata.


Kavindu-Dodan (approved):
This is a good addition for in-process evaluation mode. And we can add the same for other lang. implementations (Go & Java).

toddbaert (commented):
You could extract this duplication with generics and a lambda for the resolver method, I think, but that's optional.

I think you can add some simple tests by injecting a very simple `dataFetcher` impl that would return mocked stuff.

Kavindu-Dodan:
> This is a good addition for in-process evaluation mode. And we can add the same for other lang. implementations (Go & Java).

I created https://github.com/open-feature/java-sdk-contrib/issues/622 for Java & https://github.com/open-feature/go-sdk-contrib/issues/416 fo Go

toddbaert (approved):

PR #714, opened 2024-01-02
chore: throw ParseError on invalid flagd config

beeme1mr:
## This PR

- throws an OpenFeature ParseError when an invalid flagd config parsed


libs/shared/flagd-core/src/lib/parser.ts
```diff
@@ -1,29 +1,40 @@
...
 export function parse(flagCfg: string): Map<string, FeatureFlag> {
-  if (!isValid(JSON.parse(flagCfg))) {
-    throw new Error('invalid flagd flag configurations');
+  try {
+    const transformed = transform(flagCfg);
+    const flags: { flags: { [key: string]: Flag } } = JSON.parse(transformed);
+    const results = matcher(flags);
+    if (!results) {
+      throw new ParseError(errorMessages);
+    }
```
toddbaert, line 24: I find `matcher` and `results` a bit confusing here. The ajv docs use `validate`( https://ajv.js.org/guide/getting-started.html#basic-data-validation) . Maybe `validator` and `valid` would be clearer?

toddbaert (approved):
Approved with this var naming recommendation: https://github.com/open-feature/js-sdk-contrib/pull/714/files#r1439672002

libs/shared/flagd-core/src/lib/parser.ts
beeme1mr: Good catch, fixed.

Kavindu-Dodan (approved):

PR #744, opened 2024-01-29
fix: update disabled behavior to match spec 

beeme1mr:
## This PR

- update disabled behavior to match spec 

### Related Issues

https://github.com/open-feature/flagd/issues/1035

### Notes

Disabled flags will now behave as if they don't exist. This is consistent with the rest of the flagd ecosystem.

### Follow-up Tasks

- [ ] Update the flagd playground
- [ ] Update the JS flagd provider

### DIFF
--- a/libs/shared/flagd-core/src/lib/flagd-core.spec.ts
+++ b/libs/shared/flagd-core/src/lib/flagd-core.spec.ts
@@ -145,14 +145,16 @@ describe('flagd-core validations', () => {
   });
 
   it('should throw because the flag does not exist', () => {
-    expect(() => core.resolveBooleanEvaluation('nonexistentFlagKey', false, {}, console)).toThrow(
-      new FlagNotFoundError("flag: 'nonexistentFlagKey' not found"),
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan (approved):

libs/shared/flagd-core/src/lib/flagd-core.spec.ts
```diff
@@ -144,12 +144,16 @@ describe('flagd-core validations', () => {
     expect(() => core.resolveBooleanEvaluation('myIntFlag', true, {}, console)).toThrow(GeneralError);
   });
 
-  it('should validate flag status', () => {
-    const evaluation = core.resolveBooleanEvaluation('myBoolFlag', false, {}, console);
+  it('should throw because the flag does not exist', () => {
+    expect(() => core.resolveBooleanEvaluation('nonexistentFlagKey', false, {}, console)).toThrow(
```
toddbaert, line 148: suggest making `nonexistentFlagKey` a local var.

PR #746, opened 2024-01-29
fix: falsy boolean shorthand logic

beeme1mr:
## This PR

- fixes falsy boolean shorthand logic

### Notes

Prior to the change, the default variant would be returned when the logic was falsy. After the change, the "false" variant is used.


toddbaert (approved):
Ya this makes sense to me, nice catch

lukas-reining (approved):
Looks good!

PR #767, opened 2024-02-13
fix: init in-process error, throw on invalid rules

toddbaert:
* fixes an unhandled error if sync source not available at startup
* throws on invalid targeting rules as per flagd spec
* adds additional e2e tests

---

Implements the latest e2e tests, including those added [here](https://github.com/open-feature/flagd-testbed/pull/85). Here we decided that invalid targeting should throw, which required a small change in core.

I also found a [bug](https://github.com/open-feature/js-sdk-contrib/pull/767/files#r1488560278) on the in-process provider where there's an unhandled exception in startup.

### DIFF
--- a/libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
+++ b/libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
@@ -74,7 +74,7 @@ export class GrpcFetch implements DataFetch {
       this._syncStream = this._syncClient.syncFlags(this._request);
       this._syncStream.on('data', (data: SyncFlagsResponse) => {
         this._logger?.debug(`Received sync payload`);
-  
+
         try {
           const changes = dataCallback(data.flagConfiguration);
           if (this._initialized && changes.length > 0) {
@@ -83,7 +83,7 @@ export class GrpcFetch implements DataFetch {
         } catch (err) {
           this._logger?.debug('Error processing sync payload: ', (err as Error)?.message ?? 'unknown error');
         }
-  
+
         if (resolveConnect) {
           resolveConnect();
         } else if (!this._isConnected) {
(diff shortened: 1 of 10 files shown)


libs/shared/flagd-core/src/lib/flagd-core.spec.ts
```diff
@@ -225,4 +225,12 @@ describe('flagd-core common flag definitions', () => {
     expect(resolved.reason).toBe(StandardResolutionReasons.STATIC);
     expect(resolved.variant).toBe('false');
   });
+
+  it('should throw with invalid targeting rules', () => {
+    const core = new FlagdCore();
+    const flagCfg = `{"flags":{"isEnabled":{"state":"ENABLED","variants":{"true":true,"false":false},"defaultVariant":"false","targeting":{"invalid": ["this is not valid targeting"]}}}}`;
+    core.setConfigurations(flagCfg);
+
+    expect(() => core.resolveBooleanEvaluation('isEnabled', false, {}, console)).toThrow();
```
toddbaert, line 234: New unit test for e2e test I mentioned.

libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -165,8 +165,7 @@ export class FlagdCore implements Storage {
       try {
         targetingResolution = this._targeting.applyTargeting(flagKey, flag.targeting, evalCtx);
       } catch (e) {
-        logger.error(`Error evaluating targeting rule for flag ${flagKey}, falling back to default`, e);
-        targetingResolution = null;
+        throw new GeneralError(`Error evaluating targeting rule for flag ${flagKey}: ${(e as Error)?.message}`);
```
toddbaert, line 168: throw if targeting thows.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -70,39 +70,67 @@ export class GrpcFetch implements DataFetch {
   ) {
     this._logger?.debug('Starting gRPC sync connection');
     closeStreamIfDefined(this._syncStream);
-    this._syncStream = this._syncClient.syncFlags(this._request);
```
toddbaert, line 73: This can throw before it ever connects... when that happens the `.on('error')` handler never takes effect. This is fixed by making the catch and `.on('error')` both share a local method (`handleError`).

Kavindu-Dodan (approved):

PR #770, opened 2024-02-15
feat: use updated proto

toddbaert:
Updates to new proto. No changes needed, but getting ahead of any metadata implementation in flagd


toddbaert:
@Kavindu-Dodan this uses the new "breaking" proto - as you can see, the e2e tests still work! Using a new tag and `reserved` seems to be a great solution.

Kavindu-Dodan (approved):

PR #776, opened 2024-02-16
feat: OFREP web provider

thomaspoignant:
This is a long-running PR for the OFREP-web provider, the goal is to use this to use this PR to test the implementation of the OpenFeature Remote Evaluation Protocol to understand the limit of the protocol.

Check [`open-feature/protocol`](https://github.com/open-feature/protocol) to know more about it.


libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,246 @@
...
+    if (response.httpStatus === 200) {
+      const bulkSuccessResp = response.value;
+      const newCache: { [key: string]: ResolutionDetails<FlagValue> } = {};
+
+      bulkSuccessResp.flags?.forEach((evalResp: EvaluationResponse) => {
+        if (evalResp.key === undefined) {
+          return;
+        }
+        const key = evalResp.key;
+        const evalSuccessResp = evalResp as EvaluationSuccessResponse;
```
lukas-reining, line 159: This cast should be replaced by handling error errors on here correctly.
Maybe we can have a cache of success and error responses or so?
When evaluating a flag with an error response we might want to return the default value and the error from the response or so.

thomaspoignant: Good catch, I've reworked a bit that part and stored in the cache the evaluation in error during the bulk evaluation, to be able to throw the correct error.

libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
```diff
@@ -0,0 +1,6 @@
+import { FetchAPI } from '@openfeature/ofrep-core';
+export interface OfrepWebProviderOptions {
+  baseUrl: string;
+  pollInterval?: number; // in milliseconds
+  disablePolling?: boolean;
```
beeme1mr, line 5: Could `pollInterval` take a negative value to disable polling?

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
```diff
@@ -0,0 +1,176 @@
+import { OfrepWebProvider } from './ofrep-web-provider';
+import TestLogger from './test-logger';
+// eslint-disable-next-line @nx/enforce-module-boundaries
```
beeme1mr, line 3: Can we update the nx boundary so we don't need to disable this rule?

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,275 @@
...
+    return this.evaluate(flagKey, 'number');
+  }
+  resolveObjectEvaluation<T extends JsonValue>(flagKey: string): ResolutionDetails<T> {
+    return this.evaluate(flagKey, 'object');
+  }
+
+  /**
+   * onContextChange is called when the context changes, it will re-evaluate the flags with the new context
+   * and update the cache.
+   * @param oldContext - the old context (we are not using it)
```
beeme1mr, line 86: What do you mean by the "we are not using it" comment?

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,275 @@
...
+   * @param newContext - the new context
+   */
+  async onContextChange?(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
```
beeme1mr, line 89:
```suggestion
  async onContextChange(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
```

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,275 @@
...
+    try {
+      if (oldContext === newContext) {
```
beeme1mr, line 91: This would cause the `onContextChange` method to exit early if someone did this:

```ts
const context = { name: "mike" };

await OpenFeature.setContext(context);
context.name = "thomas";
// This would be skipped
await OpenFeature.setContext(Context);
```

This behavior may be surprising to users.

libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
thomaspoignant: Yes it is a good idea, I'll change that.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
thomaspoignant: It is not accurate, I've changed the logic and not the comment :(

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
thomaspoignant: good catch, I will remove this early return.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
thomaspoignant: Unfortunately, I don't know much about nx, not sure how to disable this rule.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,309 @@
...
+  /**
+   * onContextChange is called when the context changes, it will re-evaluate the flags with the new context
+   * and update the cache.
+   * @param oldContext - the old evaluation context
+   * @param newContext - the new evaluation context
+   */
+  async onContextChange(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
+    try {
+      const now = new Date();
+      if (this._retryPollingAfter !== undefined && this._retryPollingAfter > now) {
```
beeme1mr, line 95: This is an interesting use case that we should discuss. I'm concerned about ignoring context change, but in this case, the server has already to try again at a later time. This could put us in a situation where the UI doesn't rerender components due to a context change.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,309 @@
...
+  private startPolling() {
+    this._pollingIntervalId = setInterval(async () => {
+      try {
+        const now = new Date();
+        if (this._retryPollingAfter !== undefined && this._retryPollingAfter > now) {
+          return;
+        }
+        const res = await this._evaluateFlags(this._context);
+        if (res === EvaluationStatus.SUCCESS_WITH_CHANGES) {
+          this.events?.emit(ClientProviderEvents.ConfigurationChanged, { message: 'Flags updated' });
```
beeme1mr, line 292: It would be nice to include the list of flags that were changed. In flagd, the changes were calculated like [this](https://github.com/open-feature/js-sdk-contrib/blob/9d6fe63e8954fa56a5ed3e3fdf3117b835ab1bc1/libs/shared/flagd-core/src/lib/storage.ts).

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
thomaspoignant: For now, I keep things stale but we may consider sending an error in that case so we will send the default values until the next evaluation.

Thinking about it, putting the provider in `ERROR` is probably better in that case.
WDYT?

thomaspoignant: I've double-checked and we are already in `error` status when calling `onContextChange`, so from my point of view we are good, we will send the default value until we have reached the `Retry-after` date.

I've just spotted a bug, we were not storing the new context. Now it is stored and next time we will call the API (via polling we will have the new context set).

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,347 @@
...
+        return { status: BulkEvaluationStatus.SUCCESS_NO_CHANGES, flags: [] };
+      }
+
+      if (response.httpStatus === 400) {
+        handleEvaluationError(response);
+      }
+
+      if (response.httpStatus === 200) {
+        const bulkSuccessResp = response.value;
+        const newCache: { [key: string]: ResolutionDetails<FlagValue> | ResolutionError } = {};
```
Kavindu-Dodan, line 164: minor - is there any way to push http validations to the ofrep-core ? 🤔

libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
```diff
@@ -0,0 +1,15 @@
+export interface OfrepWebProviderOptions {
```
Kavindu-Dodan, line 1: minor - we should expose a mechanism to set authN headers. See comment [1] for server provider 

[1] - https://github.com/open-feature/js-sdk-contrib/pull/817#discussion_r1546648202

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,347 @@
...
+      if (response.httpStatus === 304) {
+        // nothing has changed since last time, we are doing nothing
+        return { status: BulkEvaluationStatus.SUCCESS_NO_CHANGES, flags: [] };
+      }
+
+      if (response.httpStatus === 400) {
+        handleEvaluationError(response);
+      }
+
+      if (response.httpStatus === 200) {
```
lukas-reining, line 162: If you change 
 ```ts
      if (response.httpStatus === 400) {
        handleEvaluationError(response);
      }
```
to 
```
 ```ts
      if (response.httpStatus !== 200) {
        handleEvaluationError(response);
      }
```

you can remove the check for status 200 and do not have to throw the error in the end.
```

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
lukas-reining: We already have that @Kavindu-Dodan.
You can see it in [line 159](https://github.com/open-feature/js-sdk-contrib/pull/776/files#diff-88744cdb4c1b6a1f79a835f0727ee832c052438bbb567cdb3a73034fe61ff817R159).
At this place, we should even know that there can only be a 200 left, which makes the type of value `BulkEvaluationSuccessResponse`, As stated here: https://github.com/open-feature/js-sdk-contrib/pull/776/files#r1546796856

The rest, mapping failure and success responses is really specific for this provider.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+import { server } from '../../../../shared/ofrep-core/src/test/mock-service-worker';
+import { ClientProviderEvents, ClientProviderStatus, OpenFeature } from '@openfeature/web-sdk';
+
+describe('OFREPWebProvider', () => {
+  beforeAll(() => server.listen());
+  afterEach(async () => {
+    server.resetHandlers();
+    await OpenFeature.close();
+  });
+  afterAll(() => server.close());
```
lukas-reining, line 13: This should use MSW client.
It is called worker I think: https://mswjs.io/docs/integrations/browser#setup

thomaspoignant: No jest is acting only as a server and I've read somewhere that we should use the server with jest.

Will try to find the link tomorrow

lukas-reining: Mh I would have assumed to use JestDomEnv here

lukas-reining: Just had a look I understand what you are saying @beeme1mr. So you are right :)

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
toddbaert: I _think_ we can just update this import to use `@openfeature/ofrep-core`. This way we let nx do all the work. All internal imports from other packages in this repo should work this way.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
```diff
@@ -0,0 +1,300 @@
+import { OfrepWebProvider } from './ofrep-web-provider';
+import TestLogger from '../../test/test-logger';
+// eslint-disable-next-line @nx/enforce-module-boundaries
+import { server } from '../../../../shared/ofrep-core/src/test/mock-service-worker';
```
toddbaert, line 4:
```suggestion
import { server } from '@openfeature/ofrep-core';
```

lukas-reining: Same answer as here: https://github.com/open-feature/js-sdk-contrib/pull/776#discussion_r1549869475

libs/providers/ofrep-web/package.json
```diff
@@ -0,0 +1,16 @@
...
+    "tslib": "^2.3.0"
+  },
+  "main": "./src/index.js",
+  "typings": "./src/index.d.ts",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/web-sdk": ">=0.4.0"
```
toddbaert, line 14: I think you need a peer dep on ofrep core?

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,344 @@
...
+      }
+
+      this._logger?.debug(`${this.metadata.name} initialized successfully`);
+    } catch (error) {
+      if (error instanceof OFREPApiUnauthorizedError || error instanceof OFREPForbiddenError) {
+        throw new ProviderFatalError('Initialization failed', error);
+      }
+      throw error;
+    }
+  }
```
toddbaert, line 265: nit: move this closer to the top... for a while I thought it was missing :sweat_smile:

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,344 @@
...
+   * Initialize the provider, it will evaluate the flags and start the polling if it is not disabled.
+   * @param context - the context to use for the evaluation
+   */
+  async initialize?(context?: EvaluationContext | undefined): Promise<void> {
+    try {
+      this._context = context;
+      await this._evaluateFlags(context);
+
+      if (this._pollingInterval > 0) {
+        this.startPolling();
```
toddbaert, line 255: Not 100% sure if we want, but worth considering if we should pass some sort of callback here that we use to resolve a promise on the first poll success, so that we don't resolve init until we know the polling is working.

I'm not sure how long this would take or if that's feasible though. Just a thought.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,344 @@
...
+          this._retryPollingAfter = retryAfterDate;
+        }
+      }
+    }
+  }
+  /**
+   * Initialize the provider, it will evaluate the flags and start the polling if it is not disabled.
+   * @param context - the context to use for the evaluation
+   */
+  async initialize?(context?: EvaluationContext | undefined): Promise<void> {
```
toddbaert, line 249: nit: move this method closer to the top... for a while I thought it was missing 😅

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
toddbaert: Oh I didn't realize this was just testing... I just saw the cross package import....

If this is testing, forget it! I think it's OK.

beeme1mr: We would probably have to create a test-specific library in NX to do this properly and it's likely not worth the effort.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,344 @@
...
+      if (error instanceof OFREPApiTooManyRequestsError) {
+        this.events?.emit(ClientProviderEvents.Stale, { message: `${error.name}: ${error.message}` });
+        return;
+      }
+
+      if (
+        error instanceof OpenFeatureError ||
+        error instanceof OFREPApiFetchError ||
+        error instanceof OFREPApiUnauthorizedError ||
+        error instanceof OFREPForbiddenError
```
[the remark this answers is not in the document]
beeme1mr: I don't think this will be an issue here because we should be able to bundle everything in the provider when we publish. We shouldn't even need to publish ofrep core.

libs/providers/ofrep-web/package.json
toddbaert: We can bundle it, but in that case we have to be aware if 2 modules bundle it, they might both export artifacts from them and then we will get the problems I described here.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,352 @@
...
+      if (!isNaN(retryAfter as unknown as number)) {
+        this._retryPollingAfter = new Date(Date.now() + (retryAfter as unknown as number) * 1000);
+      } else {
+        const retryAfterDate = new Date(retryAfter);
+        if (retryAfterDate.toString() !== 'Invalid Date') {
+          this._retryPollingAfter = retryAfterDate;
+        }
+      }
+    }
+  }
```
lukas-reining, line 273: I guess we should not implicitly convert the string header into numbers here.
I just opened a PR to solve paring the header in ofrep-shared.
With this you should be able to just do `error.retryAfterDate ` and assign it to `this._retryPollingAfter` if it is not null.

```js
private _setRetryAfter(error: OFREPApiTooManyRequestsError) {
  const retryAfterDate = error.retryAfterDate
  if(retryAfterDate){
    this._retryPollingAfter = retryAfterDate
  }
}
```

thomaspoignant: That's a good idea to put the logic in `ofrep-core`.
I will update the code when we will release the new version.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
thomaspoignant: The polling calls the same function `this._evaluateFlags(context);` as the one called before.
This is why I did not consider this, we will fail in the first call before the polling if we have an issue, so I expect the polling to work here.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
lukas-reining: It is deployed :)

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -0,0 +1,333 @@
...
+  // logger is the Open Feature logger to use
+  private _logger?: Logger;
+  // _options is the options used to configure the provider.
+  private _options: OfrepWebProviderOptions;
+
+  private _ofrepAPI: OFREPApi;
+  private _etag: string | null;
+  private _pollingInterval: number;
+  private _retryPollingAfter: Date | undefined;
+  private _cache: { [key: string]: ResolutionDetails<FlagValue> | ResolutionError } = {};
```
lukas-reining, line 61: Nit: Maybe we can have a cache type as it is used four time. But we do not have to.

lukas-reining (approved):
Just found one nit, otherwise this looks great!
Really looking forward! :)

beeme1mr (approved):

beeme1mr:
Should we try and release today?

thomaspoignant:
> Should we try and release today?

That would be awesome.
I'll release it now.

PR #795, opened 2024-03-13
feat(ofrep): ofrep core

lukas-reining:
## This PR

Adds the core implementation for OFREP providers.
It defines the types and a class `OFREPApi` which handles calling the OFREP API with fetch.

Tests are using [Mock Service Worker](https://mswjs.io/) to avoid mocking.


lukas-reining:
@beeme1mr @toddbaert I realized that the pipeline step for node 16 fails because my implementation and [MSW](https://mswjs.io/docs/faq/#requestresponsetextencoder-is-not-defined-jest) assume Node `v18.x+` (for using fetch), but the failing step is `v16.x`.
I remember that we wanted to remove Node 16 from the js-sdk repo already as it is [EOL anyways ](https://nodejs.org/en/about/previous-releases).
What do you think, can we use this chance to remove Node 16?

beeme1mr:
I'm fine updating the versions we test. Any version no longer supported by node doesn't need to be officially supported by us.

lukas-reining:
> I'm fine updating the versions we test. Any version no longer supported by node doesn't need to be officially supported by us.

I would open a new PR tomorrow morning for both JS repos then.
Or how would you do it?

.release-please-manifest.json
```diff
@@ -8,5 +8,6 @@
   "libs/providers/launchdarkly-client": "0.2.1",
   "libs/providers/go-feature-flag-web": "0.1.6",
   "libs/shared/flagd-core": "0.1.11",
+  "libs/shared/ofrep-core": "0.0.1",
```
thomaspoignant, line 11: Can we start with an experimental version here maybe?

```suggestion
  "libs/shared/ofrep-core": "0.0.1-experimental",
```

libs/shared/ofrep-core/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@openfeature/ofrep-core",
+  "version": "0.0.1",
```
thomaspoignant, line 3: Can we start with an experimental version here maybe?
```suggestion
  "version": "0.0.1-experimental",
```

thomaspoignant (commented):
I've made a first pass and it looks great.
I will try to do a more detailed one tomorrow, but I like it for now.

The only NIT thing for me is the versioning, I would start with an experimental version to be able to have a proper 0.0.1 when it is fully tested from a provider perspective.

lukas-reining:
Sure makes total sense @thomaspoignant :)

libs/shared/ofrep-core/src/lib/model/evaluation.ts
```diff
@@ -0,0 +1,87 @@
+import { EvaluationContext, FlagValue } from '@openfeature/core';
```
beeme1mr, line 1:
```suggestion
import type { EvaluationContext, FlagValue } from '@openfeature/core';
```

libs/shared/ofrep-core/src/lib/model/ofrep-api-result.ts
```diff
@@ -0,0 +1,37 @@
...
+> {
+  readonly httpStatus: S;
+  readonly value: T;
+  readonly httpResponse: R;
+}
+
+export const OFREPEvaluationErrorHttpStatuses = [400, 404, 500] as const;
+export type OFREPEvaluationErrorHttpStatus = (typeof OFREPEvaluationErrorHttpStatuses)[number];
+
+export const OFREPEvaluationSuccessHttpStatuses = [200] as const;
```
thomaspoignant, line 17: Should we consider the `304` as a successful response too?

From what I see in the code we will throw a `OFREPApiUnexpectedResponseError` if we receive a 304 in the bulk evaluation.

tsconfig.base.json
```diff
@@ -22,6 +22,7 @@
       "@openfeature/env-var-provider": ["libs/providers/env-var/src/index.ts"],
       "@openfeature/flagd-core": ["libs/shared/flagd-core/src/index.ts"],
       "@openfeature/flagd-provider": ["libs/providers/flagd/src/index.ts"],
+      "@openfeature/ofrep-shared": ["libs/shared/ofrep-core/src/index.ts"],
```
beeme1mr, line 25:
```suggestion
      "@openfeature/ofrep-core": ["libs/shared/ofrep-core/src/index.ts"],
```

This matches the file path. There are likely other places that reference `@openfeature/ofrep-shared`. Please make sure to change those too if you accept this proposal.

libs/shared/ofrep-core/src/lib/model/ofrep-api-result.ts
lukas-reining: You are right this one was missing.
I will add it later this evening. In this case I would just add an undefined value.

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (response.status === 200 && isEvaluationSuccessResponse(body)) {
+      return { httpStatus: response.status, value: body, httpResponse: response };
+    } else if (OFREPApi.isOFREFErrorHttpStatus(response.status) && isEvaluationFailureResponse(body)) {
+      return { httpStatus: response.status, value: body, httpResponse: response };
+    }
+
+    throw new OFREPApiUnexpectedResponseError(response, 'The OFREP response does not match the expected format');
+  }
+
+  public async postBulkEvaluateFlags(
```
thomaspoignant, line 98: We are missing the possibility of adding the `If-None-Match` header in the request to trigger the 304.

tsconfig.base.json
lukas-reining: Just checked it, it was only here.
No code referenced this until now so this was the only place.

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -0,0 +1,117 @@
...
+    evaluationRequest?: EvaluationRequest,
+    options?: RequestOptions,
+  ): Promise<OFREPApiBulkEvaluationResult> {
```
thomaspoignant, line 101: Just to double check with you here, but we will have access to the headers here?
We need access to the `ETag` header to collect the hash that we will check in the client providers.

libs/shared/ofrep-core/src/lib/model/ofrep-api-result.ts
thomaspoignant: Yes perfect 👍

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
lukas-reining: Actually we are not missing it.
You can add any header, auth, caching, etc. in the request options, these are the standard options of fetch.
We could add a convenience parameter here but generally it is possible.

thomaspoignant: I am fine using the default option I did not check enough sorry :(

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
lukas-reining: Yes exactly.
The idea was to have the 3 values, http status, OFREP response and fetch response in the result.
The status is mainly for making nice typing and the response is the full fetch response.

beeme1mr:
Is any of this code auto-generated from the OpenAPI definition?

lukas-reining:
> Is any of this code auto-generated from the OpenAPI definition?

No it is not.
@thomaspoignant and I were talking about it and we both thought that the generation would give us much overhead in the types and much stuff that we do not need. I actually tried it and could confirm that it was much overhead and writing it ourselfs gives us a better api with not more work. 
The types that came out were also not perfect in the sense of type inference and so.
We would change it a bit anyways and it feels like the generation makes it less expressive while not saving us too much.

libs/shared/ofrep-core/src/lib/model/ofrep-api-result.ts
lukas-reining: @thomaspoignant I added 304 handling as described for bulk evals now.
If the value is 304 the infered type in ts is undefined and this can by handled by the consumer.

thomaspoignant (approved):
Thanks for the 304.
We have a good start here 🙌

beeme1mr (approved):

PR #796, opened 2024-03-14
fix!: Allow ConfigCat provider to be used in server applications

adams85:
## This PR

The `conficat-js` package that the ConfigCat provider is based on is for browser applications only as it uses `XMLHttpRequest` under the hood. However, the provider might be used in server applications (like it was attempted in a Node.js application [here](https://www.linkedin.com/video/event/urn:li:ugcPost:7173464062830239744), at around 2:00:00).

For multiplatform scenarios, ConfigCat provides [the `conficat-js-ssr` package](https://configcat.com/docs/sdk-reference/js-ssr/) (that uses Axios under the hood). This provider should use that instead of `conficat-js` to provide a more extensive support for various JS platforms.

This PR also fixes another serious shortcoming of the provider: currently it cannot evaluate feature flags without setting `targetingKey` in the evaluation context. Simple feature flags may not use targeting at all, so requiring a `targetingKey` to evaluate them is obviously not correct.

### Related Issues

n/a

### Notes

n/a

### Follow-up Tasks

n/a

### How to test

n/a

### DIFF
--- a/libs/providers/config-cat/README.md
+++ b/libs/providers/config-cat/README.md
@@ -14,20 +14,21 @@ The OpenFeature SDK is required as peer dependency.
 
 The minimum required version of `@openfeature/server-sdk` currently is `1.6.0`.
 
-The minimum required version of `configcat-js` currently is `8.0.0`.
+The minimum required version of `configcat-js-ssr` currently is `7.1.2`.
 
 ```
-$ npm install @openfeature/server-sdk configcat-js
+$ npm install @openfeature/server-sdk configcat-js-ssr
 ```
 
 ## Usage
 
-The ConfigCat provider uses the [ConfigCat Javascript SDK](https://configcat.com/docs/sdk-reference/js/).
+The ConfigCat provider uses the [ConfigCat JavaScript (SSR) SDK](https://configcat.com/docs/sdk-reference/js-ssr/).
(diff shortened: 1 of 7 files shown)


laliconfigcat (approved):

lukas-reining:
Hey @adams85 thank you this change looks good!
Maybe we should also consider developing a client-sdk provider, as this is a server sdk provider. Which makes you change even more important.
We can release this, but it has to be a breaking change as the API changed.

lukas-reining (approved):
Looks good to me and thanks for the changes!

adams85:
Thanks for the release, @lukas-reining. I did a quick test and found that the new version (v0.6.0) works fine.

> Maybe we should also consider developing a client-sdk provider, as this is a server sdk provider. Which makes you change even more important.

Since our SSR SDK can handle both targets, I guess the client provider would be more or less just a duplicate of the server one (at least, in our case). I'd avoid maintaining another package if possible.

BTW, NPM packages kind of support multitargeting via [the `browser` field](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#browser). AFAIK, most of the bundlers support it nowadays. I have plans to look into the viability of this approach because it would be a godsend to us if we could merge our various JS SDK packages (or at least some) into one.

PR #798, opened 2024-03-14
feat!: update OpenFeature SDK peer

toddbaert:
Updates the SDK version used in the repo, AND the peer versions required by all components.

I've marked this as breaking since it requires a major version update, which I think is wise to prompt SDK upgrades from consumers since these are all sub 1.0 packages.

When this is merged, I intend to release all the artifacts in this repo.


package.json
```diff
@@ -45,10 +45,10 @@
     "@nx/rollup": "16.9.1",
     "@nx/web": "16.9.1",
     "@nx/workspace": "16.9.1",
-    "@openfeature/core": "^0.0.21",
+    "@openfeature/core": "^1.0.0",
     "@openfeature/flagd-core": ">=0.1.10",
-    "@openfeature/server-sdk": "^1.8.0",
-    "@openfeature/web-sdk": "^0.4.0",
+    "@openfeature/server-sdk": "^1.13.2",
+    "@openfeature/web-sdk": "file:.yalc/@openfeature/web-sdk",
```
toddbaert, line 51: This needs to be updated to `1.0.0` after the web SDK is released, obviously the build will fail until then, but I've tested this locally with `yalc`.

libs/providers/go-feature-flag-web/src/lib/go-feature-flag-web-provider.spec.ts
```diff
@@ -214,7 +214,7 @@ describe('GoFeatureFlagWebProvider', () => {
     it('should get a valid boolean flag evaluation', async () => {
       const flagKey = 'bool_flag';
       await OpenFeature.setContext(defaultContext);
-      OpenFeature.setProvider('test-provider', defaultProvider);
+      await OpenFeature.setProviderAndWait('test-provider', defaultProvider);
```
toddbaert, line 217: @thomaspoignant @beeme1mr These tests change due to our new "short circuiting logic" (there weren't ready since we weren't awaiting the init).

lukas-reining (approved):

thomaspoignant (approved):
🙌

PR #817, opened 2024-03-23
feat: implement OFREP server provider

lukas-reining:
## This PR

Implements the server provider for OFREP.
It uses the shared OFREP API call implementation.


libs/providers/ofrep/src/lib/ofrep-provider.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private toFlagMetadata(metadata: object): FlagMetadata {
+    // OFREP metadata is defined as any object but OF metadata is defined as Record<string, string | number | boolean>
+    const originalEntries = Object.entries(metadata);
+    const onlyPrimitiveEntries = originalEntries.filter(([, value]) =>
+      ['string', 'number', 'boolean'].includes(typeof value),
+    );
+    return Object.fromEntries(onlyPrimitiveEntries);
+  }
```
lukas-reining, line 112: @thomaspoignant we will have to see what should be done here.
Right now I am ignoring all non primitive values to make the `metadata` compatible with the real type and spec.
The problem here is that the OFREP type currently is object while it is stricter here in JS and the [spec](https://openfeature.dev/specification/types/#flag-metadata).

I see 3 possible solutions:
- Change the OFREP type for FlagMetadata to be the same as in the spec, and so also the same as the JS type
- Filter incompatible values (as I do here)
- Throw an error if the value coming from OFREP is not compatible with our solution

My preferred solution would be the first, as I think OFREP should not have broader `flagMetadata` than our specification as it creates this issue for no real value.

Go currently just does not provide a getter for a non primitive which is closest to the second option.
The FlagMetadata type there is `type FlagMetadata map[string]interface{}`, but there are only getters for primitives. 
https://github.com/open-feature/go-sdk/blob/2944608562d04521fcf92000948a22012e630471/openfeature/client.go#L185

libs/providers/ofrep/src/lib/ofrep-provider.ts
```diff
@@ -0,0 +1,113 @@
...
+      throw new TypeMismatchError();
+    }
+
+    return {
+      value: result.value.value as T,
+      variant: result.value.variant,
+      reason: result.value.reason,
+      flagMetadata: result.value.metadata && this.toFlagMetadata(result.value.metadata),
+    };
+  }
```
lukas-reining, line 103: @thomaspoignant maybe this could be moved into a function in `ofrep-core`, I could image this might be the same for the web provider. What do you think?

thomaspoignant: I am actually building something pretty similar in the web-provider, so yes I am in favour of moving it.

libs/providers/ofrep/src/lib/ofrep-provider.ts
thomaspoignant: I would probably go for the 1st option.
Since OFREP is supposed to follow the OpenFeature spec, I see no reason to keep it less restrictive in OFREP.

Kavindu-Dodan: @lukas-reining sorry, I missed the notification on this discussion. 

I also agree with @thomaspoignant on this and thank you for fixing this already in the OFREP side :) I will check this for my go sdk implementation.

lukas-reining: Changed here https://github.com/open-feature/protocol/pull/8

libs/providers/ofrep/src/lib/ofrep-provider.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export class OfrepProvider implements Provider {
+  private ofrepApi: OFREPApi;
+
+  readonly runsOn = 'server';
+  readonly metadata = {
+    name: OfrepProvider.name,
+  };
+
+  constructor(baseUrl: string) {
```
Kavindu-Dodan, line 18: Minor - knowing this is a WIP, I think we should introduce configuration options, especially targeting authentication headers. 

For go-sdk provider, I added header customization options [1]. This can be used for authentication header. 

[1] - https://github.com/open-feature/go-sdk-contrib/blob/69283f92c207f161c38886d9f9922a36aea76d05/providers/ofrep/README.md#configuration

lukas-reining: Yes totally, just did not do it yet :) 
I just pushed a change for this, including static headers and headerFactories.
I guess this should be a good start, feel free to have a look @Kavindu-Dodan.
In the core we also added a custom fetch implementation at least, this can now also be set.

thomaspoignant (approved):
LGTM 👌

libs/providers/ofrep/src/lib/ofrep-provider.ts
```diff
@@ -0,0 +1,77 @@
...
+  private toResolutionDetails<T extends EvaluationFlagValue>(
+    result: OFREPApiEvaluationResult,
+    defaultValue: T,
+  ): ResolutionDetails<T> {
+    if (result.httpStatus !== 200) {
+      handleEvaluationError(result);
+    }
+
+    if (typeof result.value.value !== typeof defaultValue) {
+      throw new TypeMismatchError();
```
toddbaert, line 68: Might want to add a message string here, just will make things easier to catch in logs, etc.

toddbaert (approved):

libs/providers/ofrep/src/lib/ofrep-provider.ts
lukas-reining: Good point :)

lukas-reining:
I think this is ready for a first release now.
What do you think @thomaspoignant, @toddbaert, @beeme1mr, @Kavindu-Dodan?

libs/providers/ofrep/README.md
```diff
@@ -0,0 +1,15 @@
+# ofrep Provider
```
Kavindu-Dodan, line 1: nit - shall we add a usage example with options?

Kavindu-Dodan (approved):
@lukas-reining overall looks good to me. I added a comment on documetnation.

libs/providers/ofrep/README.md
Kavindu-Dodan: Looks good :) left a comment on ordering of the sections :D

libs/providers/ofrep/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+OpenFeature.setProvider(
+  new OFREPProvider({
+    baseUrl: 'https://localhost:8080',
+    fetchImplementation: fetchPolyfill
+  }),
+);
+```
+
+#### Header Factory
```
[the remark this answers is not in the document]
lukas-reining: Fixed :D

libs/providers/ofrep/src/lib/ofrep-provider.ts
```diff
@@ -0,0 +1,80 @@
...
+} from '@openfeature/ofrep-core';
+
+export type OFREPProviderOptions = OFREPProviderBaseOptions;
+
+export class OFREPProvider implements Provider {
+  private ofrepApi: OFREPApi;
+
+  readonly runsOn = 'server';
+  readonly metadata = {
+    name: OFREPProvider.name,
```
beeme1mr, line 19: I think we should define what the provider name should be before merging this. Perhaps we could use the full name like this "OpenFeature Remote Evaluation Protocol Provider". We talked about avoiding using provider in the name in the past but I think it's valuable to include in this case.

FYI @toddbaert @thomaspoignant @lukas-reining

toddbaert: I almost made the same comment, so it's probably a good Idea. We should decide on a canonical name, probably without "provider".

libs/providers/ofrep/src/lib/ofrep-provider.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  constructor(private options: OFREPProviderOptions) {
+    this.ofrepApi = new OFREPApi(options.baseUrl, options.fetchImplementation);
```
beeme1mr, line 23: Should we validate the base URL in the OFREPApi constructor and provide a friendly error if it's invalid?

lukas-reining: We ware doing it in all evaluations.

libs/providers/ofrep/src/lib/ofrep-provider.ts
lukas-reining: You mean we set the following? 

```ts
  readonly metadata = {
    name: "OpenFeature Remote Evaluation Protocol Provider",
  }
```

thomaspoignant: I think we will have to revisit that when we will have the metadata endpoint.

beeme1mr: > You mean we set the following?

Yes, exactly. That's what we show in telemetry, so it should be user-friendly and consistent across implementations.

lukas-reining: Okay I set it to that, is this fine for all?
@beeme1mr @toddbaert @thomaspoignant

libs/providers/ofrep/package.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "main": "./src/index.js",
+  "typings": "./src/index.d.ts",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/server-sdk": "^1.6.0",
+    "@openfeature/ofrep-core": "^0.0.3-experimental"
```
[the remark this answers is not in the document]
lukas-reining: @toddbaert @beeme1mr @thomaspoignant what do we do about this?
For flagd we are doing peerDeps. I think both is fine and I think I prefer peers.

libs/providers/ofrep/src/lib/ofrep-provider.ts
thomaspoignant: I would add the server in it.
Something like :
```
  readonly metadata = {
    name: "OpenFeature Remote Evaluation Protocol Server Provider",
  }
```

libs/providers/ofrep/package.json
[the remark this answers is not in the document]
beeme1mr: As long as we are publishing core, we can leave it as a peer for now. It would be easy enough for us to change it later if we want.

libs/providers/ofrep/src/lib/ofrep-provider.ts
```diff
@@ -0,0 +1,133 @@
...
+        reason: StandardResolutionReasons.DEFAULT,
+        errorCode: ErrorCode.GENERAL,
+        flagMetadata: {
+          retryAfter: this.notBefore.toISOString(),
+        },
+        errorMessage: `OFREP evaluation paused due to TooManyRequests until ${this.notBefore.toISOString()}`,
+      };
+    } else if (this.notBefore) {
+      this.notBefore = null;
+    }
```
lukas-reining, line 98: @beeme1mr @toddbaert @thomaspoignant 
I decided to short circuit evaluations on 429 if there is a value in the Retry-After Header. 
Would you change anything about this resolved value here? 

This is the last thing I would like to have validated with you, afterwards I would merge.

thomaspoignant: We may also throw here to let the SDK take care of the default value + the error code.

lukas-reining: Yes I thought about it, but then I thought this is a pretty special error that might be up to the provider 🤔

thomaspoignant: I am fine with both, I raised this because in @Kavindu-Dodan go provider an error is sent for this use case, and I would try to stay consistent as much as possible in the different server providers.

https://github.com/open-feature/go-sdk-contrib/blob/ab673a26abaf3fba11c6d28ce1b3ee476c97ab6e/providers/ofrep/internal/evaluate/resolver.go#L68

lukas-reining: Yes totally and for normal errors I would be totally with you regarding throwing.
Mh I am not sure what we want to do, maybe the others have an opinion?

toddbaert: I think we always have to throw right now for errors so the hooks run. I think this is the right thing to do with a 429 for server evals. Effectively this seems to me to be the same as any other error - for instance if the service was down.

So I would change this to throw a General with an informative error.

lukas-reining: Okay sounds fine to me @toddbaert.
I think I would just propagate the TooManyRequestsError as it will be converted to a GeneralError by the SDK.

Kavindu-Dodan: I went ahead with the error approach as this results in default evaluations. I think end users should act on 429 for example by increasing their evaluation quota at the vendor.

lukas-reining: I will change it to throwing then :)

lukas-reining: I changed this now.
@thomaspoignant @beeme1mr @toddbaert @Kavindu-Dodan any obligations of merging it now?

thomaspoignant: For me we are in a good state now.

PR #827, opened 2024-04-01
chore: lock configcat dep

toddbaert:
A [recent change](https://github.com/open-feature/js-sdk-contrib/pull/825) updated this configcat dep, which brought in a breaking change to out testing (noted in-line).

This PR locks the dep to the point-version so prevent test regressions.

We should probably find a better long-term solution.

### DIFF
--- a/libs/providers/config-cat/src/lib/config-cat-provider.spec.ts
+++ b/libs/providers/config-cat/src/lib/config-cat-provider.spec.ts
@@ -17,6 +17,7 @@ describe('ConfigCatProvider', () => {
   const targetingKey = 'abc';
 
   let provider: ConfigCatProvider;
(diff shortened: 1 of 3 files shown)


lukas-reining:
Hey @toddbaert, I am just looking into it. 
Should we merge it now and I find a solution later or should I try to solve it longterm before merging?

toddbaert:
> Hey @toddbaert, I am just looking into it. Should we merge it now and I find a solution later or should I try to solve it longterm before merging?

I opened this to get your opinion... IMO I think we should merge this now since it's blocking the build, and then figure out a better solution "soon".

lukas-reining (approved):

lukas-reining:
I am now looking inot a better solution @toddbaert, have been looking for it anyways right now :)

PR #830, opened 2024-04-01
fix(config-cat): remove event emitter type import from config-cat

lukas-reining:
## This PR

Removes the `IEventEmitter` import from configcat-common.
This reduces the risk of breaking the tests because of the removal of this type (export).

It tackles the same problem as #827. 

As there is currently no option to access the internal event emitter or provide it directly, the assumption getting the event emitter from the private options can not be removed.
https://github.com/open-feature/js-sdk-contrib/compare/fix/config-cat/remove-event-emitter-dep?expand=1#diff-ebd6f4363558237806382ec9b364f56831b69ad259c090f1981a4d56a7bb0bcdL43

### DIFF
--- a/libs/providers/config-cat/src/lib/config-cat-provider.spec.ts
+++ b/libs/providers/config-cat/src/lib/config-cat-provider.spec.ts
@@ -10,15 +10,13 @@ import {
   OverrideBehaviour,
   PollingMode,
 } from 'configcat-js-ssr';
-
-import { IEventEmitter } from 'configcat-common/lib/EventEmitter';
+import { EventEmitter } from 'events';
 
 describe('ConfigCatProvider', () => {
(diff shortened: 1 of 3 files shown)


lukas-reining:
@beeme1mr @toddbaert do you know why this error comes up now? The files are unchanged and are from flagd. 
But they are breaking the build, am I overlooking anything?

https://github.com/open-feature/js-sdk-contrib/actions/runs/8509661059/job/23305672676?pr=830#step:9:133

beeme1mr:
There's still an issue with the change detection. We should only be running tests on affected libraries. Hopefully, this PR resolves that issue.
https://github.com/open-feature/js-sdk-contrib/pull/831

beeme1mr (approved):

PR #836, opened 2024-04-02
feat: Add Flagsmith Provider

kyle-ssg:
## This PR
Adds a Flagsmith Provider

### Notes
- This has been tested and has corresponding examples in Standard, React, SSR and React Native frameworks. Sample applications provided [here](https://github.com/Flagsmith/flagsmith-js-examples/pull/9).
- Flagsmith separates remote configuration from the enabled state, therefore resolveBooleanEvaluation checks the enabled state and others read a feature's remote configuration.

### How to test
- Jest tests provided


kyle-ssg:
Didn't expect any lint issues, AFAIK I've resolved this [here](https://github.com/open-feature/js-sdk-contrib/pull/837)

libs/providers/flagsmith/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+npm install @openfeature/flagsmith-provider @openfeature/web-sdk
+```
+
+## Initialising the provider
+
+The Flagsmith Provider can be created with the standard [initialization options](https://docs.flagsmith.com/clients/javascript/#example-initialising-the-sdk) and an optional Flagsmith instance to use.
+
+```javascript
+import { FlagsmithProvider } from '@openfeature/flagsmith-provider';
```
beeme1mr, line 16:
```suggestion
import { FlagsmithProvider } from '@openfeature/flagsmith-provider';
import { OpenFeature } from '@openfeature/web-sdk';
```

libs/providers/flagsmith/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Usage with React Native, SSR or custom instances
+
+The Flagsmith Provider can be constructed with a custom Flagsmith instance and optional server-side generated state, [initialization options](https://docs.flagsmith.com/clients/javascript/#example-initialising-the-sdk).
+
+Note: In order to use the React Native implementation of OpenFeature you will need to install both Flagsmith and react-native-flagsmith.
+
+```javascript
+import flagsmith from 'react-native-flagsmith' // Could also be flagsmith/isomorphic, flagsmith-es or createFlagsmithInstance()
+import { FlagsmithProvider } from '@openfeature/flagsmith-provider';
```
beeme1mr, line 32:
```suggestion
import { FlagsmithProvider } from '@openfeature/flagsmith-provider';
import { OpenFeature } from '@openfeature/web-sdk';
```

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
```diff
@@ -0,0 +1,182 @@
...
+    ...config
+  }: Omit<IInitConfig, 'identity' | 'traits'> & { logger?: Logger; flagsmithInstance?: IFlagsmith }) {
+    this._logger = logger;
+    this._client = flagsmithInstance || createFlagsmithInstance();
+    this._config = config;
+  }
+
+  get status() {
+    return this._status;
+  }
```
beeme1mr, line 45: The SDK now manages the status automatically. You may be able to simplify the initialization logic significantly. FYI @toddbaert.

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
```diff
@@ -0,0 +1,182 @@
...
+      identity,
+      onChange: (previousFlags, params, loadingState) => {
+        this.status = ProviderStatus.READY;
+        this.events.emit(ProviderEvents.Ready, {
+          message: 'Flags ready',
+        });
+        if (params.flagsChanged) {
+          this.events.emit(ProviderEvents.ConfigurationChanged, {
+            message: 'Flags changed',
+          });
```
beeme1mr, line 84: You can include a list of the flags keys that changed if you have access to that information.

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
```diff
@@ -0,0 +1,182 @@
...
+        }
+        this._config.onChange?.(previousFlags, params, loadingState);
+      },
+      onError: (error) => {
+        this.status = ProviderStatus.ERROR;
+        this.errorHandler(error, 'Initialize');
+      },
```
beeme1mr, line 91: You should be able to reject the promise and let the OpenFeature SDK handle the rest.

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
kyle-ssg: Oh great, we do with a slight tweak upstream.

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
kyle-ssg: Ok yep that would be neater. I believe there are a couple cases we'd need to reject upstream for flagsmith.init but I'm in favour of doing that.

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
kyle-ssg: Ah gotcha, so this.events.emit adjusts the status ? That's great if so I did wonder why it didn't

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+   * @private
+   */
+  private parseReason(value: any, type: FlagType): ResolutionReason {
+    if (value === undefined) {
+      return 'DEFAULT';
+    }
+
+    if (typeof value !== type) {
+      return 'ERROR';
+    }
```
lukas-reining, line 128: I think for a type mismatch you can simply throw `TypeMismatchError` and let the SDK handle it.

lukas-reining (commented):
This looks really good! Added some comments about the provider name and type nits.

libs/providers/flagsmith/README.md
```diff
@@ -0,0 +1,93 @@
+# Flagsmith Provider
+
+This provider is an implementation for the [JavaScript SDK](https://docs.flagsmith.com/clients/javascript/) of [Flagsmith](https://flagsmith.com).
```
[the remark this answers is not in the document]
toddbaert: I think this is definitely up to the folks at Flagsmith, but I agree. It would be best to make this very clear to consumers this is for web.

I would recommend you use the same pattern for the dirs/filenames as well.

libs/providers/flagsmith/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Installation
+
+```
+npm install @openfeature/flagsmith-provider @openfeature/web-sdk
```
toddbaert, line 8: NPM will auotmatically install peers, and keep them up to data according to compatibility. Since you've added the web-sdk as a peer (or maybe our tooling did) I would recommend you remove the manual installation for NPM.

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+  ResolutionReason,
+} from '@openfeature/web-sdk';
+import { createFlagsmithInstance } from 'flagsmith';
+import { IFlagsmith, IInitConfig, IState } from 'flagsmith/types';
+import { FlagType, typeFactory } from './type-factory';
+
+export class FlagsmithProvider implements Provider {
+  readonly metadata: ProviderMetadata = {
+    name: FlagsmithProvider.name,
+  };
```
toddbaert, line 20:
```suggestion
  };
  
  readonly runsOn = 'client';
```

You can add this to have exceptions thrown if the provider is used by the wrong SDK. cc @beeme1mr

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
toddbaert: The SDK now maintains the state based on the success/failure of lifecycle methods (init/context-change) and emitted events.

The implementation looks good to me now; you can more or less not think about the state. Relevant [blog post section](https://openfeature.dev/blog/reconciling-with-state#wider-implications-stateless-providers).

toddbaert (approved):
Looks really good. Left some suggestions.

libs/providers/flagsmith/README.md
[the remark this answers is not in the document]
kyle-ssg: I did consider this but the way the provider works would also work for react-native.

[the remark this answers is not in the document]
lukas-reining: Client could also be fine if this is more the thing you want to be showing of :)

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
kyle-ssg: @beeme1mr So if client.initialize rejects should we be expecting ProviderEvents.Error to emit? 

If so, for some reason this fails for me 

```
 it('should call the error handler when errored', async () => {
      const errorHandler = jest.fn();
      OpenFeature.addHandler(ProviderEvents.Error, errorHandler);
      await OpenFeature.setProviderAndWait(new FlagsmithClientProvider({...invalidConfig}));
      expect(errorHandler).toHaveBeenCalledTimes(1);
    });
```

libs/providers/flagsmith/README.md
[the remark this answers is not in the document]
kyle-ssg: flagsmith-client it is 😃

beeme1mr:
Thanks @kyle-ssg, I'll rereview today. Would you like me to publish to npm if everything looks good?

libs/providers/flagsmith/README.md
kyle-ssg: Thanks, made me realise I should have added a flagsmith peer dep too.

kyle-ssg:
> Thanks @kyle-ssg, I'll rereview today. Would you like me to publish to npm if everything looks good?

I'm going to be doing a release of the flagsmith SDK to support a couple improvements here. I think I just need clarity on [this](https://github.com/open-feature/js-sdk-contrib/pull/836#discussion_r1553877272)

libs/providers/flagsmith/src/lib/flagsmith-provider.ts
kyle-ssg: I think I've figured it out, so it does do as suggested but in addition setProviderAndWait will also reject. Makes total sense!

kyle-ssg: Ignore, figured this out. Didn't anticipate that this happens but setAndWait would of course reject.

kyle-ssg:
I've implemented all of the suggested changes but I'm finding a bit of an issue with the provider that I'm unsure how to get around.

In attempting the following in an angular application

```
    OpenFeature.addHandler(ProviderEvents.ConfigurationChanged, handleFlags);
    OpenFeature.addHandler(ProviderEvents.Error, handleFlagsError);
    OpenFeature.setProvider(flagsmithClientProvider)
``

It looks like OpenFeature.addHandler(ProviderEvents.ConfigurationChanged) doesn't trigger if client.initialise is immediately ready (which is the case if cache is retrieved). Triggering a ProviderEvents.Ready event prior to the ProviderEvents.ConfigurationChanged has no effect. I'm guessing it's essentially due to initialize not resolving prior to the events being emitted.

Im aware that I could
- await setProviderAndWait but then I don't get feature metadata
- setup a 1 time ProviderEvents.Ready but similar issue
- set setTimeout(()=>{emitEvents}) in the provider but this will result in a flicker



https://github.com/open-feature/js-sdk-contrib/assets/8608314/26be4a59-f63a-45ed-bcb5-25a55581337e

File: libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts [hunk 10]
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  /**
+   * Based on Flagsmith's state, return flag metadata
+   * @private
+   */
+  private getMetadata() {
+    return {
+      targetingKey: this._client.identity || '',
+      ...(this._client.getAllTraits() || {}),
```
beeme1mr, line 116: Nice 👍

beeme1mr (approved):
Nice job, looks great!

toddbaert:
> I've implemented all of the suggested changes and we have a couple improvements upstream ready to help out but I'm finding a bit of an issue with the provider. I don't think it's a blocking issue as I've worked around it in the examples I've made, but this still caught me out.
> 
> In attempting the following in an angular application
> 
> ```
>     OpenFeature.addHandler(ProviderEvents.ConfigurationChanged, handleFlags);
>     OpenFeature.addHandler(ProviderEvents.Error, handleFlagsError);
>     OpenFeature.setProvider(flagsmithClientProvider)
> ```
> 
> It works for the most part except It looks like `ProviderEvents.ConfigurationChanged` doesn't trigger if client.initialise is immediately ready (which is the case if cache is retrieved). Triggering a `ProviderEvents.Ready` event prior to the `ProviderEvents.ConfigurationChanged` has no effect. I'm guessing it's essentially due to initialize not resolving prior to the events being emitted.
> 
> Im aware that I could do the following, but is there anything else?
> 
> * await setProviderAndWait but then I don't get feature metadata
> * setup a 1 time ProviderEvents.Ready but similar issue and it adds complexity to the consumer application
> * set setTimeout(()=>{emitEvents},0) in the provider. This does work but will result in a flicker
> * Detect that the provider is immediately ready by checking the status after `setProvider` but it seems there's no way to access event metadata.
> 
>  ConfigurationChanged.mov

Your assumptions here are all correct, and the solutions are valid. I'll think about this a bit... there may be something that can be done in the SDK but it seems like a bit of an edge case (is it expected or important that you get a change event immediately after becoming ready?)

I guess that in a real-world scenario, if this change event was indeed the result of a change in the remote system, it's just as likely that the change would have occurred before the provider is connected, in which case it would of course not be received. Is this just part of a test scenario?

toddbaert (approved):
I think this looks great.

toddbaert:
@kyle-ssg I am merging this. Nice work and thanks, as always. A release PR will be opened which I will not merge until you approve it.

kyle-ssg:
> Your assumptions here are all correct, and the solutions are valid. I'll think about this a bit... there may be something that can be done in the SDK but it seems like a bit of an edge case (is it expected or important that you get a change event immediately after becoming ready?)
> 
> I guess that in a real-world scenario, if this change event was indeed the result of a change in the remote system, it's just as likely that the change would have occurred before the provider is connected, in which case it would of course not be received. Is this just part of a test scenario?

Yes and no, it was part of an example app where I rely on changes to push to an array of logs to demonstrate OpenFeature.  With how things work currently I think all apps will have to just treat the first ready as a change rather than relying on a single event. I think what I was trying to do relied to heavily on the meta that gets returned in the events, really if people are doing that they may as well just query the provider directly if they need info from it.

kyle-ssg:
> @kyle-ssg I am merging this. Nice work and thanks, as always. A release PR will be opened which I will not merge until you (or Flagsmith other folks) approve it.

Awesome, I'm excited to use it!

PR #841, opened 2024-04-03
feat(ofrep): parse Retry-After header in OFREPApiTooManyRequestsError

lukas-reining:
## This PR

Implements parsing the retry after header in the `OFREPApiTooManyRequestsError ` as client and server provider need that information.
It can replace the current implementation in the web-provider. https://github.com/open-feature/js-sdk-contrib/pull/776#discussion_r1550601232


thomaspoignant (approved):
Good idea to put in ofrep-core.

PR #882, opened 2024-04-16
fix: fixes an issue where OFREP does not send content type headers

lukas-reining:
## This PR

-  fixes an issue where OFREP does not send content type headers


thomaspoignant (approved):
Should we add a unit test for this?

PR #896, opened 2024-05-01
feat: Initial GrowthBook OpenFeature provider

msamper:
## This PR

- adds a new OpenFeature client provider for the GrowthBook JS SDK

### Follow-up Tasks

- [x] Update README with GrowthBook provider specific information

### How to test

Integrate OpenFeature into a test web app and set it up with this new GrowthBook provider. Ensure that GrowthBook flag evaluation works as expected using the OpenFeature API

[How to set up OpenFeature on a React app](https://openfeature.dev/docs/reference/technologies/client/web/react)

```
// Pass in gbContext the same way you would if you were instantiating GrowthBook directly
OpenFeature.setProvider(new GrowthbookClientProvider(gbContext))
```

### DIFF
--- a/libs/providers/growthbook-client/README.md
+++ b/libs/providers/growthbook-client/README.md
@@ -1,0 +1,44 @@
+# growthbook-client Provider
+
+## Installation
+
+```
+$ npm install @openfeature/growthbook-client-provider
+```
+
+## Example Setup
+
+```typescript
+import { GrowthBook, Context, InitOptions } from '@growthbook/growthbook';
+import { GrowthbookClientProvider } from '@openfeature/growthbook-client-provider';
+
+/*
+ * Configure your GrowthBook instance with GrowthBook context
+ * @see https://docs.growthbook.io/lib/js#step-1-configure-your-app
+ */
+const gbContext: Context = {
+  apiHost: 'https://cdn.growthbook.io',
+  clientKey: 'sdk-abc123',
+  // Only required if you have feature encryption enabled in GrowthBook
+  decryptionKey: 'key_abc123',
+};
+
+/*
+ * optional init options
+ * @see https://docs.growthbook.io/lib/js#switching-to-init
+ */
+const initOptions: InitOptions = {
+  timeout: 2000,
+  streaming: true,
+};
+
+OpenFeature.setProvider(new GrowthbookClientProvider(gbContext, initOptions));
+```
+
+## Building
+
+Run `nx package providers-growthbook-client` to build the library.
+
+## Running unit tests
+
+Run `nx test providers-growthbook-client` to execute the unit tests via [Jest](https://jestjs.io).
--- a/libs/providers/growthbook-client/src/lib/growthbook-client-provider.ts
+++ b/libs/providers/growthbook-client/src/lib/growthbook-client-provider.ts
@@ -1,0 +1,90 @@
+import {
+  EvaluationContext,
+  Provider,
+  JsonValue,
+  ResolutionDetails,
+  GeneralError,
+  OpenFeatureEventEmitter,
+  ProviderEvents,
+} from '@openfeature/web-sdk';
+
+import { GrowthBook, InitOptions, Context } from '@growthbook/growthbook';
+import isEmpty from 'lodash.isempty';
+import translateResult from './translate-result';
+
+export class GrowthbookClientProvider implements Provider {
+  metadata = {
+    name: GrowthbookClientProvider.name,
+  };
+
+  readonly runsOn = 'client';
+  private _client?: GrowthBook;
+  private readonly context: Context;
+  private _initOptions?: InitOptions;
+  public readonly events = new OpenFeatureEventEmitter();
+
+  constructor(growthbookContext: Context, initOptions?: InitOptions) {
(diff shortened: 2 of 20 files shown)


libs/providers/growthbook-client/src/lib/growthbook-client-provider.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  resolveStringEvaluation(flagKey: string, defaultValue: string): ResolutionDetails<string> {
+    const value = this.client.getFeatureValue(flagKey, defaultValue);
+    const featureResult = this.client.evalFeature(flagKey);
+
+    const resolution: ResolutionDetails<string> = {
+      value,
+      reason: featureResult.source,
+      variant: featureResult.experimentResult?.variationId?.toString(),
```
msamper, line 49: `result.key` instead

libs/providers/growthbook-client/src/lib/growthbook-client-provider.ts
```diff
@@ -0,0 +1,77 @@
...
+    };
+    return resolution;
+  }
+
+  resolveStringEvaluation(flagKey: string, defaultValue: string): ResolutionDetails<string> {
+    const value = this.client.getFeatureValue(flagKey, defaultValue);
+    const featureResult = this.client.evalFeature(flagKey);
```
msamper, line 44: Only call `evalFeature` and use logic from within `getFeatureValue ` for `defaultValue` passed in

libs/providers/growthbook-client/src/lib/growthbook-client-provider.ts
```diff
@@ -0,0 +1,82 @@
+import { EvaluationContext, Provider, JsonValue, ResolutionDetails, GeneralError } from '@openfeature/web-sdk';
+
+import { FeatureResult, GrowthBook, InitOptions, Context } from '@growthbook/growthbook';
+import isEmpty from 'lodash.isempty';
+
+export default function translateResult<T>(result: FeatureResult, defaultValue: T): ResolutionDetails<T> {
+  const resolution: ResolutionDetails<T> = {
+    value: result.value === null ? defaultValue : result.value,
+    reason: result.source,
```
beeme1mr, line 9: It looks like `source` can represent various error conditions. For example, I'm assuming `unknownFeature` means the SDK couldn't find the request feature flag. Could you please try and map those conditions to approprate `ErrorCode`?

libs/providers/growthbook-client/README.md
```diff
@@ -0,0 +1,43 @@
...
+  decryptionKey: 'key_abc123',
+};
+
+/*
+ * optional init options
+ * @see https://docs.growthbook.io/lib/js#switching-to-init
+ */
+const initOptions: InitOptions = {
+  timeout: 2000,
+  streaming: true,
```
beeme1mr, line 31: The provider will need to emit a change event for streaming changes to work. Here's an example of how it could look.

https://github.com/open-feature/js-sdk/blob/main/packages/client/src/provider/in-memory-provider/in-memory-provider.ts#L58

beeme1mr (commented):
Thanks for the PR. I've left some feedback but overall it looks good. Ping me if you have any questions.

lukas-reining:
Thanks @msamper, I will have a look in the next days! :)

beeme1mr (approved):

libs/providers/growthbook-client/src/lib/translate-result.ts
```diff
@@ -0,0 +1,29 @@
...
+    reason: result.source,
+    variant: result.experimentResult?.key,
+  };
+
+  if (FEATURE_RESULT_ERRORS.includes(result.source)) {
+    resolution.errorCode = translateError(result.source);
+  }
+
+  return resolution;
+}
```
lukas-reining, line 29: This function takes a `FeatureResult` as argument which is a `FeatureResult<any>`, so `result.value` is any, which means the types returned by this function are not save.

I added the typical check in the suggested change.
Another option would be to try to convert the types, but the common way is to throw a `TypeMismatchError` which will  result in the default value being returned by the client.

If you accept the change, please add a small test case @msamper.

```suggestion
export default function translateResult<T>(result: FeatureResult<unknown>, defaultValue: T): ResolutionDetails<T> {
  if (result.value !== null && typeof result.value !== typeof defaultValue) {
    throw new TypeMismatchError(`Expected flag type ${typeof defaultValue} but got ${typeof result.value}`);
  }

  const resolution: ResolutionDetails<T> = {
    value: result.value === null ? defaultValue : (result.value as T),
    reason: result.source,
    variant: result.experimentResult?.key,
  };

  if (FEATURE_RESULT_ERRORS.includes(result.source)) {
    resolution.errorCode = translateError(result.source);
  }

  return resolution;
}
```

lukas-reining (commented):
It looks great to me @msamper!
Please take a look at the return type validation, after that we can merge it :)

msamper:
@lukas-reining Thanks for your comments! Addressed and added sign-off so this should hopefully be ready to merge :)

lukas-reining (approved):
Looks really good, thanks @msamper!
I will merge that now :)

PR #916, opened 2024-05-23
feat: multi provider implementation for Node

ajwootto:
## This PR

- adds a Node implementation of a "multi provider" which can orchestrate evaluating from multiple other providers while appearing as a single provider from the perspective of the OpenFeature SDK.
- design considerations are currently documented on this public Notion page https://taplytics.notion.site/OpenFeature-Multi-Provider-1014926cb35840b9b7b5756a71b9b76d

### Notes
- based on the original "vendor migration" PR: https://github.com/open-feature/js-sdk-contrib/pull/613
### Follow-up Tasks

- figure out how the design description should be reformatted and added anywhere in the spec or documentation

### How to test

`npx nx test providers-multi-provider`


beeme1mr (commented):
I haven't had a chance to run this locally, but I wanted to provide feedback before the weekend. I noticed a few minor things; overall, it looks great. It's more complex than I anticipated but should be extremely powerful.

Monday is a holiday, so I probably won't have a chance to test it end-to-end until Tuesday.

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,325 @@
...
+
+  public readonly events = new OpenFeatureEventEmitter();
+
+  private hookContexts: WeakMap<EvaluationContext, HookContext> = new WeakMap<EvaluationContext, HookContext>();
+  private hookHints: WeakMap<EvaluationContext, HookHints> = new WeakMap<EvaluationContext, HookHints>();
+
+  metadata: ProviderMetadata;
+
+  providerEntries: RegisteredProvider[] = [];
+  private providerEntriesByName: Record<string, RegisteredProvider> = {};
```
lukas-reining, line 44: I would prefer `private providerEntriesByName: ReadonlyMap<string, RegisteredProvider> = {};` here.
This way we would also not have to have `providerEntries` maintained, we could just have 

```ts
  public get providerEntries(): RegisteredProvider[] {
    return [...this.providerEntriesByName.values()];
  }
```
which is less error prone, in the case we will later maybe update `providerEntriesByName`.
The only problem is, `ReadonlyMap` only applies to TS and we can not freeze the Map.

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,325 @@
...
+
+    for (const constructorProvider of constructorProviders) {
+      const providerName = constructorProvider.provider.metadata.name;
+      let candidateName = constructorProvider.name ?? providerName;
+
+      if (constructorProvider.name && providersByName[constructorProvider.name]) {
+        throw new Error('Provider names must be unique');
+      }
+
+      providersByName[candidateName] ||= [];
```
lukas-reining, line 79:
```suggestion
      providersByName[candidateName] ??= [];
```

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,325 @@
...
+      }
+    }
+
+    // just make sure we don't accidentally modify these later
+    Object.freeze(this.providerEntries);
+    Object.freeze(this.providerEntriesByName);
+  }
+
+  async initialize(context?: EvaluationContext): Promise<void> {
+    await Promise.all(this.providerEntries.map((provider) => provider.provider.initialize?.(context)));
```
lukas-reining, line 99: What is the expected behavior if one of the providers fails to initialize?
Is it desired that this will throw directly then?

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  async onClose(): Promise<void> {
+    await Promise.all(this.providerEntries.map((provider) => provider.provider.onClose?.()));
```
lukas-reining, line 103: Maybe we can do `Promise.allSettled` here? This way every provider will be closed, even if one fails to do so.

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,325 @@
...
+      thrownError = error;
+    }
+
+    if (this.evaluationStrategy.runMode === 'sequential') {
+      if (
+        !this.evaluationStrategy.shouldEvaluateNextProvider(strategyContext, context, evaluationResult, thrownError)
+      ) {
+        return false;
+      }
+    }
```
lukas-reining, line 240:
```suggestion
    if (this.evaluationStrategy.runMode === 'sequential') {
      return this.evaluationStrategy.shouldEvaluateNextProvider(
        strategyContext,
        context,
        evaluationResult,
        thrownError,
      );
    }
```

libs/providers/multi-provider/src/lib/strategies/BaseEvaluationStrategy.ts
```diff
@@ -0,0 +1,92 @@
...
+    context: EvaluationContext,
+    resolutions: ProviderResolutionResult<T>[],
+  ): FinalResult<T>;
+
+  protected hasError(resolution: ProviderResolutionResult<FlagValue>): boolean {
+    return 'thrownError' in resolution || !!resolution.details?.errorCode;
+  }
+
+  protected collectProviderErrors<T extends FlagValue>(resolutions: ProviderResolutionResult<T>[]): FinalResult<T> {
+    let errors: FinalResult<FlagValue>['errors'] = [];
```
lukas-reining, line 74:
```suggestion
    const errors: FinalResult<FlagValue>['errors'] = [];
```

libs/providers/multi-provider/src/lib/strategies/BaseEvaluationStrategy.ts
```diff
@@ -0,0 +1,92 @@
...
+  providerName: string;
+  providerStatus: ProviderStatus;
+};
+
+export type ProviderResolutionResult<T extends FlagValue> = {
+  provider: Provider;
+  providerName: string;
+  thrownError?: unknown;
+  details?: ResolutionDetails<T>;
+};
```
lukas-reining, line 26: Instead of `thrownError` and `details` being optional, consider something like the following (Not sure if the names are exactly right)

````ts
export type ProviderResolutionBaseResult = {
  provider: Provider;
  providerName: string;
};

export type ProviderResolutionSuccessResult<T extends FlagValue> = ProviderResolutionBaseResult & {
  details: ResolutionDetails<T>;
};

export type ProviderResolutionErrorResult = ProviderResolutionBaseResult & {
  thrownError: unknown;
};

export type ProviderResolutionResult<T extends FlagValue> =
  | ProviderResolutionErrorResult
  | ProviderResolutionSuccessResult<T>;
````

This way we can avoid type assertions like here

File: libs/providers/multi-provider/src/lib/strategies/ComparisonStrategy.ts [hunk 8]
```diff
@@ -0,0 +1,57 @@
...
+    let fallbackResolution: ProviderResolutionResult<T> | undefined;
+    let mismatch = false;
+    for (const resolution of resolutions) {
+      if (resolution.thrownError || resolution.details?.errorCode) {
+        return this.collectProviderErrors(resolutions);
+      }
+      if (resolution.provider === this.fallbackProvider) {
+        fallbackResolution = resolution;
+      }
+      if (typeof value !== 'undefined' && value !== resolution.details!.value) {
```
lukas-reining, line 40: We can avoid the non-null assertion if we use types as described in the comment in `BaseEvaluationStrategy.ts`

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,325 @@
...
+      const task = this.evaluateProviderEntry(
+        flagKey,
+        flagType,
+        defaultValue,
+        providerEntry,
+        hookContext,
+        hookHints,
+        context,
+        resolutions,
+      );
```
lukas-reining, line 164: I am a bit worried about `resolutions` being passed as parameter, but being mutated.
Maybe we can return the the new resolutions `List` and a `boolean` for `shouldEvaluateNext` as a tuple or so.
Especially as the resolutions are even out-of-order in the case of parallel evaluation.

lukas-reining (commented):
Thanks @ajwootto this looks good!
I left some comments and thoughts.
I will have another look at the `evaluateProviderEntry` tomorrow or on Monday, maybe we can simplify the code a bit here as much is going on there.

libs/providers/multi-provider/src/lib/multi-provider.ts
ajwootto: yeah, right now it operates under the assumption that the user would want all providers to successfully be initialized before proceeding to a "ready" state

lukas-reining: Okay :)

libs/providers/multi-provider/src/lib/multi-provider.ts
lukas-reining: Ah good point, makes sense!

libs/providers/multi-provider/src/lib/strategies/BaseEvaluationStrategy.ts
lukas-reining: For me it worked 🤔

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,325 @@
...
+  ) {
+    switch (typeof defaultValue) {
+      case 'string':
+        return await provider.resolveStringEvaluation(flagKey, defaultValue, context, this.logger);
+      case 'number':
+        return await provider.resolveNumberEvaluation(flagKey, defaultValue, context, this.logger);
+      case 'object':
+        return await provider.resolveObjectEvaluation(flagKey, defaultValue, context, this.logger);
+      case 'boolean':
+        return await provider.resolveBooleanEvaluation(flagKey, defaultValue, context, this.logger);
```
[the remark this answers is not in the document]
ajwootto: technically no, this is just a stylistic thing we use in our own codebase to make it easier to tell when a function is calling another async operation vs just returning a value. I can remove the awaits if that's not desired though

libs/providers/multi-provider/src/lib/multi-provider.ts
ajwootto: Updated this to use a tuple of `[shouldEvaluateNext, resolutionResult]`

libs/providers/multi-provider/src/lib/strategies/BaseEvaluationStrategy.ts
ajwootto: got it working, the only caveat is that I couldn't get the types of the `hasError` check in `determineResult` to narrow well enough, and ended up moving that logic inline

libs/providers/multi-provider/src/lib/multi-provider.ts
ajwootto: updated to use `allSettled` for both "onClose" and "initialize". It will also construct an "aggregate error" out of the error results from each provider

libs/providers/multi-provider/src/lib/errors.ts
```diff
@@ -0,0 +1,53 @@
+import { ErrorCode, GeneralError, OpenFeatureError } from '@openfeature/server-sdk';
+import { PromiseAllSettledResult } from '@opentelemetry/sdk-metrics/build/esnext/utils';
```
lukas-reining, line 2: @ajwootto I think we should not import this here.
We can use the original type here `PromiseSettledResult<Awaited<unknown>>[]` without any import.

libs/providers/multi-provider/src/lib/errors.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // log first error in the message for convenience, but include all errors in the error object for completeness
+  return new AggregateError(
+    `Provider errors occurred: ${errorsWithSource[0].source}: ${errorsWithSource[0].error}`,
+    errorsWithSource,
+  );
+};
+
+export const throwAggregateErrorFromPromiseResults = (
+  result: PromiseAllSettledResult<unknown>[],
```
lukas-reining, line 38:
```suggestion
  result: PromiseSettledResult<Awaited<unknown>>[],
```

libs/providers/multi-provider/src/lib/multi-provider.ts
lukas-reining: Great, just added a comment regarding the AllSettled type :)

libs/providers/multi-provider/src/lib/strategies/ComparisonStrategy.ts
```diff
@@ -0,0 +1,58 @@
...
+      } else {
+        value = resolution.details.value;
+      }
+    }
+
+    if (mismatch) {
+      this.onMismatch?.(resolutions);
+      return {
+        details: fallbackResolution!.details,
+        provider: fallbackResolution!.provider,
```
lukas-reining, line 52: These should not be `null-asserted`.
Yes, they will always be there because the fallback provider will always be set, but this will be hard to track if something about the logic changes and the `null-assertion` not holds true.

libs/providers/multi-provider/src/lib/strategies/ComparisonStrategy.ts
```diff
@@ -0,0 +1,58 @@
...
+    private fallbackProvider: Provider,
+    private onMismatch?: (resolutions: ProviderResolutionResult<FlagValue>[]) => void,
+  ) {
+    super();
+  }
+
+  override determineFinalResult<T extends FlagValue>(
+    strategyContext: StrategyPerProviderContext,
+    context: EvaluationContext,
+    resolutions: ProviderResolutionSuccessResult<T>[],
```
lukas-reining, line 29:
```suggestion
    resolutions: ProviderResolutionResult<T>[],
```

This should be the union type result.

libs/providers/multi-provider/src/lib/strategies/BaseEvaluationStrategy.ts
lukas-reining: Cool!
Well it is possible, if you write the `hasError` function this way:

```ts
  protected hasError(resolution: ProviderResolutionResult<FlagValue>): resolution is
    | ProviderResolutionErrorResult
    | (ProviderResolutionSuccessResult<FlagValue> & {
        details: ResolutionDetails<FlagValue> & { errorCode: ErrorCode };
      }) {
    return 'thrownError' in resolution || !!resolution.details.errorCode;
  }
```

This works, tested it, but the type is ugly :D Sure one could define that type where `ProviderResolutionResult` is defined, but maybe it is just fine to do it inline. I am not sure.

libs/providers/multi-provider/src/lib/errors.ts
ajwootto: ah yes the perils of Webstorm auto-import

libs/providers/multi-provider/src/lib/multi-provider.ts
```diff
@@ -0,0 +1,315 @@
...
+
+      const afterHookEvalDetails = {
+        ...evaluationResult,
+        flagMetadata: Object.freeze(evaluationResult.flagMetadata ?? {}),
+        flagKey,
+      };
+
+      await this.hookExecutor.afterHooks(
+        provider.hooks,
+        { ...hookContext, context: providerContext },
```
beeme1mr, line 262: This makes it so provider hooks can't use WeakMaps. Would it be possible to use the same object reference for the hook context so that functionality could persist?

beeme1mr (commented):
This looks great so far. I have one question about WeakMap support, but it's probably an edge case. I'll approve once it's no longer a draft PR and the readme has been updated.

libs/providers/multi-provider/src/lib/multi-provider.ts
beeme1mr: Here's the code I used to test this.

```ts
import { OpenFeature, InMemoryProvider, Hook, BeforeHookContext, EvaluationContext, EvaluationDetails, FlagValue, HookContext } from "@openfeature/server-sdk";
import { MultiProvider, FirstMatchStrategy } from "@openfeature/multi-provider";

class WeakMapHook implements Hook {
  private map = new WeakMap<WeakKey, boolean>();

  before(hookContext: BeforeHookContext, hookHints?: Readonly<Record<string, unknown>> | undefined): void | EvaluationContext | Promise<void | EvaluationContext> {
      this.map.set(hookContext, true);
  }

  after(hookContext: Readonly<HookContext<FlagValue>>, evaluationDetails: EvaluationDetails<FlagValue>, hookHints?: Readonly<Record<string, unknown>> | undefined): void | Promise<void> {
    if (this.map.has(hookContext)) {
      console.log("WeakMapHook executed");
    } else {
      console.log("WeakMapHook not executed");
    }
  }
}

class CustomInMemoryProvider extends InMemoryProvider {
  hooks = [
    new WeakMapHook(),
  ]

  constructor(flagConfig: ConstructorParameters<typeof InMemoryProvider>[0]) {
    super(flagConfig);
  }
}

await OpenFeature.setProviderAndWait(
  new MultiProvider([
    {
      provider: new InMemoryProvider({
        test1: {
          variants: {
            on: true,
            off: false,
          },
          defaultVariant: "on",
          disabled: false,
        },
      }),
      name: "provider1",
    },
    {
      provider: new InMemoryProvider({
        test2: {
          variants: {
            on: true,
            off: false,
          },
          defaultVariant: "on",
          disabled: false,
        },
      }),
      name: "provider2",
    },
    {
      provider: new CustomInMemoryProvider({
        test3: {
          variants: {
            on: true,
            off: false,
          },
          defaultVariant: "on",
          disabled: false,
        },
      }),
      name: "provider3",
    }
  ], new FirstMatchStrategy())
);

const client = OpenFeature.getClient();

console.log(await client.getBooleanDetails("test", false));
console.log(await client.getBooleanDetails("test1", false));
console.log(await client.getBooleanDetails("test2", false));
console.log(await client.getBooleanDetails("test3", false));
```

toddbaert: This is an important point, I think we need to use the same object instance.

ajwootto: updated to keep the reference consistent between hook calls of an individual provider. In order to do that I had to make a copy of the hook context + eval context for each provider's evaluation, so the object is consistent for an individual provider but is not consistent _across_ providers.

libs/providers/multi-provider/README.md
```diff
@@ -0,0 +1,136 @@
...
+new provider, the Multi-Provider will automatically find and return them, falling back to the old provider if the new provider does not have
+- *Multiple Data Sources*: The Multi-Provider allows you to seamlessly combine many sources of flagging data, such as environment variables, 
+local files, database values and SaaS hosted feature management systems.
+
+## Installation
+
+```
+$ npm install @openfeature/multi-provider
+```
+
```
beeme1mr, line 19:
```suggestion
> [!TIP]
> This SDK is designed to be used with the [Node.js SDK](https://openfeature.dev/docs/reference/technologies/server/javascript/).

```

beeme1mr (approved):
This looks great, thanks! Even though you mention it in the intro, I recommend adding a tip that explicitly calls out that it's intended for the Node.js SDK.

It also looks like the CI isn't working again due to not being able to find an NX binary. Regenerating the package-lock.json seems to resolve the issue.

libs/providers/multi-provider/README.md
```diff
@@ -0,0 +1,139 @@
...
+> [!TIP]
+> This SDK is designed to be used with the [Node.js SDK](https://openfeature.dev/docs/reference/technologies/server/javascript/).
```
lukas-reining, line 21:
```suggestion
> This provider is designed to be used with the [Node.js SDK](https://openfeature.dev/docs/reference/technologies/server/javascript/).
```

lukas-reining (approved):
Looks great to me, just added one possible correction for the README :)

libs/providers/multi-provider/README.md
beeme1mr: Whoop! Good catch @lukas-reining

libs/providers/multi-provider/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@openfeature/multi-provider",
```
beeme1mr, line 2: I wonder if we should add `server` in this name to make it clear that it won't work on the web.

ajwootto: I was kind of following the naming conventions I saw in the rest of the repo, where the "server" version of things didn't specify server, but the "web" version did specify "web". I assume that's because most of the server stuff was created before "web" existed. Should I go with consistency with the others, or just name it with "server" for clarity?

beeme1mr:
The test failure is unreleased to this change and caused by a brittle test and upgrades to the web sdk. We'll fix the test failures in a follow up PR. Thanks, @ajwootto, for this great contribution! 🥳

Issue #816, opened 2024-03-23
feat: ConfigCat web provider


lukas-reining:
We should add a ConfigCat web provider.
The discussion started in a PR of @adams85.

> Thanks for the release, @lukas-reining. I did a quick test and found that the new version (v0.6.0) works fine.
> 
> > Maybe we should also consider developing a client-sdk provider, as this is a server sdk provider. Which makes you change even more important.
> 
> Since our SSR SDK can handle both targets, I guess the client provider would be more or less just a duplicate of the server one (at least, in our case). I'd avoid maintaining another package if possible.
> 
> BTW, NPM packages kind of support multitargeting via [the `browser` field](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#browser). AFAIK, most of the bundlers support it nowadays. I have plans to look into the viability of this approach because it would be a godsend to us if we could merge our various JS SDK packages (or at least some) into one.
> 
> _Originally posted by @adams85 in https://github.com/open-feature/js-sdk-contrib/issues/796#issuecomment-2010312985_

@adams85 I am fully with you trying to not have a second provider codebase for the web that is just a copy of the server SDK, but in my opinion this is not the case for OpenFeature.

## Problem

The provider interface of the `@openfeature/web-sdk` is different from the one in the `@openfeature/server-sdk`.
Mainly the difference from the API side is that the web SDK evaluates synchronous while the server SDK evaluates async.
This is due to the web SDK using the `static context paradigm` while the web SDK uses the `dynamic context paradigm`.
The reasoning for this can be seen at Petes article in our blog [https://openfeature.dev/blog/catering-to-the-client-side](https://openfeature.dev/blog/catering-to-the-client-side).

## Solution

To reduce the amount of duplicate code we should move [context-tranformer.ts](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/config-cat/src/lib/context-transformer.ts) and utilities like `toResolutionDetails` from [config-cat-provider.ts](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/config-cat/src/lib/config-cat-provider.ts) to a shared folder. For the server there would be noting more left to do.

[...]

PR #918, opened 2024-05-25
feat!: implement config cat web provider

lukas-reining:
## This PR

Implements ConfigCat Web Provider as discussed here #816.
It mainly moves the context and result mapping logic to a shared package and implements the usage of snapshots for the web-sdk as discussed with @adams85.


lukas-reining:
@adams85 it took some time to find time to implement, but here it is, as we discussed in #816.
I could not add you as a reviewer, because you are not part of the org. If you like to join the org, I can add you :) 
Would be happy about a review.

libs/shared/config-cat-core/src/lib/context-transformer.ts
```diff
@@ -49,11 +49,11 @@ function stringOrUndefined(param?: unknown): string | undefined {
   return undefined;
 }
 
-export function transformContext(context: EvaluationContext): ConfigCatUser | never {
+export function transformContext(context: EvaluationContext): ConfigCatUser | undefined {
```
adams85, line 52: The `transformContextValues` function should be revised as newer versions of ConfigCat SDKs [don't require string values for custom user attribute values](https://github.com/configcat/common-js/blob/v9.3.0/src/User.ts#L12-L51) any more.

Custom attribute values should be passed to the SDK as is to let it deal with them internally. But at least, `Date` objects shouldn't be converted to ISO strings because that won't work with ConfigCat's BEFORE/AFTER date comparators.

libs/shared/config-cat-core/src/lib/result-mapping.ts
```diff
@@ -0,0 +1,44 @@
...
+  data: Omit<IEvaluationDetails, 'value'>,
+  reason?: ResolutionReason,
+): ResolutionDetails<PrimitiveType<T>> {
+  if (!isType(type, value)) {
+    throw new TypeMismatchError(`Requested ${type} flag but the actual value is ${typeof value}`);
+  }
+
+  const matchedTargeting = 'matchedEvaluationRule' in data ? data.matchedEvaluationRule : data.matchedTargetingRule;
+  const matchedPercentage =
+    'matchedEvaluationPercentageRule' in data ? data.matchedEvaluationPercentageRule : data.matchedPercentageOption;
```
adams85, line 16: If we do the suggested package version upgrades, we won't need to check for old property names.

```suggestion
  const matchedTargeting = data.matchedTargetingRule;
  const matchedPercentage = data.matchedPercentageOption;
```

libs/shared/config-cat-core/src/lib/result-mapping.ts
```diff
@@ -0,0 +1,44 @@
...
+    : T extends 'number'
+      ? number
+      : T extends 'object'
+        ? object
+        : T extends 'undefined'
+          ? undefined
+          : unknown;
+
+export function isType<T extends PrimitiveTypeName>(type: T, value: unknown): value is PrimitiveType<T> {
+  return value !== 'undefined' && typeof value === type;
```
adams85, line 43: `value !== 'undefined'` doesn't make sense to me, probably a `typeof` is missing here.

```suggestion
  return typeof value !== 'undefined' && typeof value === type;
```

libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.events.emit(ProviderEvents.Error, {
+          message: message,
+          metadata: error,
+        });
+      });
+    };
+
+    const client = getClient(...originalParameters);
+
+    if (pollingMode !== PollingMode.AutoPoll) {
```
adams85, line 78: I don't think we should allow polling modes other than Auto Polling. As the docs point out:

> Please note that when you create and utilize a snapshot, it won't refresh your local cache once the cached config data expires. Additionally, when working with [shared caching](https://configcat.com/docs/advanced/caching/#shared-cache), creating a snapshot also doesn't trigger a sync with the external cache, since the snapshot only captures the config instance stored in the client's memory. Therefore, it's recommended to use snapshots in conjunction with the Auto Polling mode, where the SDK automatically updates the local cache in the background. For other polling modes, you'll need to manually initiate a cache refresh by invoking forceRefreshAsync.

That is, to make snapshots actually work with Manual Polling or Lazy Loading, we'd need to implement a background job that keeps the SDK's cache up to date. IMO that wouldn't make much sense as it would be pretty much a re-implementation of what Auto Polling already does.

So we should either throw an exception or ignore the polling mode option if it's not set to Auto Polling. Probably throwing an exception would be the more transparent behavior to the user.

adams85 (commented):
Hey @lukas-reining,

Awesome news! I did a quick review.

libs/shared/config-cat-core/src/lib/result-mapping.ts
lukas-reining: yes typo, thanks!

libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
lukas-reining: Makes sense, will update that later or tomorrow @adams85 :)

libs/shared/config-cat-core/src/lib/context-transformer.ts
lukas-reining: Oh did not see that, will also update that! :)

libs/providers/config-cat-web/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```
+$ npm install @openfeature/config-cat-web-provider
+```
+
+#### Required peer dependencies
+
+The OpenFeature SDK is required as peer dependency.
+
+The minimum required version of `@openfeature/client-sdk` currently is `1.0.0`.
```
beeme1mr, line 15:
```suggestion
The minimum required version of `@openfeature/web-sdk` currently is `1.0.0`.
```

libs/providers/config-cat/src/lib/config-cat-provider.ts
```diff
@@ -33,50 +39,37 @@ export class ConfigCatProvider implements Provider {
...
-            metadata: error,
-          });
+    const options = originalParameters[2];
+    const oldSetupHooks = options.setupHooks;
+
+    options.setupHooks = (hooks) => {
+      oldSetupHooks?.(hooks);
+
+      hooks.on('clientReady', () => {
+        this.events.emit(ProviderEvents.Ready);
```
toddbaert, line 53: This may result in duplicate READY events, since the SDK will emit one when this function completes, and also due to this handler.

@kinyoklion opened an issue discussing this [here](https://github.com/open-feature/spec/issues/265).

toddbaert:
[This](https://github.com/open-feature/js-sdk-contrib/pull/918/files#r1643121053) is my only concern.

libs/providers/config-cat/src/lib/config-cat-provider.ts
lukas-reining: Makes sense, how do we continue on this @toddbaert?

toddbaert: We can either implement something as @kinyoklion describes or we can just add some logic to prevent the emissions of the initial event.

lukas-reining:
Hey @adams85, took me quite some time to get to it but I just change the two things you proposed :) 
Would be happy about a re-review!
I changed the mapping to support your new typings for the User object and only allow `AutoPoll`for the web sdk.
Marking this as breaking is the User mapping changed.

libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
```diff
@@ -0,0 +1,165 @@
...
+  public async initialize(): Promise<void> {
+    const options = this._configCatOptions;
+    const oldSetupHooks = options.setupHooks;
+
+    options.setupHooks = (hooks) => {
+      oldSetupHooks?.(hooks);
+
+      hooks.on('clientReady', () => {
+        this.events.emit(ProviderEvents.Ready);
+      });
```
adams85, line 61: As I described above, subscribing to `clientReady` is problematic. I suggest the same solution (i.e. moving the emitting of `ProviderEvents.Ready` after `waitForReady`).

```suggestion
```

libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
```diff
@@ -0,0 +1,165 @@
...
+        });
+      });
+    };
+
+    const client = getClient(this._sdkKey, PollingMode.AutoPoll, this._configCatOptions);
+    await client.waitForReady();
+
+    this._client = client;
+    this._clientSnapshot = client.snapshot();
+  }
```
adams85, line 86:
```suggestion
    this.events.emit(ProviderEvents.Ready);
  }
```

libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
```diff
@@ -0,0 +1,165 @@
...
+          metadata: error,
+        });
+      });
+    };
+
+    const client = getClient(this._sdkKey, PollingMode.AutoPoll, this._configCatOptions);
+    await client.waitForReady();
+
+    this._client = client;
+    this._clientSnapshot = client.snapshot();
```
adams85, line 85: I advise against storing the snapshot as that would break advanced setups using a [distributed/shared cache](https://configcat.com/docs/advanced/caching/#shared-cache). (The `configChanged` event only fires when a new version of the config is fetched from the remote server, but not when an other actor puts a new version in the shared cache.)

Creating a snapshot is a cheap operation, so we can do it on demand.

adams85 (commented):
Hey @lukas-reining,

Thanks for implementing my proposals, they look all good!

However, I've found some further issues...

libs/providers/config-cat/src/lib/config-cat-provider.ts
```diff
@@ -1,82 +1,81 @@
...
-    return this._status;
+  public static create<TMode extends PollingMode>(
+    sdkKey: string,
+    pollingMode?: TMode,
+    options?: OptionsForPollingMode<TMode>,
+  ): ConfigCatProvider<TMode> {
+    return new ConfigCatProvider(sdkKey, pollingMode, options);
   }
 
   public async initialize(): Promise<void> {
```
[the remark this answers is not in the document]
lukas-reining: This change makes sense @adams85.
I would just not emit `this.events.emit(ProviderEvents.Ready);` here anymore, as the OF SDK already does it when the `init` method resolves successfully.

But the SDK also sets the provider status to `ERROR`, when the provider emits an `Error` Event. 
Does the Config Cat SDK has a way of communicating that an error condition is over? Some other SDKs emit a `READY` event in that case, but this does not seem to be the case for Config Cat (https://configcat.com/docs/sdk-reference/node/#hooks). If not, we might have to signal the end of the error condition e.g. after a successful evaluation has been done without an error. 
Do you have an opinion on this @adams85?

libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
lukas-reining: Oh okay, I did not consider this cheap but it makes sense.
Changed it @adams85.

lukas-reining:
Thanks for the review @adams85!
I like your proposals and implemented them.
Would be happy if you could have a look again!
I left one question regarding handling the client getting back to ready status after an error.

libs/providers/config-cat/src/lib/config-cat-provider.ts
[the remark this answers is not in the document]
adams85: > I would just not emit this.events.emit(ProviderEvents.Ready); here anymore, as the OF SDK already does it when the init method resolves successfully.

That sounds perfectly fine.

> But the SDK also sets the provider status to ERROR, when the provider emits an Error Event.
Does the Config Cat SDK has a way of communicating that an error condition is over?

No, the ConfigCat SDK provides no such mechanism at the moment. It doesn't even really make it possible to distinguish between transient and persistent errors. (However, there are plans to improve at least the latter situation by including error codes in the evaluation result objects. It's already [implemented in the .NET SDK](https://github.com/configcat/.net-sdk/blob/v9.2.0/src/ConfigCatClient/Evaluation/EvaluationDetails.cs#L91), we just haven't gotten around to adding it to the JS SDK so far.)

> If not, we might have to signal the end of the error condition e.g. after a successful evaluation has been done without an error.

That sounds like an acceptable workaround to me. At least, `EvaluationDetails.errorMessage` being null or not containing text like "Config JSON is not present" indicates that some config data has already been retrieved from the server (or from the distributed cache).

([A new hook named `ConfigFetched`](https://github.com/configcat/.net-sdk/blob/v9.2.0/src/ConfigCatClient/Hooks/IProvidesHooks.cs#L23) is also planned, which may make the heuristics a bit more accurate.)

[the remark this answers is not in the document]
lukas-reining: Okay, then I think we will go for that workaround until we can use the new hook.
Will do it tomorrow morning.
Anything else you see before we can merge @adams85 ?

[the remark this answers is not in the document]
adams85: No, it's fine by me now. Please go ahead with merging.

adams85 (approved):

libs/providers/config-cat-web/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Usage
+
+The ConfigCat provider uses the [ConfigCat JavaScript (SSR) SDK](https://configcat.com/docs/sdk-reference/js-ssr/).
+
+It can either be created by passing the ConfigCat SDK options to ```ConfigCatProvider.create``` or
+the ```ConfigCatProvider``` constructor.
+The available options can be found in the [ConfigCat JavaScript (SSR) SDK](https://configcat.com/docs/sdk-reference/js-ssr/).
+
+The ConfigCat Web Provider only supports the `AutoPolling` mode because it caches all evaluation data to support synchronous evaluation of feature flags.
```
toddbaert, line 31: Good thing to mention.

libs/providers/config-cat-web/project.json
```diff
@@ -0,0 +1,87 @@
...
+            "input": "./",
+            "output": "./"
+          },
+          {
+            "glob": "README.md",
+            "input": "./libs/providers/config-cat-web",
+            "output": "./"
+          }
+        ],
+        "updateBuildableProjectDepsInPackageJson": true
```
toddbaert, line 82: Did you manually add this or does it come from the template? I just added it for those recent ofrep packages, if you remember...

libs/providers/config-cat/src/lib/config-cat-provider.ts
```diff
@@ -1,84 +1,80 @@
...
       };
 
-      this._client = getClient(...originalParameters);
+      return getClient(sdkKey, pollingMode, options);
     });
   }
 
+  public async initialize(): Promise<void> {
+    const client = this._clientFactory(this);
+    await client.waitForReady();
```
toddbaert, line 74: This makes init easy for this provider... just a point in our discussion about "auto-ready-events".

libs/shared/config-cat-core/src/lib/context-transformer.ts
```diff
@@ -0,0 +1,48 @@
+import { EvaluationContext, EvaluationContextValue } from '@openfeature/core';
```
toddbaert, line 1: Good use of OpenFeature core.

libs/shared/config-cat-core/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@openfeature/config-cat-core",
+  "version": "0.1.0",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/core": "<=1.0.0"
```
toddbaert, line 9: :+1:

libs/providers/config-cat-web/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@openfeature/config-cat-web-provider",
+  "version": "0.6.1",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "peerDependencies": {
+    "@openfeature/web-sdk": "^1.0.0",
+    "@openfeature/config-cat-core": "0.1.0",
```
toddbaert, line 10: You may want to consider removing this, though it's up to you.

`updateBuildableProjectDepsInPackageJson` will automatically add `@openfeature/config-cat-core` as a dep, making things easy for users, and in this case, I don't think you _need_ a peer dep, since all of the usage of `@openfeature/config-cat-core` is internal to this library and none of it's surface area is exposed to users.

The only time you really need peer deps is if you need to ensure the library version and a version installed by consumers matches, but this isn't applicable here as far as I can see.

toddbaert (approved):
Approved! Nice work. Please just carefully consider this: https://github.com/open-feature/js-sdk-contrib/pull/918/files#r1681377307 (I could be missing something).

libs/providers/config-cat-web/README.md
```diff
@@ -0,0 +1,148 @@
...
+const provider = ConfigCatProvider.create('<sdk_key>');
+OpenFeature.setProvider(provider);
+```
+
+### Example using different polling options and a setupHook
+
+```javascript
+import { ConfigCatProvider } from '@openfeature/config-cat-web-provider';
+
+const provider = ConfigCatProvider.create('<sdk_key>', PollingMode.AutoPoll, {
```
adams85, line 47:
```suggestion
const provider = ConfigCatProvider.create('<sdk_key>', {
```

libs/providers/config-cat-web/README.md
```diff
@@ -0,0 +1,148 @@
...
+  setupHooks: (hooks) => hooks.on('clientReady', () => console.log('Client is ready!')),
+});
+
+OpenFeature.setProvider(provider);
+```
+
+## Evaluation Context
+
+ConfigCat only supports string values in its "evaluation
+context", [there known as User Object](https://configcat.com/docs/advanced/user-object/).
```
adams85, line 57: This is [not true any more](https://configcat.com/docs/sdk-reference/js-ssr/#user-object-attribute-types) (since JS SSR SDK v8.0), so this section should be revised.

libs/providers/config-cat/README.md
```diff
@@ -14,16 +14,15 @@ The OpenFeature SDK is required as peer dependency.
...
+$ npm install @openfeature/server-sdk configcat-node
 ```
 
 ## Usage
 
 The ConfigCat provider uses the [ConfigCat JavaScript (SSR) SDK](https://configcat.com/docs/sdk-reference/js-ssr/).
-This means that the provider can be used in both server (e.g. Node.js) and client (browser) applications.
 
 It can either be created by passing the ConfigCat SDK options to ```ConfigCatProvider.create``` or
 the ```ConfigCatProvider``` constructor.
```
adams85, line 28:
```suggestion
It can be created by passing the ConfigCat SDK options to ```ConfigCatProvider.create```.
```

adams85 (commented):
Spotted some outdated info in README.md files.

libs/providers/config-cat-web/package.json
lukas-reining: This is a good point and I will use it!🙂
Will change it.

libs/providers/config-cat-web/project.json
lukas-reining: Yes I remember, and yes it was already enabled from the template, I did not add it myself.

PR #933, opened 2024-06-06
chore: loosen some test assertions, fix e2e matcher

toddbaert:
Fixes a few issues in testing.

### DIFF
--- a/libs/providers/flagd/docker-compose.yaml
+++ b/libs/providers/flagd/docker-compose.yaml
@@ -1,25 +1,17 @@
 services:
   flagd:
-    build:
-      context: test-harness
-      dockerfile: flagd/Dockerfile
+    image: ghcr.io/open-feature/flagd-testbed:v0.5.4
     ports:
       - 8013:8013
   flagd-unstable:
-    build:
-      context: test-harness
-      dockerfile: flagd/Dockerfile.unstable
+    image: ghcr.io/open-feature/flagd-testbed-unstable:v0.5.4
     ports:
       - 8014:8013
   flagd-sync:
-    build:
-      context: test-harness
-      dockerfile: sync/Dockerfile
+    image: ghcr.io/open-feature/sync-testbed:v0.5.4
     ports:
       - 9090:9090
   flagd-sync-unstable:
-    build:
-      context: test-harness
-      dockerfile: sync/Dockerfile.unstable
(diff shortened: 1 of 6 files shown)


libs/providers/flagd/docker-compose.yaml
```diff
@@ -1,25 +1,17 @@
 services:
   flagd:
-    build:
-      context: test-harness
-      dockerfile: flagd/Dockerfile
+    image: ghcr.io/open-feature/flagd-testbed:v0.5.4
```
toddbaert, line 3: Use published images.

libs/providers/flagd/src/e2e/step-definitions/flagd-json-evaluator.spec.ts
```diff
@@ -24,7 +24,8 @@ const evaluateStringFlagWithContext: StepsDefinitionCallbackFunction = ({ given,
     flagKey = key;
     defaultValue = defaultVal;
   });
-  and(/^a context containing a key "(.*)", with value "(.*)"$/, (key: string, value: string) => {
+  // the below has to match quotes strings ("str") and numbers (3) to test an error input
+  and(/^a context containing a key "(.*)", with value "?([^"]*)"?$/, (key: string, value: string) => {
```
toddbaert, line 28: This matcher probably should have never worked when we first added `3` as an error state, it looks like it was a bug eventually fixed in `jest-gherkin`. This regex matches `"abc"` and `3`, which fixes the e2e test.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.spec.ts
```diff
@@ -139,13 +139,13 @@ describe('LaunchDarklyClientProvider', () => {
...
       const res = ofClient.getBooleanDetails(testFlagKey, false);
-      expect(res).toEqual({
+      expect(res).toEqual(expect.objectContaining({
         flagKey: testFlagKey,
         flagMetadata: {},
         value: false,
         reason: 'ERROR',
         errorCode: 'TYPE_MISMATCH',
-      });
+      }));
```
toddbaert, line 148: I've loosened a few brittle assertions with this PR, so that if new files are added/defaulted they don't break.

beeme1mr (approved):

PR #947, opened 2024-06-17
feat(gofeatureflag): Clear cache if configuration changes + provider refactoring

thomaspoignant:
## This PR
- Clear cache if GO Feature Flag configuration is changing.
- Refactor the code base to have a better separation of components.


libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
```diff
@@ -218,99 +171,51 @@ export class GoFeatureFlagProvider implements Provider {
...
-        errorMessage: `Provider in a status that does not allow to serve flag: ${this.status}`,
-      };
-    }
-
-    const cacheKey = `${flagKey}-${hash(user)}`;
-    // check if flag is available in the cache
-    if (this.cache !== undefined) {
-      const cacheValue = this.cache.get(cacheKey);
+    if (this._cacheController?.get(flagKey, evaluationContext) !== undefined) {
+      const cacheValue = this._cacheController.get(flagKey, evaluationContext);
```
lukas-reining, line 178: Nit, but maybe you could save the result of `_cacheController?.get` in line 177 already and only check the value, then you can remove the following undefined check on the resolution value.

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
```diff
@@ -1,118 +1,76 @@
...
    * onClose is called everytime OpenFeature.Close() function is called.
-   * It will terminate gracefully the provider and ensure that all the data are send to the relay-proxy.
+   * It will gracefully terminate the provider and ensure that all the data are sent to the relay-proxy.
    */
   async onClose() {
-    await this.dataCollectorHook?.close();
+    this._cacheController?.clear();
+    await this._dataCollectorHook?.close();
+    this.stopPolling();
+    this.events.emit(ServerProviderEvents.Error, { message: 'Provider closed' });
```
lukas-reining, line 73: Nor sure if this should emit an error.
It feels like this is totally expected behavior that would end up in error log or so possibly.
But I am not 100% sure on it.

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
```diff
@@ -1,118 +1,76 @@
...
   /**
    * onClose is called everytime OpenFeature.Close() function is called.
-   * It will terminate gracefully the provider and ensure that all the data are send to the relay-proxy.
+   * It will gracefully terminate the provider and ensure that all the data are sent to the relay-proxy.
    */
   async onClose() {
-    await this.dataCollectorHook?.close();
+    this._cacheController?.clear();
+    await this._dataCollectorHook?.close();
+    this.stopPolling();
```
lukas-reining, line 72: This is also a bit and I guess not important but we should first stop polling and clear the cache then.
```suggestion
    this.stopPolling();
    this._cacheController?.clear();
    await this._dataCollectorHook?.close();
```

lukas-reining (approved):
Change looks great! Just left two small comments.

libs/providers/go-feature-flag/src/lib/go-feature-flag-provider.ts
thomaspoignant: You are right, I'll remove it.

PR #954, opened 2024-06-19
feat: Change fractional custom op from percentage-based to relative weighting. #946

### LINKED ISSUE #946: [FEATURE] Change fractional custom op from percentage-based to relative weighting.
### Requirements

Currently, the fractional operator in the in-process provider is percentage-based; meaning you must specify the relative distribution of the variants in terms of a percentage. Consistent with flagd, we should change this instead to be a relative weight, meaning that the integers do not need to sum to 100. For example, this fractional:

```json
"fractional": [
          [
            "red",
            25
          ],
          [
            "blue",
            25
          ],
          [
            "green",
            25
          ],
          [
            "grey",
            25
          ]
        ]
```

Would remain valid, but could also be expressed as:

```json
"fractional": [
          [
            "red",
            1
          ],
          [
            "blue",
            1
          ],
          [
            "green",
            1
          ],
          [
            "grey",
            1
          ]
        ]
```
See [issue](https://github.com/open-feature/flagd/issues/1282) and [PR](https://github.com/open-feature/flagd/pull/1313) in Go/flagd.

aepfli:
### Related Issues

Fixes #946

### DIFF
--- a/libs/shared/flagd-core/src/lib/targeting/fractional.ts
+++ b/libs/shared/flagd-core/src/lib/targeting/fractional.ts
@@ -86,7 +86,7 @@ function toBucketingList(from: unknown[]): {
       throw new Error('Invalid bucket entries');
     }
 
-    if (entry.length == 0) {
+    if (entry.length == 0 || entry.length > 2) {
       throw new Error('Invalid bucketing entry. Requires at least a variant');
     }
 
--- a/libs/shared/flagd-core/src/lib/targeting/fractional.ts
+++ b/libs/shared/flagd-core/src/lib/targeting/fractional.ts
@@ -52,10 +52,10 @@ export function fractionalFactory(logger: Logger) {
     const bucket = (Math.abs(hash) / 2147483648) * 100;
 
     let sum = 0;
-    for (let i = 0; i < bucketingList.length; i++) {
-      const bucketEntry = bucketingList[i];
+    for (let i = 0; i < bucketingList.fractions.length; i++) {
+      const bucketEntry = bucketingList.fractions[i];
 
-      sum += bucketEntry.fraction;
+      sum += relativeWeight(bucketingList.totalWeight, bucketEntry.fraction);
 
       if (sum >= bucket) {
         return bucketEntry.variant;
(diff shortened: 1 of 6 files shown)


aepfli:
it seems like i updated the submodules by accident

aepfli:
relates also to:  https://github.com/open-feature/flagd/issues/1282 - for tracking purpose in the main ticket

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -66,36 +66,45 @@ export function fractionalFactory(logger: Logger) {
...
     if (typeof entry[0] !== 'string') {
       throw new Error('Bucketing require variant to be present in string format');
     }
 
-    if (typeof entry[1] !== 'number') {
-      throw new Error('Bucketing require bucketing percentage to be present');
+    let weight = 1;
+    if (entry.length >= 2) {
+      if (typeof entry[1] !== 'number') {
+        throw new Error('Bucketing require bucketing percentage to be present');
```
toddbaert, line 100:
```suggestion
        throw new Error('Bucketing weight must be numeric');
```

toddbaert (approved):

toddbaert:
Wow thanks so much @aepfli ! Only left 1 minor comment.

libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -66,36 +66,45 @@ export function fractionalFactory(logger: Logger) {
...
+  let totalWeight = 0;
   for (let i = 0; i < from.length; i++) {
     const entry = from[i];
     if (!Array.isArray(entry)) {
       throw new Error('Invalid bucket entries');
     }
 
-    if (entry.length != 2) {
-      throw new Error('Invalid bucketing entry. Require two values - variant and percentage');
+    if (entry.length == 0) {
```
toddbaert, line 89:
```suggestion
    if (entry.length == 0) {
```
I think this check could be improved a bit - it should be 1-2, 0 or more than 2 are all invalid. This is also validated by schema, but people can bypass schema validation.

toddbaert:
@aepfli [testbed v0.5.5](https://github.com/open-feature/flagd-testbed/releases/tag/v0.5.5) is available now. If you update our test containers and the submodule to that version, the tests new e2e should work (I tested them locally on this branch).

libs/shared/flagd-core/src/lib/targeting/fractional.ts
aepfli: thank you, improved and added a test case

toddbaert (approved):

PR #961, opened 2024-06-26
fix(gofeatureflag): remove NodeJS.timeout because it fails eslint

thomaspoignant:
## This PR

The CI in PR #947 had no issue in CI, but it failed during the [release-please job](https://github.com/open-feature/js-sdk-contrib/actions/runs/9675435875/job/26692945884).

I am not sure why those checks are different, I guess we have something different in the CI.
This PR should fix the release please process for the go-feature-flag-provider.

@toddbaert @beeme1mr I am not sure if we have to update the release please file to be sure the right version is released?


lukas-reining (approved):
I hate it, happens every time😅

PR #977, opened 2024-07-02
chore: fix e2e test

toddbaert:
This should fix the hung e2e testing.


libs/providers/flagd/src/e2e/step-definitions/flagd-json-evaluator.spec.ts
```diff
@@ -75,6 +75,27 @@ defineFeature(feature, (test) => {
 
   test('Fractional operator', evaluateStringFlagWithFractional);
 
+  test('Fractional operator shorthand', ({ given, when, and, then }) => {
```
toddbaert, line 78: Unfortunately the way the bindings work in JS isn't as flexible as Java, so we need to implement this, even though it's very close to existing tests :/

beeme1mr (approved):

aepfli (approved):

PR #982, opened 2024-07-03
chore: removing build dependencies and using testcontainers for container spin up

aepfli:
## This PR

Removing the need for local dependencies of buf and the need to run docker compose to run e2e tests.

The Docker container will automatically start, and buf files will be loaded via npm from buf.build as generated SDKs.

I tried to do the same thing for flagd, but the generated buf files via https://github.com/stephenh/ts-proto are not delivered as SDKs from buf.build. Hence, I can't remove the dependency on buf. If we finalize this approach, I will add the test-container execution.

to Verify run ` nx e2e providers-flagd-web` locally, it should actually run without any docker spin ups


libs/providers/flagd-web/project.json
```diff
@@ -76,7 +73,7 @@
       "executor": "nx:run-commands",
       "options": {
         "commands": [
-          "npx jest"
+          "npx jest --forceExit"
```
aepfli, line 76: some how jest did not exist with testcontainers

toddbaert:
This is cool!

libs/providers/flagd-web/project.json
toddbaert: `--forceExit` is bad, because it sometimes hides real bugs (it has before caught bugs with gRPC connections not dying, which can cause serious problems with reconnection in the provider). If we add `--forceExit`, we will not see such bugs.

toddbaert:
Will this clash with the CI's services? Would we need to disable those?

aepfli:
> Will this clash with the CI's services? Would we need to disable those?


it will not clash with ci services, and if we use  testcontainers everywhere we should not need to have this ci services anymore - as our test will handle the containers, and more importantly provide better port mapping out of the box - see the java reference https://github.com/open-feature/java-sdk-contrib/pull/860 - where we can actually skip the services

libs/providers/flagd-web/project.json
aepfli: it seems like test containers or something did not end properly, i am not sure how to best debug this, we can investigate together if you like. the tests have been successful, but somehow a handle was still active, not sure which one ;)

libs/providers/flagd-web/src/e2e/setup.ts
```diff
@@ -11,11 +12,11 @@ OpenFeature.setProvider(
     host: 'localhost',
     port: 8013,
     tls: false,
-    maxRetries: -1,
```
toddbaert, line 14: This is the cause of your `--force-exit` / open handler issues. Without this param, the provider tries to reconnect to flagd (this socket is the open handle). Add it to where you now create this provider and things work again :)

libs/providers/flagd-web/src/e2e/step-definitions/evaluation.spec.ts
```diff
@@ -17,19 +27,43 @@ const feature = loadFeature('features/evaluation.feature');
...
+      assert(
+        OpenFeature.providerMetadata.name === FLAGD_WEB_NAME,
+        new Error(`Expected ${FLAGD_WEB_NAME} provider to be configured, instead got: ${OpenFeature.providerMetadata.name}`),
+      );
+      console.log('flagd web provider configured!');
+    });
+  }, 60000);
 
   afterAll(async () => {
+    await startedContainer.stop();
```
toddbaert, line 66: I think you should do this after shutting down OpenFeature.

libs/providers/flagd-web/src/e2e/setup.ts
aepfli: oh thank you, this makes so much sense

aepfli:
@toddbaert i kindly ask you to review this again.

i changed the way we build buf files, now it is just a npx command without any fancy directory changes, and file operations (just using the cli)

i removed all the setups for the e2e tests and added dedicated jest wrapper tests, which will handle the spin up and tear down of the container (those tests are a good starting point for https://github.com/open-feature/js-sdk-contrib/issues/986) - due to that all previous tests are now exporting a function which is called via the wrapper tests.

it looks big, but it is not really complicated, please review if you do have time

libs/providers/flagd/project.json
```diff
@@ -21,9 +21,7 @@
       "options": {
         "commands": [
           "git submodule update --init schemas",
-          "rm -f -r ./src/proto",
-          "cd schemas && buf generate buf.build/open-feature/flagd --template protobuf/buf.gen.ts.yaml",
-          "mv -v ./proto ./src"
+          "npx buf generate buf.build/open-feature/flagd --template schemas/protobuf/buf.gen.ts.yaml --output ./src/lib"
```
lukas-reining, line 24: This is a nice change!

libs/providers/flagd/src/e2e/tests/rpc-provider.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+describe('rpc', () => {
+  let containers: StartedTestContainer[] = [];
+  beforeAll(async () => {
+    containers = await setup();
+  }, 60000);
+  afterAll(async () => {
+    await OpenFeature.close();
+    for (const container of containers) {
+      container.stop();
```
lukas-reining, line 74: Is there a reason you did not await the close?
Could be better to stop them all but being sure to run all, even if there is an error.
Maybe something like:

lukas-reining:
```suggestion
      try {
        await container.stop();
      } catch {
        console.warn(`Failed to stop container ${container.getName()}`);
      }
```

lukas-reining (approved):
Great change and great to have Testcontainers in!

libs/providers/flagd/src/e2e/tests/rpc-provider.spec.ts
aepfli: no special reason - just got lost during my different approaches ;)

lukas-reining (approved):

libs/providers/flagd-web/project.json
```diff
@@ -76,7 +74,7 @@
       "executor": "nx:run-commands",
       "options": {
         "commands": [
-          "npx jest"
+          "npx jest --detectOpenHandles --runInBand"
```
toddbaert, line 77: I think we can remove at least `--detectOpenHandles`? and possibly `--runInBand`? I guess this was from troubleshooting this hanging handles?

libs/providers/flagd-web/src/e2e/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const FLAGD_NAME = 'flagd-web';
```
toddbaert, line 1: TBH we should maybe export this from the provider itself (the provider has a name).

libs/providers/flagd/src/e2e/step-definitions/evaluation.ts
```diff
@@ -0,0 +1,364 @@
+import {
```
toddbaert, line 1: I'm curious why you removed "spec" from these files.

toddbaert:
@aepfli this looks excellent to me. TBH of all our gherkin suites, I was the least happy with this one, this seems like a huge improvement. Thanks a lot. I left a few nits, and I want to check one or two things locally, but I think I'll approve soon.

libs/providers/flagd/src/e2e/step-definitions/evaluation.ts
aepfli: they are containing the test logic, but are not the actual test files, i am exporting the logic now, so i can reuse it in each provider test seperately

libs/providers/flagd/src/e2e/tests/in-process-provider.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+import { flagdRecconnectUnstable } from '../step-definitions/flagd-reconnect.unstable';
+
+// register the flagd provider before the tests.
+async function setup() {
+  const containers: StartedTestContainer[] = [];
+
+  console.log('Setting flagd provider...');
+
+  const stable = await new GenericContainer(`ghcr.io/open-feature/sync-testbed:${IMAGE_VERSION}`)
+    .withExposedPorts(9090)
```
toddbaert, line 24:
```suggestion
    .withExposedPorts(8015)
```

libs/providers/flagd/src/e2e/tests/in-process-provider.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    .start();
+  containers.push(stable);
+  OpenFeature.setProvider(
+    E2E_CLIENT_NAME,
+    new FlagdProvider({ resolverType: 'in-process', host: 'localhost', port: stable.getFirstMappedPort() }),
+  );
+
+  const unstable = await new GenericContainer(`ghcr.io/open-feature/sync-testbed-unstable:${IMAGE_VERSION}`)
+    .withExposedPorts(9090)
```
toddbaert, line 33:
```suggestion
    .withExposedPorts(9090)
```

Same as https://github.com/open-feature/js-sdk-contrib/pull/982/files#r1670560390, let's make this 8015 since that's the default and we have no clashing to worry about now.

toddbaert (approved):
Amazing. This is much nicer. Can't thank you enough! Event the console feedback is much more understandable now.

libs/providers/flagd/src/e2e/step-definitions/evaluation.ts
toddbaert: Got it.

libs/providers/flagd-web/package.json
```diff
@@ -7,5 +7,12 @@
   },
   "peerDependencies": {
     "@openfeature/web-sdk": "^1.0.0"
+  },
+  "dependencies": {
+    "@buf/open-feature_flagd.bufbuild_es": "^1.10.0-20240215170432-1e611e2999cc.1",
```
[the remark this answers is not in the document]
aepfli: that is a left over i forgot to delete - sorry

toddbaert:
I've marked this as a chore, since it's only a "feature" for developers, not for the users of the flagd/flagd-web providers.

libs/providers/flagd-web/src/e2e/constants.ts
aepfli: i will leave this for now, and tackle it at a later stage

PR #990, opened 2024-07-08
feat!: resolver signatures changes for easier overrides, cache exposure

toddbaert:
This class is likely to be extended, so I've made some changes to improve the QoL around that.

- resolvers now accept all params
- public access to a copy of the flag cache


lukas-reining (approved):
Looks good to me!

PR #1003, opened 2024-07-10
feat!: use native headers, optional query params

toddbaert:
This PR: 

- changes the types of the `Headers` used in the OFREP providers to the native fetch API (which we use anyway); this reduces some boilerplate
- adds the ability to send optional custom query params to OFREP (web and server)
- minor refactors mostly to reduce small duplication (new feature, less code :sunglasses: )

:warning:

EDIT:

This PR also uses NX settings to automatically make ofrep-core a generated dependency of ofrep-web and ofrep (a normal dependency instead of a manual peer). I don't think we want ofrep-core as a peer here, since no API surface area of ofrep-core is seen by consumers, we can keep our own copy without worry of conflict - it's and implementation detail, I believe.

This is what the generated deps look like: 

image


package.json
```diff
@@ -21,7 +21,6 @@
     "@flipt-io/flipt-client-browser": "^0.0.17",
     "@growthbook/growthbook": "^1.0.0",
     "@grpc/grpc-js": "^1.9.13",
-    "@openfeature/ofrep-core": "^0.1.4",
```
toddbaert, line 24: We shouldn't have any deps that are local to this repo (dev or otherwise). We always want to resolve the local ones.

lukas-reining (approved):
Change looks good and makes sense!

toddbaert:
> Change looks good and makes sense!

Thanks Lukas... please see the edit to my description (the bundling change) I made it after you reviewed, and I want to make sure it makes sense to everyone.

libs/providers/ofrep-web/project.json
```diff
@@ -68,7 +68,8 @@
             "input": "./libs/providers/ofrep-web",
             "output": "./"
           }
-        ]
+        ],
+        "updateBuildableProjectDepsInPackageJson": true
```
toddbaert, line 72: This now automatically generates a dependency entry for us for core (so I've removed the corresponding peer).

libs/shared/ofrep-core/src/lib/provider/ofrep-provider-options.ts
```diff
@@ -1,8 +1,4 @@
-import { FetchAPI, RequestOptions } from '../api';
-
-export type HttpHeaderList = [name: string, value: string][];
-export type HttpHeaderMap = Record<string, string>;
-export type HttpHeaders = HttpHeaderList | HttpHeaderMap;
```
toddbaert, line 5: swapped for native fetch headers, and their simple equivalent structure.

libs/providers/ofrep-web/project.json
lukas-reining: I have never used this option, but the doc reads like this making sense!
Have you tried it?

libs/shared/ofrep-core/src/lib/provider/ofrep-provider-options.ts
```diff
@@ -21,26 +17,27 @@ export type OFREPProviderBaseOptions = {
    * Optional Headers supplier function.
    * @returns HttpHeaders
    */
-  headersFactory?: () => HttpHeaders;
+  headersFactory?: () => Promise<[string, string][]>;
   /**
    * Optional static headers.
    */
-  headers?: HttpHeaders;
+  headers?: [string, string][];
```
toddbaert, line 24: This type can be used to init the native fetch headers, so it's a familiar type to people and allows us to leverage existing utilities.

lukas-reining: Yes definitely, `[string, string][]` was a subtype of it, but it makes sense to have the standard type here.

Kavindu-Dodan (approved):

libs/providers/ofrep-web/project.json
toddbaert: We actually currently use for for flagd/flagd-web to automatically pull in flagd-core, so I know it works!

lukas-reining: Sounds great!

aepfli (approved):

PR #1008, opened 2024-07-12
chore: various gherkin improvements for e2e tests

### LINKED ISSUE #986: [flagd] change e2e test to use autobinding of gherkin test
We currently have some e2e tests that use our test-harness repository to run standardized Gherkin tests. Sadly, this repository does not use autobinding and proper step definitions (see [autobinding document](https://github.com/bencompton/jest-cucumber/blob/main/docs/AutomaticStepBinding.md)). Hence, we need to implement each test case additionally.

The goal of this task is to use auto binding so our test suite will always be updated with the latest version and use the latest version for building.

- flagd e2e https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/flagd/src/e2e
- flagd-web e2e https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/flagd-web/src/e2e

relates to: #515

aepfli:
## This PR
- using gherkin auto binding as described in #986 
- unify the location for the test harness, spec and the auto binding files within flagd-core instead of flagd and flagd-web
- removing file operations to move the feature files before execution into a seperate folder (no need for this anymore)
- removed step definitions and constants of e2e tests from the built package

### Related Issues

Fixes #986 
### Follow-up Tasks

currently the path for the gherkin feature files is 'hardcoded' and not autmatically resolved relatively to the test (maybe somebody has a good idea for this)


aepfli:
okay, i failed here hard, i did not check the api, and the webclient and the server client do not have an overlapping api - hence that part of this improvements are not worth it. eg. the normalization witthin flagd-core. I could generate an wrapper around it, to be passed on to the stepdefinitions, but imho i do not think that this is worth it.

libs/providers/flagd/src/e2e/step-definitions/reconnect.ts
```diff
@@ -0,0 +1,59 @@
...
+      readyRunCount++;
+      done();
+    });
+  });
+
+  when('a PROVIDER_READY handler and a PROVIDER_ERROR handler are added', () => {
+    client.addHandler(ProviderEvents.Error, () => {
+      errorRunCount++;
+    });
+
```
aepfli, line 34:
```suggestion
```

aepfli:
@beeme1mr @toddbaert I think this is ready. I just have had no time to investigate why the flags changed test for flagd-web is failing. If I find the time I will investigate. If you on the other hand read this, before my next message and have more insights (maybe not implement fully) just let me know ;)

aepfli:
> @beeme1mr @toddbaert I think this is ready. I just have had no time to investigate why the flags changed test for flagd-web is failing. If I find the time I will investigate. If you on the other hand read this, before my next message and have more insights (maybe not implement fully) just let me know ;)

Okay it seems like we are not providing a list of changed flags - as we are reloading all of them - makes sense. The question is, if it is worth checking if the flag has changed from previous calls. Or if getting the event alone is test enough

toddbaert:
Will check this out tomorrow!

libs/providers/flagd-web/project.json
```diff
@@ -58,23 +58,11 @@
...
-        "cwd": "libs/providers/flagd-web",
-        "parallel": false
-      }
-    },
     "e2e": {
       "executor": "nx:run-commands",
       "options": {
         "commands": [
-          "npx jest"
+          "npx jest --detectOpenHandles --runInBand"
```
beeme1mr, line 65: Why was this change made?

libs/providers/flagd-web/src/e2e/step-definitions/flag.ts
```diff
@@ -0,0 +1,352 @@
...
+    ran = new Promise<boolean>((resolve) => {
+      client.addHandler(ProviderEvents.ConfigurationChanged, async (details) => {
+        // file writes are not atomic, so we get a few events in quick succession from the testbed
+        // some will not contain changes, this tolerates that; at least 1 should have our change
+
+        // All Flags are changed we do not provide a list of changed flags, that is the nature of the web sdk
+        //if (details?.flagsChanged?.length) {
+        //  flagsChanged = details?.flagsChanged;
+
+        resolve(true);
```
beeme1mr, line 280: This isn't testing anything.

libs/providers/flagd-web/src/e2e/step-definitions/flag.ts
```diff
@@ -0,0 +1,352 @@
...
+    await new Promise((resolve) => setTimeout(resolve, 3000));
+  });
+
+  then('the PROVIDER_CONFIGURATION_CHANGED handler must run', async () => {
+    expect(await ran).toBeTruthy();
+  });
+
+  and(/^the event details must indicate "(.*)" was altered$/, (flagName) => {
+    // All Flags are changed we do not provide a list of changed flags, that is the nature of the web sdk
+    //expect(flagsChanged).toContain(flagName);
```
beeme1mr, line 297: Can this be reenabled?

libs/shared/flagd-core/src/e2e/index.ts
```diff
@@ -0,0 +1,8 @@
+export const E2E_CLIENT_NAME = 'e2e';
+
+export const IMAGE_VERSION = 'v0.5.13';
+
+export function getGherkinTestPath(file: string, modulePath = 'test-harness/gherkin/'): string {
+  // TODO: find a way to resolve this in a generic manner - currently this works, because of the file structure
```
beeme1mr, line 6: Is this a future TODO or something that should be resolved before merging?

aepfli: i do not think we need to handle this currently, if would be good, but this works as is, due to the layout. I am not sure if it is really worth to investigate deeper

libs/providers/flagd-web/src/e2e/step-definitions/flag.ts
aepfli: No, we don't support this feature currently. There is no list of changed flags - hence this and the code above are commented out. This would be an enhancement to the provider, and is not part of this gherkin migration

libs/providers/flagd-web/project.json
aepfli: that might be a left over - i will check, too much time passed since i initially worked on this

libs/providers/flagd/src/e2e/tests/rpc-reconnect.spec.ts
```diff
@@ -59,11 +42,13 @@ async function setup() {
...
     ),
   );
+
   console.log('flagd provider configured!');
   return containers;
 }
 
-describe('rpc', () => {
+jest.setTimeout(30000);
+describe('in process', () => {
```
toddbaert, line 51:
```suggestion
describe('rpc', () => {
```

libs/providers/flagd/src/e2e/tests/rpc.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        OpenFeature.getProviderMetadata(E2E_CLIENT_NAME).name
+      }`,
+    ),
+  );
+
+  console.log('flagd provider configured!');
+  return containers;
+}
+
+describe('in process', () => {
```
toddbaert, line 39:
```suggestion
describe('rpc', () => {
```

libs/providers/flagd-web/project.json
toddbaert, line 65:
```suggestion
          "npx jest
```

Yep, works fine without this.

libs/providers/flagd-web/src/e2e/step-definitions/flag.ts
toddbaert: @aepfli 

@beeme1mr is right about this. `ran` will always be truthy b/c it's a promise.

You don't need to think in terms of promises here, you can do it simply with a var and some callbacks:

```ts
// define a var
let changeHandlerRan = false;

// add a handler
when('a PROVIDER_CONFIGURATION_CHANGED handler is added', () => {
  client.addHandler(ProviderEvents.ConfigurationChanged, () => {
    changeHandlerRan = true;
  });
});

// wait a bit (this is already implemented:
and(/^a flag with key "(.*)" is modified$/, async () => {
  // this happens every 1s in the associated container, so wait 3s
  await new Promise((resolve) => setTimeout(resolve, 3000));
});

// assert the handler ran (and mutated the var)
then('the PROVIDER_CONFIGURATION_CHANGED handler must run', async () => {
  expect(changeHandlerRan).toBeTruthy();
});
```

Re-writing these tests as above, the tests actually fail if you comment out the mutation of the `changeHandlerRan` var in the handler, proving the test actually works.

I think you'd have to do this for all the `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_READY` event tests.

toddbaert (changes requested):
Amazing!

A few minor nits I left where we are logging the wrong thing.

My only blocker is [this](https://github.com/open-feature/js-sdk-contrib/pull/1008#discussion_r1829696583).

libs/shared/flagd-core/src/e2e/index.ts
beeme1mr: Could you please review the comment and resolve this thread?

libs/providers/flagd-web/src/e2e/step-definitions/flag.ts
aepfli: changed it, and also added more proper documentation for the comments. i only need to create the issue to actually use the changed flags.

libs/providers/flagd-web/src/e2e/step-definitions/flag.ts
beeme1mr, line 1: Could you please address these linting issues?

beeme1mr (approved):
Looks good but please address the linting issues before merging. Nice job!

toddbaert (approved):

toddbaert:
Amazing!

PR #1010, opened 2024-07-18
feat(flagd-core): add bulk evaluation method

beeme1mr:
## This PR

- adds a bulk evaluation API

### Notes

The bulk evaluation API makes it easier to use flagd core to build an OFREP-compatible API.


libs/shared/flagd-core/src/lib/flagd-core.ts
```diff
@@ -135,7 +162,7 @@ export class FlagdCore implements Storage {
    * @throws {GeneralError} - If the variant specified in the flag is not found.
    */
   resolve<T extends FlagValue>(
-    type: FlagValueType,
+    type: FlagValueType | 'any',
```
aepfli, line 165: Would undefined also work here instead of any? - if we need `'any'` i would suggest to place it in a constant or something like this, to make it more accessible and understandable.

aepfli (approved):
lgtm, i am just not sure if `'any'` is the way to go

libs/shared/flagd-core/src/lib/flagd-core.ts
beeme1mr: This isn't an `any` type in the TypeScript sense. It's a string representation of the flag type. During bulk evaluation, we don't care what type it is because it will be enforced later by the SDK.

I'm happy to use a different name but I struggled to think of anything better.

aepfli: i am fine with `any` but it would be cool to use it as a type rather than a string like 
```
type AnyFlagValueType = 'any';

and then we use the this type, rather than the string. the string has something magical to it. where as a defined type is easier understood, and can be better documented

Comment (toddbaert) [reply libs/shared/flagd-core/src/lib/flagd-core.ts:165]: Can we instead make an overload that omits this param?

Comment (beeme1mr) [reply libs/shared/flagd-core/src/lib/flagd-core.ts]: I've refactored this part to remove the `any` type. Thanks for your feedback.

File: libs/shared/flagd-core/src/lib/flagd-core.ts [hunk 2]
```diff
@@ -121,10 +122,36 @@ export class FlagdCore implements Storage {
...
+      }
+    }
+    return values;
+  }
+
   /**
    * Resolves the value of a flag based on the specified type type.
    * @template T - The type of the flag value.
-   * @param {FlagValueType} type - The type of the flag value.
+   * @param {FlagValueType} type - The type of the flag value. Use 'any' to skip type validation.
```
toddbaert, line 154:
```suggestion
   * @param {FlagValueType} type - The type of the flag value.
```

aepfli (approved):

PR #1026, opened 2024-07-25
fix(flagsmith): boolean defaults and flag state evaluation fixes

### LINKED ISSUE #1025: Flagsmith doesn't use boolean defaults and uses enablement status instead of flag value
# Introduction
Hey team, we've began using Flagsmith in scenarios other than the most common or basic, specifically creating a boolean flag that can be overriden in certain cenarios for part of the clients based off a segment.

Basically the flag is usually a truthy value, but a segment override that will be edited over time may decrease the amount of clients that receive `true` and will now get a `false` value. Kinda like a soft phase-out of a feature.

While building that we noticed that the provider didn't take into account falsey default values in boolean flags, using a fixed truthy value in the code, and the evaluation for these boolean values was based off their existance/enablement, and not their actual value, preventing us from using these overrides that change the value, not the enabled status.

I took the liberty to aside from reporting this, also preparing a reproducible environment and a PR fix, but please feel free to discuss and opinate on the issue or solution.

# Links
Reproducible Example: https://github.com/uriell/openfeature-flagsmith-boolean-issues
Pull Request: https://github.com/open-feature/js-sdk-contrib/pull/1025

uriell:
## This PR
Hey team, I wrote a more detailed issue (#1025) but here is a summary:
>The Flagsmith provider doesn't take into account falsey default values in boolean flags, and it is currently using a fixed truthy value in the code, and the evaluation for these boolean values was based off their existance/enablement, and not their actual value, preventing us from using overrides that change the value, not the enabled status.

This PR includes:
- The fix in `resolveBooleanEvaluation` to allow boolean flags to use a specified default value;
- The fix in `evaluate` to ensure boolean flags don't have a special scenario and also consume it's value instead of checking for the existance of the flag;
- Test scenarios that now validate the state value of boolean flags instead of their enablement;
- Additional test scenarios to ensure default values specified outside of constructor default values are also applied and used;

### Related Issues
Fixes #1025 

### Notes
- I took the liberty to make an adjustment to the readme since the suggested contribution command in the flagsmith provider's README wasn't it's actual project name;
- I noticed the provider was written by one of Flagsmith's Founders (@kyle-ssg) and there was a mention in #836 regarding the behavior for boolean flags using the enablement status so there may be additional context they may have regarding Flagsmith's expected behavior, but upon using their platform it seemed counterintuitive in this scenario.

### Follow-up Tasks
- I haven't identified any follow up tasks, but feel free to flag any you may see.

### How to test
- The test scenarios were adjusted to cover the changes, but you may also experiment with the reproducible example, that has the local changes alongside it: https://github.com/uriell/openfeature-flagsmith-boolean-issues

### DIFF
--- a/libs/providers/flagsmith-client/README.md
+++ b/libs/providers/flagsmith-client/README.md
@@ -88,11 +88,11 @@ following [OpenFeature events](https://openfeature.dev/specification/types#provi
 
 ## Building
 
-Run `nx package providers-flagsmith` to build the library.
+Run `nx package providers-flagsmith-client` to build the library.
 
 ## Running unit tests
 
-Run `nx test providers-flagsmith` to execute the unit tests via [Jest](https://jestjs.io).
+Run `nx test providers-flagsmith-client` to execute the unit tests via [Jest](https://jestjs.io).
 
 ## Examples
 
--- a/libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.spec.ts
+++ b/libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.spec.ts
@@ -270,10 +270,16 @@ describe('FlagsmithProvider', () => {
       expect(details.value).toEqual(0);
       expect(details.reason).toEqual('ERROR');
     });
-    it('should use defaults for flags that do not exist', async () => {
+    it('should use defaults for number flags that do not exist', async () => {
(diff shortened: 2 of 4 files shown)


uriell:
I will temporarily close this Pull Request as I think a more thought out solution needs to be discussed in #1025 first.

PR #1048, opened 2024-08-29
fix(flagd): improve targeting key check in fractional operator

beeme1mr:
## This PR

- improve targeting key check in fractional operator

### Notes

The old logic was always truthy. This corrects the behavior and brings it in line with Java.

https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/targeting/Fractional.java#L43

### How to test

Run the jest test suite. The test added this PR failed before the fix.

### DIFF
--- a/libs/shared/flagd-core/src/lib/targeting/fractional.ts
+++ b/libs/shared/flagd-core/src/lib/targeting/fractional.ts
@@ -29,12 +29,12 @@ export function fractionalFactory(logger: Logger) {
       bucketBy = args[0];
       buckets = args.slice(1, args.length);
     } else {
-      bucketBy = `${flagdProperties[flagKeyPropertyKey]}${context[targetingPropertyKey]}`;
-      if (!bucketBy) {
+      const targetingKey = context[targetingPropertyKey];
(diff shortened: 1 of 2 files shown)


libs/shared/flagd-core/src/lib/targeting/fractional.ts
```diff
@@ -29,12 +29,12 @@ export function fractionalFactory(logger: Logger) {
       bucketBy = args[0];
       buckets = args.slice(1, args.length);
     } else {
-      bucketBy = `${flagdProperties[flagKeyPropertyKey]}${context[targetingPropertyKey]}`;
-      if (!bucketBy) {
+      const targetingKey = context[targetingPropertyKey];
```
beeme1mr, line 32: This was the bug... the flag key is set automatically by flagd.

toddbaert: Nice!

toddbaert (approved):
Nice catch.

PR #1052, opened 2024-09-03
fix(flipt): return the correct default value when provided in Flipt

erka:
## This PR

Resolve the issue with the returning correct default value

### Related Issues

Fixes flipt-io/flipt/issues/3418

### DIFF
--- a/libs/providers/flipt/src/lib/flipt-provider.ts
+++ b/libs/providers/flipt/src/lib/flipt-provider.ts
@@ -169,7 +169,7 @@ export class FliptProvider implements Provider {
         };
       }
 
-      if (!variantEvaluation.match) {
+      if (!variantEvaluation.match && variantEvaluation.reason != 'DEFAULT_EVALUATION_REASON') {
         return {
           value: defaultValue,
           reason: StandardResolutionReasons.DEFAULT,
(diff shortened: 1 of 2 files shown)


libs/providers/flipt/src/lib/flipt-provider.ts
```diff
@@ -169,7 +169,7 @@ export class FliptProvider implements Provider {
         };
       }
 
-      if (!variantEvaluation.match) {
+      if (!variantEvaluation.match && variantEvaluation.reason != 'DEFAULT_EVALUATION_REASON') {
```
beeme1mr, line 172: Is it okay if the reason is "TARGETING_MATCH" when using the default value defined in Flipt?

erka: @beeme1mr it follows the same pattern as `resolveBooleanEvaluation`.

Also the official OpenFeature documentation says
- `DEFAULT` - no dynamic evaluation occurred or dynamic evaluation yielded no result
- `TARGETING_MATCH` - the result of a dynamic evaluation, such as a rule or specific user-targeting

Is there something I'm missing? I don't see any conflict here.

beeme1mr: There's no conflict. I was just confirming that this was the intended behavior.

beeme1mr (approved):

libs/providers/flipt/src/lib/flipt-provider.ts
erka: @beeme1mr thank you!

PR #1071, opened 2024-10-16
chore: update flipt-node provider

markphelps:
## This PR
Upgrades flipt provider to latest flipt-client-node dep

### Related Issues

Fixes #1234523


toddbaert:
I don't think the package.lock will have any impact on consumers, I think you might want to change the minimum version of the package in the package.json.

libs/providers/flipt/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openfeature/flipt-provider",
-  "version": "0.1.1",
+  "version": "0.1.2",
```
markphelps, line 3: do I need to do this or will the bots do it?

beeme1mr (approved):
Approving, but please revert the package version change so it doesn't mess up the publishing step.

beeme1mr (approved):

PR #1100, opened 2024-11-22
ci(flagd): remove sync server as it is part of flagd by now

aepfli:


toddbaert (approved):
Nice - also happy to be using the default port now.

PR #1104, opened 2024-12-05
fix(go-feature-flag-web): avoid infinite loop in waitWebsocketFinalSt…

thomaspoignant:
## This PR
This PR fixes the issue https://github.com/thomaspoignant/go-feature-flag/issues/2753 for the web provider when we had an infinite loop when trying to connect with the websocket and an invalid API key.

### Related Issues

Fixes https://github.com/thomaspoignant/go-feature-flag/issues/2753


thomaspoignant:
Test are broken, I'll fix them.

thomaspoignant:
@augustinasrce I had to change the approach and based the wait state with a timeout.
With this new approach it will still retry to connect with an exponential backoff (check the function `reconnectWebsocket` to see how it works).

I will appreciate a new review.

PR #1117, opened 2024-12-13
feat: Accept timeout for LD waitForInitialization

RedbackThomson:
Add a new option to the LaunchDarkly provider constructor, `initializationTimeout`, which lets users optionally configure the timeout used in the LDClient `waitForInitialization` method. This timeout is critical because it otherwise would cause the LDClient to hang indefinitely in cases where the network connection cannot be established.


kinyoklion (approved):

kinyoklion:
@RedbackThomson You will need to sign-off on your commit. You should be able to amend your commit with the signoff.

RedbackThomson:
Done and force-pushed

toddbaert (approved):

Issue #1118, opened 2024-12-17
[flagd] e2e tests should use test-harness version.txt as test-image version


aepfli:
With the release of flagd-testbed `v0.5.18`, we now have a `version.txt` containing the version for the correlating docker images to be used within the e2e tests.

We should upgrade our git test-harness submodule and use this file to determine the version.

PR #1129, opened 2024-12-29
test(flagd): rework e2e tests to new format

aepfli:
This migrates all flagd e2e tests to the new gherkin format.

All IMPLEMENTED functionality is tested, no coverage has been removed. There's currently a lot of unimplemented functionality in this provider compared to Java. I will create/link issues to the annotation exclusions before merging this issue.

Resolves: https://github.com/open-feature/js-sdk-contrib/issues/1118
Resolves: https://github.com/open-feature/js-sdk-contrib/issues/1282


toddbaert:
I'll take this over if you don't mind @aepfli

libs/providers/flagd/src/e2e/tests/rpc.spec.ts
```diff
@@ -1,56 +1,32 @@
...
+    resolverType: 'rpc',
+    options: {},
+    config: undefined,
+    events: [],
+  };
+  autoBindSteps(
+    loadFeatures(GHERKIN_FLAGD, {
+      tagFilter:
+        // remove filters as we add support for features
+        '@rpc and not @targetURI and not @customCert and not @events and not @stream and not @grace and not @metadata and not @contextEnrichment and not @caching',
```
toddbaert, line 25: If these exceptions make sense to everyone, I will make sure we have issues for each of them and add those issues numbers here (for the annotations we will eventually support).

@aepfli please note that we don't support caching fully here yet since it also relies on the STALE/grace behavior for testing.

libs/providers/flagd/src/e2e/tests/in-process.spec.ts
```diff
@@ -1,60 +1,31 @@
...
+  const state: State = {
+    resolverType: 'in-process',
+    options: {},
+    config: undefined,
+    events: [],
+  };
+  autoBindSteps(
+    loadFeatures(GHERKIN_FLAGD, {
+      // remove filters as we add support for features
+      tagFilter:
```
toddbaert, line 23: https://github.com/open-feature/js-sdk-contrib/pull/1129/files#r2103295114

beeme1mr (approved):

libs/providers/flagd/src/e2e/step-definitions/contextSteps.ts
```diff
@@ -0,0 +1,32 @@
...
+        state.context[key] = mapValueToType(value, type);
+      },
+    );
+    given(
+      /^a context containing a nested property with outer key "(.*)" and inner key "(.*)", with value "(.*)"$/,
+      (outer, inner, value) => {
+        if (state.context == undefined) {
+          state.context = {};
+        }
+        state.context[outer] = { [inner]: value };
```
chrfwow, line 23: Do we want to overwrite existing values, or append to the existing map?

libs/providers/flagd/src/e2e/step-definitions/eventSteps.ts
```diff
@@ -0,0 +1,58 @@
...
+      expect(state.events.find((value) => value.type == type)).toBeDefined();
+      state.events = [];
+    });
+
+    then(/^the (.*) event handler should have been executed within (\d+)ms$/, async (type: string, ms: number) => {
+      await waitFor(() => expect(state.events.find((value) => value.type == type)).toBeDefined(), { timeout: ms });
+      const actual = state.events.find((value) => value.type == type);
+      expect(actual).toBeDefined();
+      state.events = [];
+      console.error('here bin cih');
```
chrfwow, line 45: leftover log

libs/providers/flagd/src/e2e/step-definitions/flagSteps.ts
```diff
@@ -0,0 +1,77 @@
...
+    then(/^the reason should be "(.*)"$/, (expectedReason) => {
+      expect(state.details?.reason).toBe(expectedReason);
+    });
+
+    then(/^the variant should be "(.*)"$/, (expectedVariant) => {
+      expect(state.details?.variant).toBe(expectedVariant);
+    });
+
+    then('the resolved metadata should contain', (table) => {
+      // TODO: implement metadata
```
chrfwow, line 71: Do we leave that over for another PR?

libs/providers/flagd/src/e2e/step-definitions/providerSteps.ts
```diff
@@ -0,0 +1,88 @@
...
+    }, 50000);
+
+    afterAll(async () => {
+      await OpenFeature.close();
+      await container.stop();
+    });
+
+    beforeEach(async () => {
+      // TODO: I'm not sure what the point of this is
+      if (container.isStarted()) {
```
chrfwow, line 24: Could this be missing a `!`?

libs/providers/flagd/src/e2e/step-definitions/providerSteps.ts
```diff
@@ -0,0 +1,88 @@
...
+        return await container.start();
+      }
+      return Promise.resolve();
+    }, 50000);
+
+    afterEach(async () => {
+      // everything breaks without this
+      await OpenFeature.close();
+      if (state.client) {
+        await fetch('http://' + container.getLaunchpadUrl() + '/stop');
```
chrfwow, line 34: Don't we have a `container.stop()` method?

libs/providers/flagd/src/e2e/step-definitions/providerSteps.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      state.client = OpenFeature.getClient(providerType);
+      state.providerType = providerType;
+    });
+
+    when(/^the connection is lost for (\d+)s$/, async (time) => {
+      console.log('stopping flagd');
+      await fetch('http://' + container.getLaunchpadUrl() + '/restart?seconds=' + time);
+      await new Promise((r) => setTimeout(r, 1000));
```
chrfwow, line 81: In theory, these waits should not be needed any more, as the container now only responds when it is done

libs/providers/flagd/src/e2e/step-definitions/providerSteps.ts
aepfli: we do not want to stop the container, we want to stop flagd inside the container but the container should still be running. else we have to somehow ensure that we use the right port when starting the container again, etc. and it gets just tideous especially to handle this in all language. That is the reason we have the launchpad

libs/providers/flagd/src/e2e/step-definitions/providerSteps.ts
aepfli: I think we can remove this line. this was pre launchpad, where we started and stopped the container manually

toddbaert: Confirmed we don't need it!

libs/providers/flagd/src/e2e/step-definitions/providerSteps.ts
toddbaert: Yep, you're right. At one point I needed these or events tests broke, but I must have fixed it with some other change. Now everything passes fine. :heart:

aepfli (commented):
I can not approve this pr, because I am the original owner, but everything looks good to me, thank you for taking it over

libs/providers/flagd/src/e2e/step-definitions/configSteps.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+    when('a config was initialized', () => {
+      state.config = getConfig(state.options);
+    });
+
+    then(
+      /^the option "(.*)" of type "(.*)" should have the value "(.*)"$/,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      (_name: string, _type: string, _value: string) => {
+        // TODO: implement with configuration unification
```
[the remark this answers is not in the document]
toddbaert: Intentional, but I've added a a link to the related issue.

PR #1147, opened 2025-01-07
chore: update nx packages

toddbaert:
This PR: 

- updates to nx@latest (had to do 17, then latest to get new lint modules to work). I did:
  - `npx nx migrate 17`
  - `npx nx migrate latest`
  - then just run the lint fix and manually fixed a couple lint issues
  - removed deprecated `workspace-lint` from CI (this validates the project config, but it's automatic now): https://github.com/nrwl/nx/pull/16212
  - added the new lint rule below
- updates @types/node (required to fix new compilation issues)
- updates the generators

:warning: 

There is one significant change in how this repo works: _we no longer automatically compute the deps for all packages, BUT we instead throw a lint error if nx calculates that you have failed to include a dep in your package.json which you import in your code._

For example, flagd uses the `lru-cache` package. Because we import it in our ts files, if we DON'T include it in our pacakge.json, we see this lint error (which causes the build to fail):

```
nx run providers-flagd:lint
      Linting "providers-flagd"...
      
      /home/todd/git/js-sdk-contrib/libs/providers/flagd/package.json
        9:3  error  The "providers-flagd" project uses the following packages, but they are missing from "peerDependencies":
          - lru-cache  @nx/dependency-checks
      
      ✖ 1 problem (1 error, 0 warnings)
        1 error and 0 warnings potentially fixable with the `--fix` option.
```

I think this is actually an improvement over the last behavior, since it will cause devs to think carefully about their deps.

:warning: 

I have updated all the deps for all the packages appropriately according to this check. Please check the deps of your package.json.


thomaspoignant (approved):

libs/providers/flipt-web/src/lib/flipt-web-provider.ts
```diff
@@ -66,7 +66,7 @@ export class FliptWebProvider implements Provider {
     }
   }
 
-  async onContextChange(_oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
+  async onContextChange(): Promise<void> {
```
lukas-reining, line 69: Did you do this manually @toddbaert?
Or did the linting change with the deps?

lukas-reining (commented):
Good that we update that!

libs/providers/flipt-web/src/lib/flipt-web-provider.ts
toddbaert: This was linting. They were both unused and the linter now complains about that. We could disable the rule if you think it's a bad idea.

lukas-reining: No it makes sense to me :) 
Just wanted to know

toddbaert:
Unfortunately, the nx-computed dependencies created by `updateBuildableProjectDepsInPackageJson` no longer work (it was deprecated); this means we aren't properly computing deps at package time anymore. There's a new way to do this I need to get working.

toddbaert:
> Unfortunately, the nx-computed dependencies created by `buildableProjectDepsInPackageJsonType` no longer work (it was removed); this means we aren't properly computing deps at package time anymore. There's a new way to do this I need to get working.
> 
> Marking draft until I figure this out.

Looks like we can use `@nx/dependency-checks` for a comparable feature. I almost have it working.

toddbaert:
:warning: 

There is one significant change in how this repo works: _we no longer automatically compute the deps for all packages, BUT we instead throw a lint error if nx calculates that you have failed to include a dep in your package.json which you import in your code._

For example, flagd uses the `lru-cache` package. Because we import it in our ts files, if we DON'T include it in our pacakge.json, we see this lint error (which causes the build to fail):

```
nx run providers-flagd:lint
      Linting "providers-flagd"...
      
      /home/todd/git/js-sdk-contrib/libs/providers/flagd/package.json
        9:3  error  The "providers-flagd" project uses the following packages, but they are missing from "peerDependencies":
          - lru-cache  @nx/dependency-checks
      
      ✖ 1 problem (1 error, 0 warnings)
        1 error and 0 warnings potentially fixable with the `--fix` option.
```

I think this is actually an improvement over the last behavior, since it will cause devs to think carefully about their deps.

:warning: 

I have updated all the deps for all the packages appropriately according to this check. Please check the deps of your package.json.

.eslintrc.json
```diff
@@ -28,17 +25,41 @@
...
+        "no-extra-semi": "off"
+      }
     },
     {
       "files": "*.json",
       "parser": "jsonc-eslint-parser",
       "rules": {}
+    },
+    {
+      "files": ["*.json"],
```
toddbaert, line 47: This is the new lint rule which ensures projects have all the dependencies they need.

libs/providers/flagd/src/lib/service/in-process/file/file-fetch.spec.ts
```diff
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import { FileFetch } from './file-fetch';
 import { FlagdCore } from '@openfeature/flagd-core';
-import { Logger } from '@openfeature/core';
+import { Logger } from '@openfeature/server-sdk';
```
toddbaert, line 4: We should avoid direct imports from core in server/web libs

libs/providers/flipt-web/src/lib/context-transformer.ts
```diff
@@ -1,4 +1,4 @@
-import { EvaluationContext } from '@openfeature/server-sdk';
```
toddbaert, line 1: we we're importing from server in a web lib :fearful: 

This was caught because the new lint tried to add the server-sdk to the deps.

libs/providers/flagd/docker-compose.yaml
```diff
@@ -1,17 +0,0 @@
-services:
```
toddbaert, line 1: We now use TestContainers to start/stop all containers during testing.

lukas-reining (approved):
Looks good to me! Nice update!

.eslintrc.json
beeme1mr: Should this be `*.json` or something more strict like `**/package.json`?

.prettierignore
```diff
@@ -2,3 +2,6 @@
 
 /dist
 /coverage
+
+/.nx/cache
```
beeme1mr, line 6: Can this just be `/.nx`?

jest.config.ts
```diff
@@ -1,5 +1,5 @@
-import { getJestProjects } from '@nx/jest';
+import { getJestProjectsAsync } from '@nx/jest';
 
-export default {
-  projects: getJestProjects(),
-};
+export default async () => ({
+  projects: await getJestProjectsAsync(),
```
beeme1mr, line 4: Do you know why this chance was necessary? It doesn't really matter, I'm just curious.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -1,5 +1,5 @@
 import { ClientReadableStream, ServiceError, credentials } from '@grpc/grpc-js';
-import { Logger } from '@openfeature/core';
+import { Logger } from '@openfeature/server-sdk';
 import { GeneralError } from '@openfeature/server-sdk';
```
beeme1mr, line 3: Please combine these imports.

libs/providers/flagsmith-client/package.json
```diff
@@ -1,9 +1,7 @@
 {
   "name": "@openfeature/flagsmith-client-provider",
   "version": "0.1.2",
-  "dependencies": {
-    "tslib": "^2.3.0"
```
beeme1mr, line 5: Is `tslib` not needed?

libs/providers/ofrep/package.json
```diff
@@ -1,9 +1,7 @@
 {
   "name": "@openfeature/ofrep-provider",
   "version": "0.2.0",
-  "dependencies": {
-    "tslib": "^2.3.0"
```
beeme1mr, line 5: Is this no longer required?

libs/providers/ofrep/project.json
```diff
@@ -18,24 +18,14 @@
...
+      "outputs": ["{options.outputFile}"]
     },
     "test": {
       "executor": "@nx/jest:jest",
       "outputs": ["{workspaceRoot}/coverage/{projectRoot}"],
       "options": {
-        "jestConfig": "libs/providers/ofrep/jest.config.ts",
-        "passWithNoTests": true
-      },
-      "configurations": {
```
beeme1mr, line 34: Did you remove this config or was it done as part of the migration?

libs/shared/flagd-core/package.json
```diff
@@ -11,6 +11,9 @@
   },
   "dependencies": {
     "ajv": "^8.12.0",
-    "tslib": "^2.3.0"
+    "object-hash": "^3.0.0",
+    "imurmurhash": "^0.1.4",
+    "semver": "7.6.3",
+    "json-logic-engine": "4.0.2"
```
beeme1mr, line 17:
```suggestion
    "json-logic-engine": "^4.0.2"
```

libs/shared/flagd-core/package.json
```diff
@@ -11,6 +11,9 @@
   },
   "dependencies": {
     "ajv": "^8.12.0",
-    "tslib": "^2.3.0"
+    "object-hash": "^3.0.0",
+    "imurmurhash": "^0.1.4",
+    "semver": "7.6.3",
```
beeme1mr, line 16:
```suggestion
    "semver": "^7.6.3",
```

.prettierignore
toddbaert: I didn't add this, so I guess it was an automation I'd rather not touch.

libs/providers/ofrep/package.json
toddbaert: It's not

libs/providers/flagsmith-client/package.json
toddbaert: It's not

libs/providers/ofrep/project.json
toddbaert: I did not manually remove it.

jest.config.ts
toddbaert: It was automated. I saw it too and thought it was interesting.

I would guess this just blocked before, and now because it's async it's note and that probably allows the node process to do whatever other things while it waits for this, hypothetically improving speed.

.eslintrc.json
toddbaert: This is what I saw in most of the examples. I'm not sure if it maybe considers the package.json too somehow: https://nx.dev/nx-api/eslint-plugin/documents/dependency-checks#manual-setup

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
toddbaert: done

beeme1mr (approved):

kinyoklion (approved):

jonathannorris (approved):
> Unfortunately, the nx-computed dependencies created by `buildableProjectDepsInPackageJsonType` no longer work (it was removed); this means we aren't properly computing deps at package time anymore. There's a new way to do this I need to get working.

We've also had this issue while trying to upgrade our NX usage, these NX migrations are not for the faint of heart.

PR #1200, opened 2025-01-31
feat(ofrep-core): add abort timeout to fetch call

beeme1mr:
## This PR

- adds a `timeoutMs` option that defaults to 10 seconds

### Notes

Uses a signal timeout to aborts slow fetch requests. Here's an interesting article about abort controllers if you're interested.

https://kettanaito.com/blog/dont-sleep-on-abort-controller

### Follow-up Tasks

We should consider bumping  the min peer version of the OFREP providers to 

### How to test

Unit tests have been added.

### DIFF
--- a/libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
+++ b/libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
@@ -49,9 +49,10 @@ function isomorphicFetch(): FetchAPI {
   return fetch;
 }
 
+const DEFAULT_TIMEOUT_MS = 10_000;
+
 export class OFREPApi {
   private static readonly jsonRegex = new RegExp(/application\/[^+]*[+]?(json);?.*/, 'i');
-  private _etag?: string;
 
   constructor(
     private baseOptions: OFREPProviderBaseOptions,
@@ -74,7 +75,9 @@ export class OFREPApi {
   private async doFetchRequest(req: Request): Promise<{ response: Response; body?: unknown }> {
     let response: Response;
     try {
-      response = await this.fetchImplementation(req);
+      response = await this.fetchImplementation(req, {
+        signal: AbortSignal.timeout(this.baseOptions.timeoutMs ?? DEFAULT_TIMEOUT_MS),
+      });
     } catch (err) {
       throw new OFREPApiFetchError(err, 'The OFREP request failed.', { cause: err });
     }
--- a/libs/shared/ofrep-core/src/lib/api/ofrep-api.spec.ts
+++ b/libs/shared/ofrep-core/src/lib/api/ofrep-api.spec.ts
@@ -24,344 +24,374 @@ describe('OFREPApi', () => {
     server.listen();
   });
   beforeEach(() => {
-    jest.useFakeTimers();
-    api = new OFREPApi({ baseUrl: 'https://localhost:8080' });
+    api = new OFREPApi({
+      baseUrl: 'https://localhost:8080',
+      // Short timeout to speed up tests
+      timeoutMs: 500,
+    });
   });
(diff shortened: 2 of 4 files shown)


libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -49,9 +49,10 @@ function isomorphicFetch(): FetchAPI {
   return fetch;
 }
 
+const DEFAULT_TIMEOUT_MS = 10_000;
```
beeme1mr, line 52: I think 10 seconds is a reasonable default but I'd like others opinion on this.

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -102,6 +105,23 @@ export class OFREPApi {
     }
   }
 
+  /**
+   * Create a timeout signal that will abort the request after the timeout.
+   */
+  private createTimeoutSignal(timeoutMs = DEFAULT_TIMEOUT_MS): AbortSignal | undefined {
```
beeme1mr, line 111: I added this method to be as defensive as possible. I don't love this solution though. Please let me know if there's a better way to handle this.

beeme1mr: Unfortunately, `AbortSignal.timeout` is too new to use it without a polyfill.

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static

lukas-reining: This looks fine to me. I also think it is acceptable to just implement this here. 
`AbortSignal.timeout` is good but I think replacing it with the two lines is fine here.

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -102,6 +105,23 @@ export class OFREPApi {
...
+      const controller = new AbortController();
+      // Uses a setTimeout instead of AbortSignal.timeout to support older browsers.
+      setTimeout(
+        () => controller.abort(new DOMException(`This signal is timeout in ${timeoutMs}ms`, 'TimeoutError')),
+        timeoutMs,
+      );
+      return controller.signal;
+    } catch {
+      return undefined;
+    }
```
lukas-reining, line 122: What do we have to try here? As far as I can see nothing from this throws but I guess I am overlooking something.

lukas-reining (approved):
Looks good to me! Good addition!

toddbaert (approved):
Approved, but consider @lukas-reining 's question about the try/catch.

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
beeme1mr: I've removed this

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -74,7 +75,16 @@ export class OFREPApi {
   private async doFetchRequest(req: Request): Promise<{ response: Response; body?: unknown }> {
     let response: Response;
     try {
-      response = await this.fetchImplementation(req);
+      const timeoutMs = this.baseOptions.timeoutMs ?? DEFAULT_TIMEOUT_MS;
+      const controller = new AbortController();
+      // Uses a setTimeout instead of AbortSignal.timeout to support older browsers.
```
beeme1mr, line 80:
```suggestion
      // Uses a setTimeout instead of AbortSignal.timeout to support older runtimes.
```

PR #1202, opened 2025-02-04
feat: support proxy routing via gRPC default_authority

raflFaisal:
## This PR
To provide a :authority header to request route to target backend services using default_authority parameter by the JS clients. This enable better integration with service proxies like Envoy, facilitating more flexible and robust gRPC client configurations.

### Related Issues
https://github.com/open-feature/js-sdk-contrib/issues/1187
https://github.com/open-feature/flagd/issues/1532

### How to test
Use defaultAuthority parameter to pass down the target authority when sending a grpc request from client side.

### DIFF
--- a/libs/providers/flagd/README.md
+++ b/libs/providers/flagd/README.md
@@ -96,7 +96,8 @@ This mode is useful for local development, test cases, and for offline applicati
 
 ### Default Authority usage (optional)
 
-This is useful for complex routing or service-discovery usecases that involve a proxy (e.g., Envoy). Please refer to this [link](https://github.com/open-feature/js-sdk-contrib/issues/1187) for more information.
+This is useful for complex routing or service-discovery use cases that involve a proxy (e.g., Envoy).
+Please refer to this [GitHub issue](https://github.com/open-feature/js-sdk-contrib/issues/1187) for more information.
 
 ```ts
   OpenFeature.setProvider(new FlagdProvider({
--- a/libs/providers/flagd/README.md
+++ b/libs/providers/flagd/README.md
@@ -38,6 +38,7 @@ Options can be defined in the constructor or as environment variables. Construct
 | selector                               | FLAGD_SOURCE_SELECTOR          | string  | -                                                              |                  |
 | cache                                  | FLAGD_CACHE                    | string  | lru                                                            | lru, disabled    |
 | maxCacheSize                           | FLAGD_MAX_CACHE_SIZE           | int     | 1000                                                           |                  |
+| defaultAuthority                       | FLAGD_DEFAULT_AUTHORITY        | string  | -                                                              | rpc, in-process  |
 
 #### Resolver type-specific Defaults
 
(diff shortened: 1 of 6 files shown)


libs/providers/flagd/README.md
```diff
@@ -38,6 +38,7 @@ Options can be defined in the constructor or as environment variables. Construct
 | selector                               | FLAGD_SOURCE_SELECTOR          | string  | -                                                              |                  |
 | cache                                  | FLAGD_CACHE                    | string  | lru                                                            | lru, disabled    |
 | maxCacheSize                           | FLAGD_MAX_CACHE_SIZE           | int     | 1000                                                           |                  |
+| defaultAuthority                       | FLAGD_DEFAULT_AUTHORITY        | string  | -                                                              | rpc, in-process  |
```
beeme1mr, line 41: Reminder to myself to include this configuration option in the flagd provider spec.

https://flagd.dev/reference/specifications/providers/#configuration

libs/providers/flagd/src/lib/configuration.ts
```diff
@@ -68,6 +68,13 @@ export interface Config {
    * @default 1000
    */
   maxCacheSize?: number;
+
+  /**
+   * The target host (authority) when routing requests through a proxy (e.g. Envoy)
+   *
+   * @default ''
```
beeme1mr, line 75: I believe this is technically undefined but an empty string may be more clear in the docs.

beeme1mr (approved):

libs/providers/flagd/src/lib/configuration.ts
raflFaisal: Thanks for pointing this. I will remove this line to keep in sync with other param (e.g. offlineFlagSourcePath)

libs/providers/flagd/README.md
toddbaert: I'm already making some flagd doc changes, I can do it.

toddbaert:
I'm OK with this solution, but it's distinctly different from the solution in other providers as you noted [here](https://github.com/open-feature/js-sdk-contrib/issues/1187):

> Unfortunately, for Node.js, such support for Custom Name Resolution is not available.

While I'm OK with this solution, the title of this PR (which is what will show up in the release notes) doesn't seem accurate. It can address a subset of the same problems of a custom name resolver, but unless I'm very confused, it's not a custom name resolver.

raflFaisal:
@toddbaert You are correct, let me change the PR title that matches with the code change.

"feat: support proxy routing via gRPC default_authority"

Let me know if it looks good to you

toddbaert:
> @toddbaert You are correct, let me change the PR title that matches with the code change.
> 
> "feat: support proxy routing via gRPC default_authority"
> 
> Let me know if it looks good to you

Yep, that sounds good to me. Just make sure to update and readme or inline docs as well to say the same sort of thing.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
```diff
@@ -27,13 +27,20 @@ export class GrpcFetch implements DataFetch {
   private _isConnected = false;
 
   constructor(config: Config, syncServiceClient?: FlagSyncServiceClient, logger?: Logger) {
-    const { host, port, tls, socketPath, selector } = config;
+    const { host, port, tls, socketPath, selector, defaultAuthority } = config;
+    let clientOptions: ClientOptions | undefined;
```
toddbaert, line 31: nit: Unless the types are pretty weird, I'd think you could just make this an empty-object const, and then add the `grpc.default_authority` conditionally, which reads better to me.

toddbaert (approved):
Approved with one minor nit here: https://github.com/open-feature/js-sdk-contrib/pull/1202/files#r1945277742

Looks really good overall. Thanks for the good README and explanations!

beeme1mr:
I'll merge and release tomorrow. Please consider Todd's comment but I don't think it's a blocker.

PR #1229, opened 2025-03-17
fix(gofeatureflag): Error for configurationHasChanged when slash was missing in endpoint

chrillebile:
## This PR

If the endpoint did not have a trailing slash then configurationHasChanged request would not work correctly, as it would try to do a request on `https://example.comv1/flag/change` instead of `https://example.com/v1/flag/change`
URL(this.endpoint) is being used for all other requests to mitigate this, except for configurationHasChanged.

### How to test

Before you could not set the endpoint to `https://example.com`, you had to set `https://example.com/`.
Now you can set `https://example.com` or `https://example.com/`.


toddbaert:
@thomaspoignant any concerns with this?

PR #1221, opened 2025-03-12
feat(aws-ssm): implement AWS SSM provider

gdegiorgio:
> [!NOTE]  
> This PR is only a proposal, I thought adding a widely used service like AWS SSM might have value but I don't know if it fits in the roadmap anyway.

## This PR

Create a new provider : AWS SSM

AWS SSM (Systems Manager) is a service that helps automate operational tasks across AWS resources.
One of its key components is AWS SSM Parameter Store, which is a secure and scalable storage solution for configuration data and secrets.

- Retrieve feature flags via SSM Paramter Store
- Use a configurable LRU Cache in order to saving calls to AWS

PR #1236, opened 2025-03-19
fix(generator): update generated path  `project.json`, `jest.config.ts`, `.eslintrc.json`

gdegiorgio:
## This PR

This PR fix wrong path generation in freshly created provider after `npm run generate-provider` command

### Related Issues

Running `test` and `lint` commands on a freshly generated provider ( by using `npm run generate-provider`) throws errors  which makes harder starting contribution on the project

Steps to reproduce: 

- Run `npm run generate-provider` and create a new `example` provider
- Launch `npx nx run example:lint` and you will get ` NX   Failed to load config "../.eslintrc.json" to extend from.` error
- Launch `npx nx run example:test` and you will get  ` NX   Can't find a root directory while resolving a config file path.` error

### How to test

Repeat all steps and you will be able to run lint and test command on a freshly created provider

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,9 +31,9 @@ jobs:
       - run: git branch --track main origin/main || true
 
       - run: if ! npx nx format:check ; then echo "Format check failed. Please run 'npx nx format:write'."; fi
-      - run: npx nx affected --target=lint --parallel=3
-      - run: npx nx affected --target=test --parallel=3 --ci --code-coverage
-      - run: npx nx affected --target=build --parallel=3
+      - run: npx nx affected --target=lint --parallel=3 --exclude=js-sdk-contrib
+      - run: npx nx affected --target=test --parallel=3 --ci --code-coverage --exclude=js-sdk-contrib
+      - run: npx nx affected --target=build --parallel=3 --exclude=js-sdk-contrib
 
   e2e:
       runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


beeme1mr (approved):
Wow, thank you for looking into this. NX must have changed the generator behavior recently. FYI @toddbaert

toddbaert:
Just a reminder to make sure this doesn't add the root project.json as mentioned [here](https://github.com/open-feature/js-sdk-contrib/pull/1221#discussion_r2006304543).

gdegiorgio:
> Just a reminder to make sure this doesn't add the root project.json as mentioned [here](https://github.com/open-feature/js-sdk-contrib/pull/1221#discussion_r2006304543).

Hi @toddbaert , I've opened #1240 in order to prevent to add generated root `project.json` in VCS. Currently the problem seems to be `@nx/js` plugin, we should look further into it.

toddbaert:
@gdegiorgio I was unable to get this to work in combination with https://github.com/open-feature/js-sdk-contrib/pull/1240.

I think we need to just "bite the bullet" and finally deal with verdaccio and the project.json that keeps getting added to the root. To remove the infinite loop in all our runs, I've added `--exclude=js-sdk-contrib` to all our scripts. Otherwise, we get stuck in an infinite cycle.

Hopefully there is no side effects cc @beeme1mr @lukas-reining

toddbaert (approved):

PR #1240, opened 2025-03-20
chore: ignore root generated `project.json`

gdegiorgio:
This PR adds .verdaccio inside .gitignore in order to prevent that folder to be added to VCS.

# Related Issues

When running the open-feature nx generator with publishable : true in libraryGenerator options, this will cause @nx/js to create an unwanted verdaccio configuration along with a root `project.json` causing `e2e` CI task to infinite loop 

(See comments in #1221)

From @nx/js [20.3.2](https://github.com/nrwl/nx/blob/20.3.1/packages/js/src/generators/library/library.ts)

  if (options.publishable) {
    tasks.push(await setupVerdaccio(tree, { ...options, skipFormat: true }));
  }
  
  
This issue would require further investigation but would prevent newcomers to contribute easily to project. 
  
## How to test
Generating a new provider via npm run generate-provider will create a root `project.json` that will not be added to VCS


toddbaert:
See: https://github.com/open-feature/js-sdk-contrib/pull/1236#issuecomment-2752018931

PR #1242, opened 2025-03-26
fix(config-cat): Rework error reporting

### LINKED ISSUE #1212: Config-cat-web causes infinite rerenders with @openfeature/react-sdk
Testing the ConfigCat provider, it's causing infinite rerenders when the key doesn't exist. As far as I'm aware, per the spec, when an error occurs, the default value must be returned, which for the InMemoryProvider happens, but not with Config Cat's. 

Image

If you want to reproduce the issue, you can use [juandelacruz23/openfeature-configcat-error](https://github.com/juandelacruz23/openfeature-configcat-error). You can create your own sdk key at https://app.configcat.com/auth/startnow and supply it through a .env.

adams85:
## This PR

Attempts to fix #1212 by reworking error reporting. (More on the reason for the change in the related issue.)

### Related Issues

Fixes #1212

### Notes

This PR is not meant to be a complete solution, more of a quick fix to the related issue. It should be revised and improved later.

### Follow-up Tasks

n/a

### How to test

n/a


libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
```diff
@@ -137,13 +141,22 @@ export class ConfigCatWebProvider implements Provider {
...
+      // Ideally, we would check ConfigCat client's initialization state in its "background" polling loop.
+      // This is not possible at the moment, so as a workaround, we do the check on feature flag evaluation.
+      // There are plans to improve this situation, so let's revise this
+      // as soon as ConfigCat SDK implements the necessary event.
 
-    if (this._hasError && !evaluationData.errorMessage && !evaluationData.errorException) {
-      this._hasError = false;
-      this.events.emit(ProviderEvents.Ready);
+      this._isProviderReady = true;
+      setTimeout(() => this.events.emit(ProviderEvents.Ready), 0);
```
beeme1mr, line 159: Is there a reason to not emit this event right away? Also, this will likely trigger a re-render in the React SDK but it should no longer trigger a re-render loop.

beeme1mr (commented):
This approach seems good to me. I think it's a good idea to offload the majority of the status logic to the SDK.

libs/providers/config-cat-web/src/lib/config-cat-web-provider.ts
adams85: It's just that it doesn't feel right to invoke unknown user code during the evaluation of a feature flag, when normally that code would be invoked asynchronously, on the client's "background" polling loop. (As I mentioned in the related issue, the necessary event is not emitted by ConfigCat client at the moment but it's planned to be implement soon.)

Regardless, I'm ok with emitting the event synchronously if you say it's fine.

beeme1mr: The event would take two event loop cycles to run but I don't think it's a big deal either way. It's up to you.

adams85: Well, then I'd leave it as is.

lukas-reining:
Sorry for the delay @adams85, many of us were at KubeCon last week. 
Will review tomorrow :)

lukas-reining (approved):
Looks good to me!

adams85:
Can we go ahead and merge this?

lukas-reining:
Yes, looks good to me @adams85!

PR #1268, opened 2025-04-10
feat(goff web): Support tracking events

thomaspoignant:
## This PR
Support tracking events for GO Feature Flag


libs/providers/go-feature-flag-web/src/lib/model.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { ErrorCode, EvaluationContextValue, FlagValue } from '@openfeature/web-sdk';
+import { ErrorCode, EvaluationContext, EvaluationContextValue, FlagValue } from '@openfeature/web-sdk';
+import { TrackingEventDetails } from '@openfeature/core';
```
beeme1mr, line 3: Is this not exported from `@openfeature/web-sdk`? If not, we should probably do that.

thomaspoignant: Good catch, I have fixed it.

PR #1351, opened 2025-07-10
fix(flagsmith): Fetch identity flags when context changes

### LINKED ISSUE #1232: Flagsmith: Inconsistent Identity Propagation in `FlagsmithClientProvider` on Context Updates
### **Description**  
When using Flagsmith as an OpenFeature provider, updating the context does **not always** trigger an identity API call. This results in:  
- Missing traits in the `/identities` API request  
- Incorrect segmentation behavior  

The issue arises when the provider is set **before** the context.

#### ✅ **Working Flow**  
This sequence correctly triggers the `/identities` endpoint with the expected traits:  
```javascript
const flagsmithClient = new FlagsmithClient();
await OpenFeature.setContext({ targetingKey: 'id', traitA: 'abc' });
OpenFeature.setProvider(flagsmithClient);
```  

#### ❌ **Non-Working Flow**  
In this case, the identity API call is **not** made after setting the context, resulting in empty traits:  
```javascript
const flagsmithClient = new FlagsmithClient();
OpenFeature.setProvider(flagsmithClient);
await OpenFeature.setContext({ targetingKey: 'id', traitA: 'abc' });
```

### **Observed Behavior**  
- Without a `targetingKey`, segmentation defaults to environment-level settings.  
- Features that should be disabled **remain enabled**, with a "STATIC" reason.  
- Updating the context **after** setting the provider does **not** trigger a re-fetch with the correct identity data.  

### **Expected Behavior**  
- Whenever context is updated, the provider should correctly propagate identity and traits and fetch correct state of flags for target user.  

### **Possible Cause**  
In `FlagsmithClientProvider`, when the context updates, the provider calls:  
```javascript
this._client.getFlags();
```  
However, this does **not** include the new identity information. Updating this call to:  
```javascript
this._client.identify(identity, context);
```
resolves the issue.  

🔍 **[[Problematic line in the source code](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts#L56)](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts#L56).**  

### **Environment**  
- OpenFeature JS SDK with Flagsmith as the provider.

rolodato:
## This PR

Use updated traits in the evaluation context whenever it changes.

Various README improvements.

### Related Issues

Fixes #1232

### Notes

Tests pass, but I wasn't able to figure out how to type the fix properly to avoid the `traits as any` and an ESLint `no-unused-vars` warning.

### Follow-up Tasks

N/A

### How to test

Added a failing test which is fixed by this PR.


rolodato:
One for @kyle-ssg to have a look maybe? I can't assign reviewers for this PR while it's still in draft.

libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts
```diff
@@ -52,7 +52,24 @@ export class FlagsmithClientProvider implements Provider {
         ...(context || {}),
       });
       this.events.emit(ProviderEvents.Stale, { message: 'context has changed' });
-      return isLogout ? this._client.logout() : this._client.getFlags();
+      if (isLogout) {
+        return this._client.logout();
```
beeme1mr, line 56: I see this isn't new behavior but I'm curious why this is necessary. Do you know why the logout would need to be called during initialization?

rolodato: It confused me as well for sure. I speculate that this was a shortcut, since this provider implements `onContextChange` by directly calling `initialize`: https://github.com/rolodato/js-sdk-contrib/blob/453c70cf829cfee02720e73d78736ddc89d82ce3/libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts#L106

A better solution could have been to keep the [post-initialization logic](https://github.com/rolodato/js-sdk-contrib/blob/453c70cf829cfee02720e73d78736ddc89d82ce3/libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts#L45-L73) in `onContextChange`, and only have initialization logic in `initialize`.

After this fix, we'd still need to upgrade the Flagsmith SDK dependency to v9, since this provider uses a quite out of date version. I'd prefer leaving this PR as-is and make the code nicer when we're making that upgrade.

PR #1358, opened 2025-07-16
feat: demo null value defaulting

toddbaert:
Demo only.

### DIFF
--- a/libs/shared/ofrep-core/src/test/handlers.ts
+++ b/libs/shared/ofrep-core/src/test/handlers.ts
@@ -121,14 +121,21 @@ export const handlers = [
       }
 
       const scopeValue = new URL(info.request.url).searchParams.get('scope');
+      const key = info.params.key;
+      const reason = key.includes('null-value')
+        ? StandardResolutionReasons.DEFAULT
+        : requestBody.context?.targetingKey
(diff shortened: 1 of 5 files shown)


libs/shared/ofrep-core/src/test/handlers.ts
```diff
@@ -121,14 +121,20 @@ export const handlers = [
...
 
       const scopeValue = new URL(info.request.url).searchParams.get('scope');
+      const key = info.params.key;
+      const reason = key.includes('null-value')
+        ? StandardResolutionReasons.DEFAULT
+        : requestBody.context?.targetingKey
+          ? StandardResolutionReasons.TARGETING_MATCH
+          : StandardResolutionReasons.STATIC;
+
+      const value = key.includes('null-value') ? null : true;
```
toddbaert, line 131: Send back a `value=null` and `reason=DEFAULT`, this is my preferred proposal for implementing a "defer to code default" evaluation in flagd.

lukas-reining (commented):
This is exactly how I imagined it :)

libs/shared/ofrep-core/src/test/handlers.ts
beeme1mr: Would we need to send a variant at all?

PR #1360, opened 2025-07-31
feat: adds rocketflag provider

jgunnink:
## This PR

- Adds [RocketFlag](https://rocketflag.app) as a new provider

### Related Issues

Go SDK provider: https://github.com/open-feature/go-sdk-contrib/pull/725

### Notes

This is my first contribution to this repository. I welcome any suggestions and nitpicks so this code can be as useful to the open feature community as possible.

### Follow-up Tasks

I'll update the documentation with links to the providers once this is merged in and available.

### How to test

`npx nx run RocketFlag:test` from the root of the project

### DIFF
--- a/libs/providers/rocketflag/README.md
+++ b/libs/providers/rocketflag/README.md
@@ -1,0 +1,15 @@
+# RocketFlag Provider
+
+## Installation
+
+```
+$ npm install @openfeature/rocketflag-provider
+```
+
+## Building
+
+Run `nx package providers-rocketflag` to build the library.
+
+## Running unit tests
+
+Run `nx test providers-rocketflag` to execute the unit tests via [Jest](https://jestjs.io).
--- a/libs/providers/rocketflag/package.json
+++ b/libs/providers/rocketflag/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "@openfeature/rocketflag-provider",
+  "version": "0.0.1",
+  "dependencies": {
+    "tslib": "^2.3.0"
+  },
+  "main": "./src/index.js",
+  "typings": "./src/index.d.ts",
+  "scripts": {
+    "publish-if-not-exists": "cp $NPM_CONFIG_USERCONFIG .npmrc && if [ \"$(npm show $npm_package_name@$npm_package_version version)\" = \"$(npm run current-version -s)\" ]; then echo 'already published, skipping'; else npm publish --access public; fi",
+    "current-version": "echo $npm_package_version"
+  },
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "@openfeature/web-sdk": "^1.6.0"
+  }
+}
--- a/libs/providers/rocketflag/src/lib/rocketflag-provider.spec.ts
+++ b/libs/providers/rocketflag/src/lib/rocketflag-provider.spec.ts
@@ -1,0 +1,135 @@
+import type { EvaluationContext } from '@openfeature/web-sdk';
+import { OpenFeature, StandardResolutionReasons, ErrorCode } from '@openfeature/web-sdk';
+import type { FlagStatus, UserContext } from './rocketflag-provider';
+import { RocketFlagProvider } from './rocketflag-provider';
+
+// Create a mock RocketFlag client for testing
+const mockClient = {
+  getFlag: jest.fn<Promise<FlagStatus>, [string, UserContext]>(),
+};
+
+// Mock Logger
+const mockLogger = {
+  info: jest.fn(),
+  warn: jest.fn(),
+  error: jest.fn(),
+  debug: jest.fn(),
+};
+
+describe('RocketFlagProvider', () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  it('should have the correct metadata name', () => {
+    const provider = new RocketFlagProvider(mockClient);
+    expect(provider.metadata.name).toBe('RocketFlagProvider');
+  });
+
+  describe('resolveBooleanEvaluation', () => {
+    it('should return STALE initially, then resolve to the correct value with TARGETING_MATCH', async () => {
+      const provider = new RocketFlagProvider(mockClient);
+      const flagKey = 'test-flag-targeting';
+      const targetingContext: EvaluationContext = { targetingKey: 'user@example.com' };
+
+      mockClient.getFlag.mockResolvedValue({ enabled: true });
+
+      const initialDetails = provider.resolveBooleanEvaluation(flagKey, false, targetingContext, mockLogger);
+      expect(initialDetails.reason).toBe(StandardResolutionReasons.STALE);
+      expect(initialDetails.value).toBe(false);
+
+      await new Promise((resolve) => setTimeout(resolve, 0));
+
+      const finalDetails = provider.resolveBooleanEvaluation(flagKey, false, targetingContext, mockLogger);
+
+      expect(finalDetails.value).toBe(true);
+      expect(finalDetails.reason).toBe(StandardResolutionReasons.TARGETING_MATCH);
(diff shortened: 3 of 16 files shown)


lukas-reining:
Hey @jgunnink, sorry for the delay, I will have a look tomorrow :)

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+      const cacheKey = JSON.stringify({ flagKey, context });
+
+      // Fetch in the background.
+      fetchFlagAndUpdateCache(flagKey, defaultValue, context, cacheKey, client, cache, emitter, logger);
+
+      // Immediately return a cached value if available.
+      if (cache.has(cacheKey)) {
+        // The .get() method can return undefined, so we handle that case.
+        return cache.get(cacheKey) as ResolutionDetails<boolean>;
+      }
```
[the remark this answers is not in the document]
jgunnink: This is expected. The goal is to respond as quickly as possible, but still to make the call in the background anyway, which will update the cache if necessary.

libs/providers/rocketflag/src/lib/rocketflag-provider.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      expect(initialDetails.value).toBe(false);
+
+      await new Promise((resolve) => setTimeout(resolve, 0));
+
+      const finalDetails = provider.resolveBooleanEvaluation(flagKey, false, targetingContext, mockLogger);
+
+      expect(finalDetails.value).toBe(true);
+      expect(finalDetails.reason).toBe(StandardResolutionReasons.TARGETING_MATCH);
+      expect(mockClient.getFlag).toHaveBeenCalledWith(flagKey, { cohort: 'user@example.com' });
+      expect(mockClient.getFlag).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
jgunnink: As discussed above in the implementation, this is expected.

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+    metadata: {
+      name: 'RocketFlagProvider',
+    },
+    runsOn: 'client' as const,
+    hooks: [],
+
+    initialize: async (_context: EvaluationContext, initLogger?: Logger): Promise<void> => {
+      logger = initLogger;
+      logger?.debug('Initialising RocketFlagProvider...');
+    },
```
beeme1mr, line 88: Is it possible in Rocketflag to bulk evaluate or fetch a copy of the ruleset for local evaluation? The current implementation would likely work fine in React where the components would rerender once the flag has finished evaluating but the experience wouldn't be great if the user is using the web SDK directly

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (targetingKey && typeof targetingKey === 'string' && targetingKey !== '') {
+    userContext.cohort = targetingKey;
+  }
+
+  client
+    .getFlag(flagKey, userContext)
+    .then((flagStatus) => {
+      const details: ResolutionDetails<boolean> = {
+        value: flagStatus.enabled,
+        reason: userContext.cohort ? StandardResolutionReasons.TARGETING_MATCH : StandardResolutionReasons.DEFAULT,
```
beeme1mr, line 45: `TARGETING_MATCH` should indicate why a value was evaluated the way it was. In this case, it doesn't look like you'll have enough information in the response to tell.

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+    })
+    .catch((error: unknown) => {
+      const err = error instanceof Error ? error : new Error(String(error));
+      const details: ResolutionDetails<boolean> = {
+        value: defaultValue,
+        reason: StandardResolutionReasons.ERROR,
+        errorCode: ErrorCode.GENERAL,
+        errorMessage: err.message,
+      };
+      cache.set(cacheKey, details);
```
beeme1mr, line 59: It looks like errors will be cached indefinitely. Is that okay for your use case?

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+      emitter.emit(ProviderEvents.ConfigurationChanged, { flagsChanged: [flagKey] });
```
beeme1mr, line 60: You may also want to consider emitting an error event if you think the content would be useful to consumers.

libs/providers/rocketflag/README.md
beeme1mr, line 1: Please add some basic usage information, include a link to relevant Rocketflag pages, and mention limitations like that the provider only supports booleans. It doesn't have to be too elaborate but we typically direct folks to the readme from the OpenFeature Ecosystem page.

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
jgunnink: No, RocketFlag doesn't have a bulk evaluate api that you're referring to.

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
jgunnink: I'm not sure I understand, but I will try to answer.

If the user provides a cohort value in the context, then the resolution reason will be a result of targeting match. As to _why_ the API responded that way, that information is not made available in the API response.

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
jgunnink: Thanks, I'll consider it down the line if I get feedback about it. For now, I am ok with caching the details of the error as the cached value (errorCode and errorMessage).

libs/providers/rocketflag/README.md
jgunnink: Good to know thank you. I will update the readme soon.

libs/providers/rocketflag/src/lib/rocketflag-provider.ts
```diff
@@ -0,0 +1,141 @@
...
+      flagKey: string,
+      defaultValue: boolean,
+      context: EvaluationContext,
+      evalLogger: Logger,
+    ): ResolutionDetails<boolean> => {
+      logger = evalLogger; // Capture the logger for async operations.
+      const cacheKey = JSON.stringify({ flagKey, context });
+
+      // Fetch in the background.
+      fetchFlagAndUpdateCache(flagKey, defaultValue, context, cacheKey, client, cache, emitter, logger);
```
lukas-reining, line 100: Depending on how the app is structured, this could lead to a large number of requests. 
The React and Angular SDK e.g. will do many requests as e.g the Angular SDK has a directive that conditionally renders based on the flag. If this is several times in the same templates or e.g. because a view is composed of several components that each change based on the flag, for each of them a request will be done.  For React, if hooks are used in several sub components, this would accidentally trigger many requests too. 
This can be fixed by only evaluating in the root of the view but it would be better if users did not have to worry too much. 

Maybe you could assign something like a TTL to resolved values and only fetch on "non-existence" or stale flags. 
But this would be up to you to decide.

libs/providers/rocketflag/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@openfeature/rocketflag-provider",
```
lukas-reining, line 2: If you plan to add a Node provider, it would be good to include in the name that this is meant for web. 
The conventions tends to be suffixing the name with `-web`.

PR #1361, opened 2025-08-01
feat(ofrep): adding error details for 401, 403 and 429

thomaspoignant:
## This PR
Current implementation does not include any error details when failing in 401, 403 or 429.


thomaspoignant:
Good catch @beeme1mr I've move it to the exception directly.

lukas-reining (approved):

PR #1372, opened 2025-09-07
feat: update otel hook to use new semconv

lukas-reining:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523

PR #1389, opened 2025-10-07
feat: add debounce hook

toddbaert:
Adds "debounce" hook.

This is a utility "meta" hook, which can be used to effectively debounce or rate limit other hooks based on various parameters.
This can be especially useful for certain UI frameworks and SDKs that frequently re-render and re-evaluate flags (React, Angular, etc).

The hook maintains a simple LRU cache and keeps a record of recent evaluations based on a user-defined key-generation function (keySupplier).

Simply wrap your hook with the debounce hook by passing it a constructor arg, and then configure the remaining options.
In the example below, we wrap a logging hook so that it only logs a maximum of once a minute for each flag key, no matter how many times that flag is evaluated.

```ts
// a function defining the key for the hook stage 
const supplier = (flagKey: string, context: EvaluationContext, details: EvaluationDetails<T>) => flagKey;

const hook = new DebounceHook<string>(loggingHook, {
  afterCacheKeySupplier: supplier,  // if the key calculated by the supplier exists in the cache, the wrapped hook's stage will not run
  ttlMs: 60_000,                    // how long to cache keys for
  maxCacheItems: 100,               // max amount of items to keep in the LRU cache
  cacheErrors: false                // whether or not to cache the errors thrown by hook stages
});
```


libs/hooks/debounce/project.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "debounce",
+  "$schema": "../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "hooks/src",
```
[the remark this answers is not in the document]
toddbaert: Same as https://github.com/open-feature/js-sdk-contrib/pull/1389#discussion_r2411738516

libs/hooks/debounce/src/lib/utils/simple-lru-cache.ts
```diff
@@ -0,0 +1,74 @@
...
+  set(key: string, value: T) {
+    if (key) {
+      if (this.cacheMap.size >= this.maxItems) {
+        this.evictOldest();
+      }
+      this.cacheMap.set(key, {
+        value,
+      });
+    }
+  }
```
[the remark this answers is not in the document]
toddbaert: This is true - maps are kept in insertion order, but to update the order apparently `set` doesn't do it! I need to delete and add again to reset the order instead of just calling set.

I will fix this and add tests.

tsconfig.base.json
```diff
@@ -22,6 +22,7 @@
       "@openfeature/config-cat-core": ["libs/shared/config-cat-core/src/index.ts"],
       "@openfeature/config-cat-provider": ["libs/providers/config-cat/src/index.ts"],
       "@openfeature/config-cat-web-provider": ["libs/providers/config-cat-web/src/index.ts"],
+      "@openfeature/debounce-hook": ["hooks/src/index.ts"],
```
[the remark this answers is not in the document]
toddbaert: I guess if both these are changed it shouldn't matter

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  private maybeSkipAndCache(
+    stage: 'before' | 'after' | 'error' | 'finally',
+    keyGenCallback: () => string | null | undefined,
+    hookCallback: () => void,
+  ) {
+    // the cache key is a concatenation of the result of calling keyGenCallback and the stage
+    const dynamicKey = keyGenCallback();
+    const cacheKeySuffix = stage;
+    const cacheKey = dynamicKey + cacheKeySuffix;
```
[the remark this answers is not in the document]
toddbaert: 🤔 Ok I'll add something with more entropy than a colon then.

tsconfig.base.json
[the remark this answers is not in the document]
toddbaert: The publishing seems fine. Looks like NX just changed how it generates these.

I wonder if I should update all these paths now just for consistency with new generated code? 

WDYT @beeme1mr ?

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,182 @@
...
+   */
+  finallyCacheKeySupplier?: (
+    flagKey: string,
+    context: EvaluationContext,
+    details: EvaluationDetails<T>,
+  ) => string | null | undefined;
+  /**
+   * Whether or not to debounce and cache the errors thrown by hook stages.
+   * If false (default) stages that throw will not be debounced and their errors not cached.
+   */
```
guidobrei, line 87: Also clarify what happens if set to `true`
```typescript
// * If true, stages that throw will be debounced and every subsequent call during the debounce time will throw the cached error.
```

guidobrei (approved):
Nice job. Only 2 minor comments.

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,190 @@
...
+            // cache error
+            this.cache.set(cacheKey, new CachedError(error));
+          }
+          throw error;
+        }
+        return;
+      }
+    } else {
+      hookCallback.call(this.innerHook);
+    }
```
lukas-reining, line 188:
```suggestion
    const dynamicKey = keyGenCallback();

    // if the keyGenCallback returns nothing, we don't do any caching 
    if (!dynamicKey) {
      hookCallback.call(this.innerHook);
    }
    
    const cacheKeySuffix = stage;
    const cacheKey = `${dynamicKey}::${cacheKeySuffix}`;
    const got = this.cache.get(cacheKey);

    if (got) {
      // throw cached errors
      if (got instanceof CachedError) {
        throw got;
      }
      return;
    } 
    
    try {
      hookCallback.call(this.innerHook);
      this.cache.set(cacheKey, true);
    } catch (error: unknown) {
      if (this.cacheErrors) {
        // cache error
        this.cache.set(cacheKey, new CachedError(error));
      }
      throw error;
    }    
```

This was a bit hard to read to me. Maybe we can flatten it like suggested?

libs/hooks/debounce/src/lib/utils/fixed-size-expiring-cache.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+    this.maxItems = options.maxItems;
+    if (options.ttlMs < 1) {
+      throw new Error('ttlMs must be a positive integer');
+    }
+    this.ttl = options.ttlMs;
+  }
+
+  /**
+   * Gets a key from the cache, updating its recency.
```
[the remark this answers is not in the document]
beeme1mr: Did you push this update?

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,190 @@
...
+  ) {
+    this.cacheErrors = options.cacheErrors || false;
+    this.cache = new FixedSizeExpiringCache<true | CachedError>({
+      maxItems: options.maxCacheItems,
+      ttlMs: options.debounceTime,
+    });
+  }
+
+  before(hookContext: HookContext, hookHints?: HookHints) {
+    this.maybeSkipAndCache(
```
lukas-reining, line 122: I think we need to consider `hookContext.hookData` here. As the stages are cached individually, the hook data would be missing if `before` is evaluated from cache, while is`finally` relies on it and is not cached yet.
The new OTEL Hooks in #1372 (which should even work in the web-sdk now, which we might not want to debounce but nevermind :D) rely on the Hook data, and an example like: https://github.com/open-feature/js-sdk/blob/main/packages/web/test/hooks-data.spec.ts#L45 would suffer from it. 

I think that would especially apply in cases where we have thrashing on the cache as then there could be very large gaps in TTL between the stages resulting in a case where `before` is cached and `after` is not, calling `after` with empty hook data.

I see two options:
1. Caching them all together instead of individually could fix this? Then we would not care about hook data in debounced cases. 
3. We could cache the hook data and see if the former stage has cached hook data. But I think this could become problematic as there might not only be one `hook data` set because of different combinations of hook data from (non-)cached `before`, `error` and `after` stages that can all contribute to the hook data observed in the `finally` stage.

In the end, I think I would prefer to option 1, as I think the set of stages should always be the same when cached, which it mostly "nearly" is due to similar (not same) TTLs if there are no evictions. 

Does that make sense or am I getting it wrong @toddbaert?

libs/hooks/debounce/src/lib/utils/fixed-size-expiring-cache.ts
```diff
@@ -0,0 +1,94 @@
...
+   *
+   * @param key key for the entry
+   * @returns value or key or undefined
+   */
+  get(key: string): T | void {
+    if (key) {
+      const entry = this.cacheMap.get(key);
+      if (entry) {
```
beeme1mr, line 51: Could a valid value ever be an empty string? I wonder if it would be slightly safer to check for undefined.

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,190 @@
...
+    );
+  }
+
+  private maybeSkipAndCache(
+    stage: 'before' | 'after' | 'error' | 'finally',
+    keyGenCallback: () => string | null | undefined,
+    hookCallback: () => void,
+  ) {
+    // the cache key is a concatenation of the result of calling keyGenCallback and the stage
+    const dynamicKey = keyGenCallback();
```
beeme1mr, line 159: Do we need to be defensive here since users can supply their own function?

lukas-reining (commented):
This looks good, I left some nits you can ignore. 
I have question regarding hook data that I would like to clarify before approving :)

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,190 @@
...
+  get innerError() {
+    return this._innerError;
+  }
+}
+
+export type Options<T extends FlagValue = FlagValue> = {
+  /**
+   * Function to generate the cache key for the before stage of the wrapped hook.
+   * If the cache key is found in the cache, the hook stage will not run.
+   * If not defined, the DebounceHook will no-op for this stage (inner hook will always run for this stage).
```
beeme1mr, line 36: Why don't we default to using the 'flagKey'? It seems like that would be the most common use case. It looks like the hook doesn't do anything if you don't set the `beforeCacheKeySupplier`, but it's optional. I think we should make it either required or set a default. My preference would be to set a default.

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,190 @@
...
+    keyGenCallback: () => string | null | undefined,
+    hookCallback: () => void,
+  ) {
+    // the cache key is a concatenation of the result of calling keyGenCallback and the stage
+    const dynamicKey = keyGenCallback();
+
+    // if the keyGenCallback returns nothing, we don't do any caching
+    if (dynamicKey) {
+      const cacheKeySuffix = stage;
+      const cacheKey = `${dynamicKey}::${cacheKeySuffix}`;
```
beeme1mr, line 164: I guess the colons aren't necessary with the current list of stages. However, it's probably worth keeping it still.

libs/hooks/debounce/src/lib/debounce-hook.ts
```diff
@@ -0,0 +1,190 @@
...
+      const got = this.cache.get(cacheKey);
+
+      if (got) {
+        // throw cached errors
+        if (got instanceof CachedError) {
+          throw got;
```
beeme1mr, line 170: Won't this throw the wrapped error? Should we throw the underlying error instead?

libs/hooks/debounce/README.md
beeme1mr, line 1: If we want to target the web only, please make that very clear at the top of the file.

We could consider making this work on the server and client, like @lukas-reining did for the telemetry hooks. Since this hook is so flexible, I could see some use cases on the server, specifically when running it with a logging hook.

libs/hooks/debounce/README.md
```diff
@@ -0,0 +1,48 @@
...
+// a function defining the key for the hook stage; if this matches a recent key, the hook execution for this stage will be bypassed
+const supplier = (flagKey: string, context: EvaluationContext, details: EvaluationDetails<T>) => flagKey;
+
+const hook = new DebounceHook<string>(loggingHook, {
+  debounceTime: 60_000,             // how long to wait before the hook can fire again (applied to each stage independently) in milliseconds
+  afterCacheKeySupplier: supplier,  // if the key calculated by the supplier exists in the cache, the wrapped hook's stage will not run
+  maxCacheItems: 100,               // max amount of items to keep in the cache; if exceeded, the oldest item is dropped
+  cacheErrors: false                // whether or not to debounce errors thrown by hook stages
+});
+```
```
beeme1mr, line 38: Consider extending the example to show the hook being registered with OpenFeature.

libs/hooks/debounce/src/lib/debounce-hook.ts
toddbaert: Good point.

I think I also agree with 1...

@beeme1mr

libs/hooks/debounce/src/lib/debounce-hook.ts
toddbaert: [This comment](https://github.com/open-feature/js-sdk-contrib/pull/1389#discussion_r2414225730) has implications here.

Issue #1096, opened 2024-10-31
Update the flagd provider to conform to the latest spec


toddbaert:
Update the flagd provider to conform to the latest spec: https://flagd.dev/reference/specifications/providers/

For the Java provider, that means:

- [ ] implement `retryGraceAttempts` param:
    - defaults to 5
    - when provider disconnects, if reconnect attempts <= `retryGraceAttempts`, provider emits `STALE`
    - when reconnect attempts > `retryGraceAttempts` provider is emits `ERROR`
- [ ] implement metadata-fetch with in-process init:
    - call <https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1#flagd.sync.v1.FlagSyncService.GetMetadata> simultaneously with connecting the sync stream
        - fail init if metadata call fails
    - add `contextEnricher` function param:
        - converts payload from sync-metadata to Evaluation context on init
        - stores the result from above and adds it with every evaluation, probably using a hook (Java implementation: <https://github.com/open-feature/java-sdk-contrib/blob/4b8aa8559fea5b995b10902eeabf1e9d4876b7bf/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdProvider.java#L72C23-L72C39>)
- [ ] implement gRPC (HTTP2) `keepalive` param
    - add keepalive option to gRPC connections
- [ ] implement `streamDeadlineMs` (defaults to 600000 millis)
    - only applies to event stream (RPC) and sync stream (in-process)
- [ ] review and align all configuration params as specified in configuration: https://flagd.dev/reference/specifications/providers/#configuration

PR #1430, opened 2025-12-18
feat: Add  FLAGD_STREAM_DEADLINE_MS parameter

marcozabel:
## This PR

Is adding the stream deadline parameter to the gRPC channel.

### Related Issues

Parent: https://github.com/open-feature/js-sdk-contrib/issues/1096
Sub: https://github.com/open-feature/js-sdk-contrib/issues/1287

### DIFF
--- a/libs/providers/flagd/src/lib/service/grpc/grpc-service.ts
+++ b/libs/providers/flagd/src/lib/service/grpc/grpc-service.ts
@@ -164,7 +164,8 @@ export class GRPCService implements Service {
     // close the previous stream if we're reconnecting
     closeStreamIfDefined(this._eventStream);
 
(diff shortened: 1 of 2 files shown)


toddbaert (approved):
Thanks @marcozabel !

I made one small but important change. It's slightly confusing, but deadlines in gRPC are not time-deltas - they are timestamps. In other words, every time you set a deadline, you have to specify the exact time in the future the request will expire. Just setting the deadline as a const like `60000` means the request expires at `Wed Dec 31 1969 19:01:00 GMT-0500 (Eastern Standard Time)` (so all our e2e tests failed). I fixed this very simply here: https://github.com/open-feature/js-sdk-contrib/pull/1430/changes/264ed646fbb65a00f52bf963d777a9a0ec4b02a6

Issue #1424, opened 2025-12-11
[flagd] add retry policy, and make configurable


toddbaert:
Add a configurable retry backoff options to the flagd in-process provider, which allows to configure the retry policy of gRPC connections. 

See equivalent implementation [here](https://github.com/open-feature/go-sdk-contrib/pull/806) (use the same variables and env-var names)

- grpc-js retry policy example: https://github.com/grpc/grpc-node/blob/164d14f28261ad1fbbc99d076fe850dad018b2bb/examples/retry/README.md?plain=1#L29-L56
- use the same policy as seen here: https://github.com/open-feature/go-sdk-contrib/pull/806/files#diff-431e1694cab09598ed0a65d71eb6f14d0352c3a208ed215ce03f01711bdde2b8R24-R45

marcozabel:
> [@marcozabel](https://github.com/marcozabel) the mentioned PR is not related to this issue

Yeah, I mistakenly referenced it and it did not disappear in this thread after fixing it.

PR #1437, opened 2025-12-22
feat: Add retry backoff to grpc connections

marcozabel:
## This PR
Is adding a configuration for the backOff timing during reconnection in Grpc Providers.

### Related Issues
Implements: https://github.com/open-feature/js-sdk-contrib/issues/1424

### DIFF
--- a/libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
+++ b/libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
@@ -4,7 +4,7 @@ import { GeneralError } from '@openfeature/server-sdk';
 import type { SyncFlagsRequest, SyncFlagsResponse } from '../../../../proto/ts/flagd/sync/v1/sync';
 import { FlagSyncServiceClient } from '../../../../proto/ts/flagd/sync/v1/sync';
 import type { Config } from '../../../configuration';
-import { buildClientOptions, closeStreamIfDefined, createChannelCredentials} from '../../common';
+import { buildClientOptions, closeStreamIfDefined, createChannelCredentials } from '../../common';
 import type { DataFetch } from '../data-fetch';
 
 /**


libs/providers/flagd/src/lib/service/common/grpc-util.ts
```diff
@@ -31,3 +31,37 @@ export const createChannelCredentials = (tls: boolean, certPath?: string): Chann
...
+  return {
+    'grpc.service_config': JSON.stringify({
+      loadBalancingConfig: [],
+      methodConfig: [
+        {
+          name: [{ service: serviceName }],
+          retryPolicy: {
+            maxAttempts: 3,
+            initialBackoff: `${Math.round(initialBackoff / 1000)}s`,
+            maxBackoff: `${Math.round(maxBackoff / 1000)}s`,
```
toddbaert:
```suggestion
            initialBackoff: `${Math.round(initialBackoff / 1000).toFixed(2)}s`,
            maxBackoff: `${Math.round(maxBackoff / 1000).toFixed(2)}s`,
```

I think we should improve the precision here to support sub-seconds.

libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.ts
toddbaert: This is an existing bug, but the options are totally reset here if this one is included. We may want to avoid using a `let` and instead use a `const` to prevent this sort of bug.

In any case, lets fix this so that the the retry policy is used regardless of if the defaultAuthority is set or not.

toddbaert:
In addition to the minor points above, we need to add the policy to the RPC mode as well - client [here](https://github.com/open-feature/js-sdk-contrib/blob/592fdff7a05c9b31f0f4df23176e4a2799ea838d/libs/providers/flagd/src/lib/service/grpc/grpc-service.ts#L167) .

PR #1464, opened 2026-02-03
feat: An option to add the User-Agent header (GoFeatureFlag)

olvlvl:
## This PR
(sorry, I cannot edit the title: User-Agent -> Custom Headers)

Adds support for custom headers. One could use the feature to define a `User-Agent` to prevent requests from being blocked by servers that require browser-like identification or specific device signatures.

### Notes
Our WAF requires the `User-Agent` header to be defined. Requests to the relay-proxy are being denied because the header is not specified.


olvlvl:
Replaced by: https://github.com/open-feature/js-sdk-contrib/pull/1465

PR #1465, opened 2026-02-04
feat: Add optional custom headers (GoFeatureFlag)

olvlvl:
## This PR
Adds support for custom headers. One could use the feature to define a `User-Agent` to prevent requests from being blocked by servers that require browser-like identification or specific device signatures.

### Notes
Our WAF requires the `User-Agent` header to be defined. Requests to the relay-proxy are being denied because the header is not specified.


thomaspoignant (approved):
🚀 LGTM

PR #1475, opened 2026-02-13
chore(go-feature-flag): Update wasm_exec to upgrade tinygo build version + bump ofrep dependency

thomaspoignant:
## This PR
New `wasm_exec.js` that follow the new version of tinygo we are using to compile the evaluation logic.


thomaspoignant:
Tests are failing because we need a new version of the WASM module.

Issue #1480, opened 2026-02-24
feat(flagd-core): add WebWorker / V8 isolate compatibility mode


jonathannorris:
### Problem

`@openfeature/flagd-core` cannot run in Cloudflare Workers or other V8 isolate-based environments (Deno Deploy, Vercel Edge Runtime, etc.) because two dependencies use dynamic code generation, which these runtimes explicitly block:

| Dependency | Usage | Failure |
|---|---|---|
| `ajv` | JSON Schema validation | `new Function()` called at module load time |
| `json-logic-engine` | Targeting rule evaluation | `eval()` used by `.build()` compilation mode |

The `ajv` issue fails at **import time** — a Worker crashes before any flag evaluation occurs.

### Proposed Solution

Add an opt-in `workers?: boolean` option to `FlagdCore` that:

1. Switches `json-logic-engine` from compiled mode (`.build()`) to interpreter mode (`.run()`)
2. Uses pre-compiled ajv validators generated at build time via `ajv/dist/standalone`, instead of compiling schemas at runtime

Defaults to `false` for full backwards compatibility. The interpreted JSONLogic path is ~10-20x slower per evaluation but remains well within acceptable latency for typical use (a few evaluations per request, in the microsecond range).

### Required Changes (`libs/shared/flagd-core/`)

- Add `FlagdCoreOptions` interface with `workers?: boolean` and export it from the package public API
- Accept `options` in `FlagdCore` constructor and thread through the evaluation pipeline
- Switch `json-logic-engine` to interpreter mode when `workers: true`
- Avoid top-level ajv schema compilation; use pre-compiled validators when `workers: true`
- Add a build script to generate static validators using `ajv/dist/standalone` and wire into the build/release workflow
- Add compatibility tests confirming correct evaluation in worker mode (targeting rules, fractional, semver, string operators)

### Reference

A working prototype lives at [`DevCycleHQ-Sandbox/js-sdk-contrib`](https://github.com/DevCycleHQ-Sandbox/js-sdk-contrib) on branch `feat/workers-compatibility`, validated end-to-end in a Cloudflare Worker exposing the OFREP API.

PR #1481, opened 2026-02-24
feat(flagd-core): pre-compile AJV validators for edge runtime compatibility (1/2)

jonathannorris:
## Summary

Part 1 of 2 for [#1480](https://github.com/open-feature/js-sdk-contrib/issues/1480) — edge function runtime compatibility for `@openfeature/flagd-core`.

Pre-compiles AJV validators at build time so that **no dynamic code generation** (`new Function()`) is ever needed for schema validation at runtime. This makes schema validation compatible with edge function runtimes that restrict dynamic code evaluation.

### Changes

- **`scripts/build-validators.js`** — new build script that generates CJS pre-compiled validators from the flagd JSON schema using `ajv`'s standalone output mode
- **`src/lib/generated/validators.js`** — generated pre-compiled validator (checked in; regenerated by `npm run build:validators`)
- **`src/lib/generated/validators.d.ts`** — TypeScript declarations for the generated module (inlined `ValidateFunction` type, no `ajv` import needed by consumers)
- **`src/lib/parser.ts`** — switched to pre-compiled validator; removed runtime `new Ajv()` / `ajv.compile()` entirely
- **`package.json`** — `ajv` moved from `dependencies` → `devDependencies` (no longer needed at runtime); `build:validators` script added
- **`project.json`** — `buildValidators` Nx target added and wired as a dependency of the `package` target
- **`.gitattributes`** — marks `validators.js` as a generated file for GitHub
- **`.eslintrc.json`** — excludes `src/lib/generated` from linting and `scripts/**` / `src/lib/generated/**` from `@nx/dependency-checks`

### Why pre-compile instead of a runtime option?

Pre-compiled validators avoid the `new Function()` startup cost, so they tend to be at least as fast as runtime compilation in standard environments — and unlike runtime compilation, they work in edge function runtimes that restrict dynamic code evaluation. Since there's no meaningful downside to pre-compiling here, this PR removes runtime AJV compilation entirely rather than gating it behind an option.

### Testing

- 81/81 tests pass (no test changes in this PR — restricted environment tests are in PR 2)
- Lint clean (`eslint --max-warnings=0`)

### Follow-up

PR 2/2 (#1482) adds `FlagdCoreOptions.disableDynamicCodeGeneration` to switch `json-logic-engine` to interpreter mode (`.run()` instead of `.build()`), completing edge runtime compatibility for targeting rules. It also adds the restricted Jest environment test infrastructure.

### DIFF
--- a/libs/shared/flagd-core/project.json
+++ b/libs/shared/flagd-core/project.json
@@ -23,10 +23,7 @@
     "buildValidators": {
       "executor": "nx:run-commands",
       "options": {
-        "commands": [
-          "git submodule update --init flagd-schemas",
-          "node scripts/build-validators.js"
-        ],
+        "commands": ["git submodule update --init flagd-schemas", "node scripts/build-validators.js"],
         "cwd": "libs/shared/flagd-core",
         "parallel": false
       },


toddbaert (commented):
@jonathannorris is there any reason to commit the generated stuff instead of just .gitignore-ing it?

We've generally done this with proto-files, etc.

Not a blocker for me either way, but I have a mild preference for committing less.

PR #1508, opened 2026-04-02
feat: added local persistence to ofrep-web provider

jsalaber:
## This PR

- Replaced disableLocalCache boolean option with a cacheMode string: 'local-cache-first' (default), 'network-first', or 'disabled'
- Added cacheTtl option (default 30 days) — expired cache entries are evicted on read
- Added cacheKeyPrefix option — included in the storage key hash to prevent collisions when multiple provider instances share the same origin
- Persisted cache entries now store the ETag alongside flags, so it is restored on cold start for efficient revalidation
local-cache-first: serves flags from localStorage immediately on init, then refreshes from the network in the background
- network-first: blocks init on the network request, falls back to the persistent cache only on transient errors (network, 5xx); auth errors (401/403) surface immediately and are never masked by cached values
- Auth errors during background polling now emit PROVIDER_ERROR instead of clearing the cache

### Related Issues

Implements this ADR: https://github.com/open-feature/protocol/pull/64


libs/providers/ofrep-web/tsconfig.lib.json
```diff
@@ -5,7 +5,10 @@
     "outDir": "../../../dist/out-tsc",
     "declaration": true,
     "types": [],
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "paths": {
+      "@openfeature/ofrep-core": ["../../../dist/libs/shared/ofrep-core/index.esm.js"]
```
marcozabel: q: Why is this being added?

kaushalkapasi (approved):

marcozabel (approved):

beeme1mr (changes requested):
The ADR referenced in the PR has been closed. There's a new one but it hasn't been merged yet. Let's old off merging this until that ADR has been accepted.

https://github.com/open-feature/protocol/pull/75/changes

toddbaert:
@jsalaber - looks really good to me - the only concern I have is there's no test coverage for the `network-first` option as far as I can see; please correct me if I'm wrong.

Once that's added I will approve.

(I think you also might need to run an `npm i`).

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -282,38 +340,64 @@ export class OFREPWebProvider implements Provider {
...
+    }
+  }
+
+  private async _tryLoadFlagsFromCache(context?: EvaluationContext | undefined): Promise<boolean> {
+    const cached = this._storage.retrieve(context?.targetingKey ?? '', this._cacheTtl);
+    if (cached) {
+      this._isUsingCache = true;
+      this._flagCache = cached.flags;
+      this._etag = cached.etag;
+      void this._refreshFlagsInBackground();
```
jonathannorris, line 421: ADR 0009 calls out this race directly under Implementation Notes:

> If `onContextChanged()` is called while a background refresh is still in-flight, the provider should cancel or discard the in-flight request. The context-change evaluation supersedes it and should be the authoritative write to the persisted entry.

On a cache-hit `initialize(userA)`, `_tryLoadFlagsFromCache` kicks off a background `_refreshFlagsInBackground()` for userA. If `onContextChange(userA, userB)` lands before that fetch resolves, the late userA response will still write to `this._flagCache` and `this._etag` after userB has been set, so evaluations briefly return userA's values under userB's context and the next poll sends userA's ETag on a userB request.

`onContextChange` itself also calls `_tryLoadFlagsFromCache` on cache hit, which starts a *second* concurrent background refresh, so two in-flight fetches can resolve in either order.

Can we add cancellation or a generation-token guard so only the most recent authoritative fetch is allowed to update `_flagCache` / `_etag` / storage?

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -210,10 +265,13 @@ export class OFREPWebProvider implements Provider {
 
       const listUpdatedFlags = this._getListUpdatedFlags(this._flagCache, newCache);
       this._flagCache = newCache;
-      this._etag = response.httpResponse?.headers.get('etag');
+      const newEtag = response.httpResponse?.headers.get('etag') ?? null;
+      this._storage.store(context?.targetingKey ?? '', newCache, newEtag);
+      this._etag = newEtag;
       this._flagSetMetadataCache = toFlagMetadata(
```
jonathannorris, line 292: Nit: `_flagSetMetadataCache` is only populated by `_fetchFlags`, so after a cache-hit cold start (before the first successful background refresh completes), a `FLAG_NOT_FOUND` error response will have `flagMetadata: undefined` even if the server had originally returned flag-set metadata. Per-flag metadata is preserved inside each cached `EvaluationResponse`, so the success path is fine.

ADR 0009's example persisted envelope only lists `data.flags`, so this isn't strictly a spec miss, but you might consider persisting the flag-set metadata alongside the flags so consumers get consistent metadata across cold start and refresh.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -133,28 +151,40 @@ export class OFREPWebProvider implements Provider {
    */
   async onContextChange(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
     try {
+      if (oldContext?.targetingKey !== newContext?.targetingKey) {
```
jonathannorris, line 156: Open question: should we clean up the old `targetingKey`'s persisted entry here on user switch?

ADR 0009 Implementation Notes:

> Providers should clear or replace persisted entries when the `targetingKey` changes, such as on logout or user switch.

Today the new entry is written under a different hashed storage key, but the old entry sits in `localStorage` until TTL expiry (30 days default). `Storage.clear()` exists but isn't called from the provider.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
```diff
@@ -318,4 +332,239 @@ describe('OFREPWebProvider', () => {
     await new Promise((resolve) => setTimeout(resolve, 400));
     expect(reconcilingHandler).toHaveBeenCalledTimes(2);
   });
+
+  describe('persistent local cache', () => {
```
jonathannorris, line 338: The new tests don't seem to cover `cacheMode: 'network-first'` at all. ADR 0009 specifies a specific fallback behavior that's worth a dedicated test:

> If the request fails with a transient or server error, or times out... If one exists, populate the in-memory cache from it and return from `initialize()` (SDK emits `PROVIDER_READY`). Evaluations use `CACHED` as the reason.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -223,7 +281,7 @@ export class OFREPWebProvider implements Provider {
     }
   }
 
-  /**
+    /**
```
jonathannorris: Nit: stray indentation on the JSDoc opener — `    /**` should be `  /**`. Looks like a merge artifact.

libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
```diff
@@ -1,12 +1,51 @@
...
+   *   by cached values.
+   * - `'disabled'`: no persistence at all. `initialize()` always blocks on the network.
+   *   Persistence-related options have no effect.
+   *
+   * Default: `'local-cache-first'`
+   */
+  cacheMode?: CacheMode;
+
+  /**
+   * cacheTtl is the maximum age (in milliseconds) of a persisted cache entry before it is
```
jonathannorris: should we just use seconds?

jonathannorris: also thoughts on `cacheTTL`?

libs/providers/ofrep-web/src/lib/store/storage.ts
```diff
@@ -0,0 +1,128 @@
+import type { Logger } from '@openfeature/core';
+import MurmurHash3 from 'imurmurhash';
```
jonathannorris: `imurmurhash` isn't resolvable from the published package. `libs/providers/ofrep-web/package.json` only declares `@openfeature/ofrep-core` as a runtime dep, and walking the full tree — `@openfeature/ofrep-core@2.1.0` (peer: `@openfeature/core`), `@openfeature/web-sdk@1.8.0` (peer: `@openfeature/core`), and `@openfeature/core@1.10.0` (no deps) — nothing pulls in `imurmurhash`. It works in the monorepo because the dep is hoisted from the root workspace `package.json`, but Nx's rollup config uses `"external": "all"` so the published tarball won't bundle it, and consumers who `npm install @openfeature/ofrep-web-provider` will hit a missing-module error at runtime.

Rather than declaring the dep, consider using the Web Crypto API (`crypto.subtle.digest('SHA-256', ...)`) — it's a W3C standard available in all modern browsers (secure contexts) and Node 16+ as `globalThis.crypto.subtle`, no dependency needed. Truncating to 64 bits (16 hex chars) keeps storage keys compact and is still collision-safe at this scale.

The tradeoff is that `getStorageKey` would need to become async, which cascades through `store` / `retrieve` / `clear` and their callers in the provider (which are already async, so the cascade is tolerable).

toddbaert: I believe if it wasn't failing due to the `npm ci` error, our linter CI job would have caught this (and will again once the `npm ci` issue is fixed) it forces all imported deps to appear in the package's dependencies or will fail.

I _think_ it's worth using async and browser-native hash algo here as @jonathannorris is saying - though SHA1 is probably good enough since cryptographic security isn't needed. I don't think the browser naively has any fast hashes like murmur or md5.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
toddbaert: [Agree](https://github.com/open-feature/js-sdk-contrib/pull/1508/#issuecomment-4306880900). I think this needs a test.

libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
jsalaber: i'm cool with `cacheTTL`

libs/providers/ofrep-web/src/lib/ofrep-web-provider.spec.ts
jsalaber: tests added

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
jsalaber: yep I can clear the cache for the old user

libs/providers/ofrep-web/src/lib/store/storage.ts
jsalaber: removed imurmurhash in favour of natively available `crypto`

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
jsalaber: added a `contextRevision` to track the most up to date context to discard any outdated context data response

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
jsalaber: it now persists the flagset metadata in the persistent local storage

toddbaert (approved):

jonathannorris:
Heads up — looks like the conflict resolution in c70cf03e accidentally reverted two of the previously-addressed review items:

- **Flag-set metadata persistence** ([#discussion_r3133055199](https://github.com/open-feature/js-sdk-contrib/pull/1508#discussion_r3133055199)): the prior version had `metadata` on `PersistedEntry`, written by `Storage.store()`, restored by `_initNetworkFirst` and `_tryLoadFlagsFromCache`, and a dedicated test ("restores flag-set metadata from the persisted entry so FLAG_NOT_FOUND includes it on a cold cache start"). All four are gone in the current head.
- **Clearing the old `targetingKey` entry on user switch** ([#discussion_r3133064290](https://github.com/open-feature/js-sdk-contrib/pull/1508#discussion_r3133064290)): the `void this._storage.clear(oldContext?.targetingKey ?? '')` call in `onContextChange` and the "clears the old targeting key persisted entry when the targeting key changes" test are also gone.

Everything else from the prior review pass (`_contextRevision` generation token, JSDoc indent, network-first tests, SubtleCrypto migration) is still in place. Probably just want to cherry-pick those two changes back in.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -170,6 +205,34 @@ export class OFREPWebProvider implements Provider {
...
+  /**
+   * network-first initialization: block on the network request; fall back to the persisted
+   * cache only on transient / server errors. Auth and configuration errors (401, 403) are
+   * never masked by cached values.
+   */
+  private async _initNetworkFirst(context?: EvaluationContext): Promise<void> {
+    try {
+      await this._fetchFlags(context);
+    } catch (error) {
+      // Auth/config errors surface immediately — no cache fallback.
```
jonathannorris: ADR 0009 §98:

> If the request fails with an authorization or configuration error (`401`, `403`, `400`): preserve the existing initialization failure behavior (SDK emits `PROVIDER_ERROR` with error code `PROVIDER_FATAL`). Do not fall back to cached values, even if a persisted entry exists.

And §207:

> providers must not silently fall back to persisted data for authorization failures, invalid requests, or other responses that indicate a configuration or protocol problem. In `network-first` mode this applies even when a matching persisted entry exists

The `cacheMode` JSDoc here also explicitly says "Auth and configuration errors (401, 403, 400) are surfaced immediately and never masked by cached values."

But `_initNetworkFirst` only catches `OFREPApiUnauthorizedError` (401) and `OFREPForbiddenError` (403) as no-fallback errors. A 400 response surfaces from `_fetchFlags` as `TargetingKeyMissingError`, `InvalidContextError`, `ParseError`, or `GeneralError` — none of which are caught by this branch — so the code falls through to the cache fallback and serves stale flags as `CACHED` under what is actually a configuration problem.

Suggest also rethrowing on the `OpenFeatureError` 400-class subclasses (or whatever the appropriate predicate is for "this came from a 400 response"). Worth a dedicated `network-first` test for 400 → cache-not-used.

jonathannorris:
I think we should update the README in this PR too. `OFREPWebProviderOptions` now adds `cacheMode`, `cacheTTL`, and `cacheKeyPrefix`, but the package docs don't mention them yet. I recommend adding a short caching section covering those options and the `local-cache-first` startup behavior, where initial evaluations may resolve with reason `CACHED` before the background refresh completes.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -170,6 +205,40 @@ export class OFREPWebProvider implements Provider {
...
+    try {
+      await this._fetchFlags(context);
+    } catch (error) {
+      // Auth/config errors, protocol errors, and bad request / server errors surface immediately — no cache fallback.
+      const status = error instanceof OFREPApiError ? error.response?.status : undefined;
+      if (
+        error instanceof OFREPApiUnauthorizedError ||
+        error instanceof OFREPForbiddenError ||
+        error instanceof OpenFeatureError ||
+        (status !== undefined && OFREPApi.isOFREFErrorHttpStatus(status))
```
jonathannorris: The 400 fix is right, but the new gate also catches 500 via `OFREPEvaluationErrorHttpStatuses` ([400, 404, 500]). ADR §94 puts 5xx in the "transient or server error" bucket where cache fallback should apply, and §98 only lists `401`, `403`, `400` as the no-fallback set.

This also creates an inconsistency: 500 surfaces immediately, but 502/503/504 still fall back (they come through as `OFREPApiUnexpectedResponseError`, which isn't in the gate). Two adjacent server-error statuses behave differently.

Suggest narrowing the gate to auth/config only — 400, 401, 403, 404 — and reverting the 500 test back to expecting cache fallback. The README's "transient failures (network unavailable, timeout)" line should also list 5xx.

jonathannorris (approved):

libs/providers/ofrep-web/src/lib/store/storage.ts
```diff
@@ -0,0 +1,141 @@
...
+  async retrieve(
+    targetingKey: string,
+    ttlSeconds: number,
+  ): Promise<{ flags: FlagCache; etag: string | null; metadata?: Record<string, unknown> } | undefined> {
+    if (this._disabled) return undefined;
+    try {
+      const raw = localStorage.getItem(await this.getStorageKey(targetingKey));
+      if (!raw) return undefined;
+
+      const entry = JSON.parse(raw) as PersistedEntry;
```
beeme1mr, line 110: Do we need to consider that the data in local storage may have been manually manipulated? It's probably fine to leave as is but it's probably worth briefly considering the implications.

beeme1mr (approved):

libs/providers/ofrep-web/src/lib/store/storage.ts
jsalaber: I don't think it's really feasible, since everything on the browser is pretty much available to the user, we'd need to make some sort of protocol to authenticate stored configs with the server

beeme1mr:
I'll merge this tomorrow unless someone objects. FYI @lukas-reining @thomaspoignant

lukas-reining (approved):
This looks very good to me!

PR #1509, opened 2026-04-08
feat(ofrep-web): add SSE event stream support (#1497)

marcozabel:
Implement real-time flag change notifications via Server-Sent Events per ADR-0008. The provider now parses the optional `eventStreams` array from bulk evaluation responses and connects to SSE endpoints using the native EventSource API.

Changes:
- Add EventStream types to ofrep-core (BulkEvaluationSuccessResponse)
- Add SseManager class handling EventSource lifecycle, event coalescing, and inactivity timeout via Page Visibility API
- Extend postBulkEvaluateFlags with flagConfigEtag/flagConfigLastModified query params for SSE-triggered re-fetches
- Integrate SseManager into OFREPWebProvider with automatic fallback to polling when SSE is unavailable
- Add inactivityDelaySec option to OFREPWebProviderOptions
- Add 20 new tests covering SSE connection, refetch, coalescing, inactivity timeout, error fallback, and context change handling

## This PR

- adds this new feature

### Related Issues
https://github.com/open-feature/js-sdk-contrib/issues/1497


toddbaert:
@marcozabel the CI failure is pre-existing, some dep-linting error that somehow got passed CI before. I've [fixed it in main](https://github.com/open-feature/js-sdk-contrib/commit/c035137b5ab5d098fcc1c9ec409887caa30223fa). Rebasing here should resolve it.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -57,13 +59,23 @@ export class OFREPWebProvider implements Provider {
...
     this._ofrepAPI = new OFREPApi(this._options, this._options.fetchImplementation);
     this._pollingInterval = this._options.pollInterval ?? this.DEFAULT_POLL_INTERVAL;
+    this._sseManager = new SseManager(
+      {
+        onRefetch: (metadata) => this._handleSseRefetch(metadata),
+        onError: () => this._handleSseError(),
+      },
+      this._options.inactivityDelaySec,
+      this._logger,
+    );
```
toddbaert: Minor: could the `SseManager` be instantiated lazily (e.g. on first `_connectSseIfAvailable` call that actually has `eventStreams`)? That way users who never get SSE URLs from their server don't have an unused manager sitting around.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -180,15 +200,17 @@ export class OFREPWebProvider implements Provider {
...
+    context?: EvaluationContext | undefined,
+    sseMetadata?: SseRefetchMetadata,
+  ): Promise<EvaluateFlagsResponse> {
     try {
       const evalReq: EvaluationRequest = {
         context,
       };
 
-      const response = await this._ofrepAPI.postBulkEvaluateFlags(evalReq, this._etag);
+      const response = await this._ofrepAPI.postBulkEvaluateFlags(evalReq, this._etag, sseMetadata);
```
toddbaert: On inactivity resume, `SseManager` calls `onRefetch(undefined)` which correctly omits SSE metadata, but `_fetchFlags` still sends `this._etag` as `If-None-Match`. [ADR-0008 guideline #3](https://github.com/open-feature/protocol/blob/ee7c27b485136d80d59b7ad965e87c6e006819bb/service/adrs/0008-sse-for-bulk-evaluation-changes.md?plain=1#L147) requires the inactivity resume re-fetch to be fully unconditional; without `If-None-Match`, `flagConfigEtag`, or `flagConfigLastModified`. I think this is a deviation, IIUC.

toddbaert: I think we might want a specific test for this, if possible, since it was explicitly specified but it's a fine detail.

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,247 @@
...
+    const serverInactivity = sseStreams.find((s) => s.inactivityDelaySec !== undefined)?.inactivityDelaySec;
+    this._inactivityDelaySec = this._clientInactivityOverride ?? serverInactivity ?? DEFAULT_INACTIVITY_DELAY_SEC;
+
+    const urls = new Set<string>();
+    for (const stream of sseStreams) {
+      const url = resolveUrl(stream);
+      if (url) {
+        urls.add(url);
+      }
+    }
```
toddbaert: I'd prefer a reducer, but that's a nit.

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,247 @@
...
+    this.disconnect();
+
+    const sseStreams = eventStreams.filter((s) => s.type === 'sse');
+    if (sseStreams.length === 0) {
+      return;
+    }
+
+    // Determine effective inactivity delay:
+    // 1. Client override > 2. Server value > 3. Default (120s)
+    const serverInactivity = sseStreams.find((s) => s.inactivityDelaySec !== undefined)?.inactivityDelaySec;
```
toddbaert: Nit: since `inactivityDelaySec` comes from a parsed JSON response, it could be `null` at runtime. `!= null` instead of `!== undefined` would be a bit more defensive.

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    if (urls.size === 0) {
+      return;
+    }
+
+    this._urls = urls;
+    for (const url of urls) {
+      this._connectToUrl(url);
+    }
```
toddbaert: Another nit: use .map :D

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    const es = new EventSource(url);
+
+    es.addEventListener('message', (event: MessageEvent) => {
+      this._handleMessage(event);
+    });
+
+    es.addEventListener('error', () => {
+      this._logger?.warn('SSE connection error for URL: ' + url);
```
toddbaert: The ADR notes that SSE URLs may contain auth tokens or channel credentials and must not be logged. This line logs the full URL, which could leak sensitive material. Maybe log a generic message without the URL, or redact the query string?

toddbaert (changes requested):
I left a few comments. I think there's a couple deviations from the ADR, but definitely on the right track!

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -57,13 +59,23 @@ export class OFREPWebProvider implements Provider {
   private _flagSetMetadataCache?: MetadataCache;
   private _context: EvaluationContext | undefined;
   private _pollingIntervalId?: number;
+  private _sseManager: SseManager;
+  private _sseActive = false;
```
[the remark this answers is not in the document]
jonathannorris: I don't see much semantic difference between `_sseEnabled` and `_sseActive`, but I also wonder if the existance of an `_sseManager` if we make it optional could imply the same thing.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -286,6 +312,55 @@ export class OFREPWebProvider implements Provider {
     return toResolutionDetails(response, defaultValue);
   }
 
+  /**
+   * Connect to SSE event streams if present in the response.
+   * If SSE connects, stop polling; if no SSE, does not start polling
+   * (that is handled by the caller).
+   */
+  private _connectSseIfAvailable(result: EvaluateFlagsResponse): void {
```
jonathannorris, line 451: nitpick: but should we use a capitalized `SSE` for all these functions?

toddbaert: The major style guides I'm familiar with would recommend `sseConnection` - (ie: treat the acronym as a word with only the first char capitalized)

- https://google.github.io/styleguide/tsguide.html
- https://github.com/airbnb/javascript#naming--camelCase

That said I don't really care.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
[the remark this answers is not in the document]
marcozabel: I removed the SseActive attribute because the existence of the SseManager did imply this. 👍

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
marcozabel: I added two tests 👍

libs/providers/ofrep-web/src/lib/sse-manager.ts
marcozabel: I removed the URL 👍

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
marcozabel: I'm in favor of treating it as a name 😄

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,247 @@
...
+      // EventSource auto-reconnects; if all connections fail,
+      // the provider-level error callback triggers polling fallback.
+      this._callbacks.onError();
+    });
```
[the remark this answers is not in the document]
marcozabel: I drafted a a grace period. Let me know what you think 👍

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -137,11 +142,15 @@ export class OFREPWebProvider implements Provider {
...
       if (this._retryPollingAfter !== undefined && this._retryPollingAfter > now) {
-        // we do nothing because we should not call the endpoint
         return;
       }
 
-      await this._fetchFlags(newContext);
+      // Context change: re-fetch without SSE metadata
+      const result = await this._fetchFlags(newContext);
+      this._connectSseIfAvailable(result);
+      if (!this._sseManager && this._pollingInterval > 0 && !this._pollingIntervalId) {
```
chrfwow: what is the `_pollingIntervalId`?

marcozabel: When creating the polling, we use `setInterval` to periodically make requests. This method returns the id of the interval for later reference and we store it in `_pollingIntervalId`.

toddbaert (approved):
All my concerns have been addressed.

libs/shared/ofrep-core/src/lib/model/event-stream.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * A structured endpoint for SSE connections.
+ * Providers construct the URL as `origin + requestUri`.
+ */
+export interface EventStreamEndpoint {
+  origin: string;
```
jonathannorris: Per ADR-0008 / `openapi.yaml`, `endpoint.origin` is optional — when absent, providers should fall back to the configured OFREP `baseUrl` origin.

Two issues:
1. `origin` is typed as required here.
2. In `sse-manager.ts`, `new URL(requestUri, stream.endpoint.origin)` will throw when `origin` is undefined, and `SseManager` has no `baseUrl` to fall back to.

Suggest making `origin?: string` and passing `baseUrl` into `SseManager` for the fallback.

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -173,10 +173,19 @@ export class OFREPApi {
   public async postBulkEvaluateFlags(
     requestBody?: EvaluationRequest,
     etag: string | null = null,
+    sseMetadata?: { flagConfigEtag?: string; flagConfigLastModified?: string | number },
```
jonathannorris: Minor: consider exporting a named type (e.g. `BulkEvaluationRefetchMetadata`) from `ofrep-core` instead of an inline structural type here. The equivalent `SseRefetchMetadata` lives in `ofrep-web`, which means any future non-web OFREP provider would duplicate it. Since this is the shared package's public API, the type belongs alongside `EventStream`.

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
```diff
@@ -173,10 +173,19 @@ export class OFREPApi {
...
+    const params = new URLSearchParams(this.baseOptions.query);
+    if (sseMetadata?.flagConfigEtag) {
+      params.set('flagConfigEtag', sseMetadata.flagConfigEtag);
+    }
+    if (sseMetadata?.flagConfigLastModified !== undefined) {
+      params.set('flagConfigLastModified', String(sseMetadata.flagConfigLastModified));
+    }
+    const qs = params.toString();
+    if (qs) {
+      url = url + `?${qs}`;
```
jonathannorris, line 189: No test currently asserts that `flagConfigEtag` / `flagConfigLastModified` actually appear on the outgoing request URL — the existing tests stop at the callback/argument layer. Since these query params are the spec conformance surface (ADR-0008 §SSE metadata transport), worth adding a test in `ofrep-api.spec.ts` that intercepts the fetch and asserts the URL.

Also worth covering that pre-existing `baseOptions.query` params are still preserved after the switch to `URLSearchParams` here.

libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
```diff
@@ -9,4 +9,11 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
...
    */
   pollInterval?: number; // in milliseconds
+
+  /**
+   * Client-side override for the SSE inactivity timeout in seconds.
+   * When set, this takes precedence over the server-provided `inactivityDelaySec`.
+   * If neither is set, defaults to 120 seconds.
+   */
+  inactivityDelaySec?: number;
 };
```
jonathannorris, line 76: ADR-0008 Implementation Notes recommend a `changeDetection` option with values `sse` (default) / `polling` / `none`:

- `polling` ignores `eventStreams` entirely
- `none` disables all background refresh (only `onContextChange` triggers refetch)

Currently there's no way to force polling when a server returns `eventStreams`, and no way to disable background refresh. Worth adding, or tracking as a follow-up so other OFREP providers (Java, Swift, etc.) align on the same option name.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -286,6 +307,75 @@ export class OFREPWebProvider implements Provider {
...
+   */
+  private _connectSseIfAvailable(result: EvaluateFlagsResponse): void {
+    if (result.eventStreams && result.eventStreams.length > 0) {
+      this.stopPolling();
+      if (!this._sseManager) {
+        this._sseManager = new SseManager(
+          {
+            onRefetch: (metadata) => this._handleSseRefetch(metadata),
+            onStale: () => this._handleSseStale(),
+            onError: () => this._handleSseError(),
```
jonathannorris, line 462: Per ADR-0008, when `onContextChange` triggers a re-fetch and the eventStream URL set is unchanged, existing connections should be reused. The current `SseManager.connect` unconditionally calls `disconnect()` first, so every context change tears down and reconnects even when the URLs are identical.

Recommend diffing the incoming URL set against `this._urls` and skipping the teardown when they match. Without this, every context change produces a brief SSE dark window where a `refetchEvaluation` event could be missed, plus an avoidable EventSource round-trip. This matters more for providers that frequently switch context (e.g. multi-user web apps).

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -286,6 +307,75 @@ export class OFREPWebProvider implements Provider {
...
+    }
+    // Fall back to polling if it's enabled and not already running
+    if (this._pollingInterval > 0 && !this._pollingIntervalId) {
+      this._logger?.info('SSE error — falling back to polling');
+      this._sseManager.dispose();
+      this._sseManager = undefined;
+      this.startPolling();
+    }
+  }
+
```
jonathannorris, line 547: ADR-0008 says: *"if polling is disabled, continue SSE reconnection attempts and rely on explicit refresh triggers such as `onContextChange`."*

Currently when `pollInterval === 0` and SSE hits a fatal error (`readyState === CLOSED`), this handler falls through without doing anything — and since native `EventSource` doesn't auto-reconnect from CLOSED, the provider is silently stuck on cached flags until `onContextChange` fires.

Recommend adding a retry loop with backoff on the else branch so SSE reconnection continues when polling is off. Worth a test too, since this is the hard case the ADR explicitly calls out.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
jsalaber: `connectSSEIfAvailable` looks marginally better

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
marcozabel: I added an exponential backoff 👍

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
marcozabel: It's checking now, if connections can be reused 👍

libs/shared/ofrep-core/src/lib/api/ofrep-api.ts
marcozabel: I added that test 👍

toddbaert:
@lukas-reining another I'd appreciate your input on if you have time. :pray:

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,299 @@
...
+  get connectedUrls(): ReadonlySet<string> {
+    return this._urls;
+  }
+
+  private _connectToUrl(url: string): void {
+    if (this._disposed) {
+      return;
+    }
+
+    const es = new EventSource(url);
```
lukas-reining: I am wondering if we should have an event source implementation here that supports headers and uses the custom headers that the provider accepts in general. 
For the provider we decided that it should have custom header support [[1]](https://github.com/open-feature/protocol/blob/main/guideline/dynamic-context-provider.md#configuration)

I acknowledge that we have specified the url the following[[2]](https://github.com/open-feature/protocol/blob/main/service/adrs/0008-sse-for-bulk-evaluation-changes.md)

```
url (string, optional): The endpoint URL. This is the default representation and is opaque to the provider. 
It may include authentication tokens, channel identifiers, or other vendor-specific query parameters. Implementations must treat this URL as sensitive -- it may contain auth tokens or channel credentials -- and must not log or persist the full URL including query string.
```
But I think many real life implementations will still use something like OAuth2 with Bearer auth or something similar, often behind proxies that can not simply accept a token in the URL. 
In these 
Also Token/Key in the URL also can (not must) be considered problematic in contexts where there is at least one proxy in between that logs the request URL which can be problematic in some use cases.

lukas-reining: Another (additional) option to consider would be to allow injecting the `EventSource` implementation which is a common pattern due to the limitations the default W3C one has. 
@open-feature/spec-evaluation-maintainers @open-feature/sdk-javascript-maintainers

libs/providers/ofrep-web/src/lib/sse-manager.ts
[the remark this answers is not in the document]
lukas-reining: This makes sense and the grace period implementation looks good!

libs/providers/ofrep-web/src/lib/sse-manager.ts
lukas-reining: ~If the server responds with a HTTP `503` here, the event source will go into `CLOSED` immediately and not try to reconnect. If there is a hick up in the infra, e.g. during load peaks, this could result in `EventSource` connections being close, with 503 on reconnection. If this happens, it will not try to reconnect again.~

~The question is, if we want to handle this, which I would prefer, and if we implement an `OpenFeatureEventSource` or do it for all event sources if we go with injectable event source implementations.~

```mermaid
sequenceDiagram
    participant Client as OFREP provider
    participant Server as Server
    
    Client->>Server: GET /events
    Server-->>Client: 200 OK (text/event-stream)
    Note over Client: readyState: OPEN
    
    rect rgb(255, 200, 200)
        Note over Server: Load peak
        Server-xClient: Connection lost
        Note over Client: readyState: CONNECTING
        
        Client->>Server: GET /events (auto-reconnect)
        Server-->>Client: 503 Service Unavailable
        
        Note over Client: readyState: CLOSED<br/>No more reconnection attempts
    end
    
    Note over Server: Server recovers
    Note over Client,Server: Connection remains broken
```

lukas-reining: This also relates to @toddbaert's comment: https://github.com/open-feature/js-sdk-contrib/pull/1509/changes#r3059523635

libs/providers/ofrep-web/src/lib/sse-manager.spec.ts
```diff
@@ -0,0 +1,473 @@
...
+    }
+  }
+
+  static reset(): void {
+    MockEventSource.instances = [];
+  }
+}
+
+// Install mock EventSource globally
+(globalThis as Record<string, unknown>)['EventSource'] = MockEventSource;
```
lukas-reining: Having the `EventSource` implementation optionally injectable would also solve this hack :D

lukas-reining (commented):
This looks very good to me! 
But I would like to clarify the reconnection and header injection before approving.

libs/providers/ofrep-web/src/lib/sse-manager.ts
jonathannorris: Worth flagging that the exponential-backoff retry in `_handleSseError` is intended to recover from this, but it currently no-ops.

After a fatal error, `_connections` and `_urls` aren't cleared. The provider's backoff timer eventually calls `_sseManager.connect(sameUrls)`, but `connect()`'s `urlsUnchanged` check (sse-manager.ts:95-102) sees `_connections.length > 0` (closed instance still in array) and short-circuits. Provider is silently stuck.

Minimal fix in the fatal branch here:

```ts
this._connections = [];
this._urls = new Set();
this._callbacks.onError();
```

Worth a regression test: fatal error → backoff fires → assert a new `EventSource` is created.

jonathannorris: The headers ask is worth scrutinizing against the ADR's reasoning.

ADR-0008 deliberately chose URL-based auth for SSE: (1) it works with native browser `EventSource`, (2) it avoids expanding `Access-Control-Allow-Headers` (the same CORS argument the ADR makes for `flagConfigEtag`/`flagConfigLastModified` in [line 104](https://github.com/open-feature/protocol/blob/main/service/adrs/0008-sse-for-bulk-evaluation-changes.md)).

On proxy URL logging: the ADR addresses it in line 64 (don't log the URL, vendors can issue short-lived tokens if log exposure is a concern). Proxies log headers too, so headers don't fix this.

Injectable EventSource is reasonable , it solves the spec-file mock hack and supports non-browser environments, but I'd suggest landing it as a separate PR if and when a vendor needs it, rather than expanding the scope here. Custom-header support specifically conflicts with the ADR's design intent and would need a follow-up ADR conversation.

libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
```diff
@@ -9,4 +9,20 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
...
+   * If neither is set, defaults to 120 seconds.
+   */
+  inactivityDelaySec?: number;
+
+  /**
+   * Controls the background change-detection strategy per ADR-0008.
+   *
+   * - `'sse'` (default): use SSE when the server advertises event streams, fall back to polling.
+   * - `'polling'`: always use polling, even when the server returns `eventStreams`.
+   * - `'none'`: no background refresh; flags are only re-fetched on explicit context changes.
```
jonathannorris, line 39: Worth adding test coverage for the new behaviors landed in this round:

1. **`changeDetection` modes**: only `'sse'` (default) is covered implicitly by existing tests. `'polling'` (server returns `eventStreams`, provider ignores it) and `'none'` (no background refresh at all) need explicit coverage on both init and `onContextChange` paths.

2. **SSE retry-with-backoff** when polling is disabled: the new logic in `_handleSseError` (exponential backoff, counter reset, timer cleanup on `onClose`, gating by `changeDetection !== 'none'`) is entirely untested. A regression test asserting "fatal error → backoff fires → new EventSource is created" would also surface the stale `_connections` bug I flagged in [Lukas's 503 thread](https://github.com/open-feature/js-sdk-contrib/pull/1509#discussion_r3209787969).

libs/providers/ofrep-web/src/lib/sse-manager.ts
lukas-reining: > ADR-0008 deliberately chose URL-based auth for SSE

No it does not. It only states that `It may include authentication tokens`. This is permissive, not prescriptive.

> it avoids expanding Access-Control-Allow-Headers (the same CORS argument the ADR makes for flagConfigEtag/flagConfigLastModified in [line 104](https://github.com/open-feature/protocol/blob/main/service/adrs/0008-sse-for-bulk-evaluation-changes.md)).

I am not saying that we should support any auth use cases directly, I just want to keep it open by allowing the custom headers. As said in many real life scenarios there are OAuth proxies in front of apps that require the use of a header. And especially with the opaqueness requirement for the URLs, configuring the SSE exceptions in the routing will be very hard. 

> On proxy URL logging: the ADR addresses it in line 64 (don't log the URL, vendors can issue short-lived tokens if log exposure is a concern). Proxies log headers too, so headers don't fix this.

Behind things like corporate proxies and as said something like OAuth gateways, you can never assume no logging. There can be any middlebox in between. 
I get the point about short lived tokens but that would also mean reevaluating after the connection broke. 
And for JWTs, it is generally discouraged: https://support.auth0.com/center/s/article/securely-passing-a-jwt-in-url-query-parameters-for-a-get-request.
In OAuth 2.1 it is actually not even allowed anymore: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1#section-5.1

lukas-reining: > Worth flagging that the exponential-backoff retry in _handleSseError is intended to recover from this, but it currently no-ops.

Oh yeah I was confused you are right @jonathannorris. Sure the implementation already should do that and the only gap is the missing reset.

jonathannorris: Fair enough, you're right. I was reading more into the ADR than it actually says, and the OAuth 2.1 prohibition on Bearer-in-URL is unambiguous.

I think the injectable `EventSource` that landed covers the practical need, header-based auth lives at the SSE client layer via a polyfill, provider stays small.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -286,6 +311,107 @@ export class OFREPWebProvider implements Provider {
...
+      const delayMs = Math.min(1_000 * Math.pow(2, this._sseRetryCount), 60_000);
+      this._sseRetryCount++;
+      this._logger?.info(`SSE error — polling disabled, retrying SSE in ${delayMs}ms (attempt ${this._sseRetryCount})`);
+      this._sseRetryTimerId = setTimeout(async () => {
+        this._sseRetryTimerId = undefined;
+        try {
+          const result = await this._fetchFlags(this._context);
+          this._connectSseIfAvailable(result);
+        } catch {
+          // _connectSseIfAvailable was not reached; onError will schedule the next retry
```
jonathannorris: The empty `catch` here strands the retry loop. The comment says "onError will schedule the next retry," but if `_fetchFlags` throws (network error, 5xx on bulk-eval, etc.), no `EventSource` is constructed and no `onError` ever fires — `_sseRetryCount` is never reset, but no further attempts happen either.

Concrete scenario: SSE goes down because the OFREP server is having a bad day. The bulk-eval endpoint is also failing (e.g. 503s). The first retry hits this catch, swallows the error, and the provider is silently stuck.

Suggest scheduling the next attempt from the catch:

```ts
} catch (error) {
  this._logger?.warn(`SSE retry fetchFlags failed: ${error} — scheduling next attempt`);
  this._handleSseError();  // continues exponential backoff
}
```

libs/providers/ofrep-web/src/lib/sse-manager.ts
lukas-reining: Yes, I think so too @jonathannorris!

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,307 @@
...
+    // readyState EventSource.CONNECTING — transient error, browser is auto-reconnecting.
+    // Enter grace period: emit STALE now, ERROR only if recovery doesn't happen.
+    if (!this._stale) {
+      this._stale = true;
+      this._logger?.info('SSE connection interrupted — entering grace period');
+      this._callbacks.onStale();
+      this._gracePeriodTimerId = setTimeout(() => {
+        this._logger?.warn('SSE grace period expired — giving up');
+        this._stale = false;
+        this._callbacks.onError();
```
jonathannorris, line 209: The fatal branch clears `_connections` and `_urls` before calling `onError()` (lines 179–182) so the backoff retry's `connect()` can create a fresh `EventSource`. This path does not.

After grace period expiry, the old `EventSource` is still in `_connections` in `CONNECTING` state. When the provider's backoff timer fires and calls `connect(sameUrls)`, the `urlsUnchanged` check (`_connections.length > 0 && ...`) short-circuits and returns without opening a new connection. Provider is silently stuck on the orphaned `EventSource`, which also leaks and keeps retrying in the background.

Suggest mirroring the fatal branch:

```ts
this._gracePeriodTimerId = setTimeout(() => {
  this._logger?.warn("SSE grace period expired — giving up");
  this._stale = false;
  for (const conn of this._connections) conn.close();
  this._connections = [];
  this._urls = new Set();
  this._callbacks.onError();
}, DEFAULT_GRACE_PERIOD_SEC * 1000);
```

Worth a regression test: grace period expires → provider schedules backoff → retry calls `connect()` → new `EventSource` is created (not short-circuited).

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -93,16 +97,22 @@ export class OFREPWebProvider implements Provider {
     try {
       this._context = context;
 
+      let result: EvaluateFlagsResponse | undefined;
       if (this._cacheMode === 'network-first') {
         await this._initNetworkFirst(context);
```
jonathannorris: `_initNetworkFirst` calls `_fetchFlags` internally but discards the return value, so any `eventStreams` in the response are silently dropped and SSE is never set up in `network-first` mode. Suggest returning the result from `_initNetworkFirst` and passing it to `_connectSseIfAvailable`.

jonathannorris (approved):

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,315 @@
...
+      return;
+    }
+
+    // readyState EventSource.CONNECTING — transient error, browser is auto-reconnecting.
+    // Enter grace period: emit STALE now, ERROR only if recovery doesn't happen.
+    if (!this._stale) {
+      this._stale = true;
+      this._logger?.info('SSE connection interrupted — entering grace period');
+      this._callbacks.onStale();
+      this._gracePeriodTimerId = setTimeout(() => {
```
jonathannorris, line 198: Good catch by @erka — the grace period is only cancelled on a received message (`_handleMessage`), not on the `open` event itself. If the connection recovers but no message arrives before the 30s window expires, `onError` fires incorrectly even though the connection is healthy. The fix is to listen to the `open` event on each `EventSource` and clear the grace period there:

```ts
es.addEventListener("open", () => {
  if (this._stale) {
    this._clearGracePeriod();
    this._stale = false;
  }
});
```

toddbaert: I added this: https://github.com/open-feature/js-sdk-contrib/pull/1509/changes/833425e7b6a3780f2c581f95915d9bc9b7b80408, thanks @erka

toddbaert:
I'd like to merge this tomorrow - please let me know if you have objections @lukas-reining @jsalaber @marcozabel @chrfwow

toddbaert:
> I'd like to merge this tomorrow - please let me know if you have objections @lukas-reining @jsalaber @marcozabel @chrfwow

This has been open for a long time now, and as far as I can tell all the concerns are resolved, particularly those brought up by @lukas-reining . I am going to merge. If there's anything anyone wants changed, we can still fix that before release (I will release this week).

lukas-reining:
> This has been open for a long time now, and as far as I can tell all the concerns are resolved, particularly those brought up by @lukas-reining . I am going to merge. If there's anything anyone wants changed, we can still fix that before release (I will release this week).

Yes it is fine! Sorry for the delay.

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    // Diff URLs — skip teardown when connections are active and the set is unchanged
+    const urlsUnchanged =
+      this._connections.length > 0 && urls.size === this._urls.size && [...urls].every((u) => this._urls.has(u));
+
+    if (urlsUnchanged) {
+      return;
+    }
+
```
erka, line 105: ```It never reconnects to the new SSE endpoint if the URLs change during a refetch.```

PR #1510, opened 2026-04-09
feat(ofrep)!: Disable polling by default

marcozabel:
## This PR
Disabled the default polling interval to make way for SSE.

### Related Issues

Part of https://github.com/open-feature/js-sdk-contrib/issues/1497

### DIFF
--- a/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
+++ b/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
@@ -5,8 +5,9 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
    * pollInterval is the time in milliseconds to wait between we call the OFREP
    * API to get the latest evaluation of your flags.
    *
-   * If a negative number is provided, the provider will not poll the OFREP API.
-   * Default: 30000
+   * If a negative number or 0 is provided, the provider will not poll the OFREP API.
+   * This is the default behavior. Polling is available as an opt-in configuration.
+   * Default: 0 (disabled)
    */
   pollInterval?: number; // in milliseconds
 };
--- a/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
+++ b/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
@@ -10,4 +10,12 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
    * Default: 0 (disabled)
    */
   pollInterval?: number; // in milliseconds
+
+  /**
+   * When enabled, the provider will re-fetch flags whenever the page becomes visible
+   * (e.g. user switches back to the tab).
+   *
+   * Default: false (disabled)
+   */
+  refreshOnVisibilityChange?: boolean;
 };
--- a/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
+++ b/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
@@ -17,12 +17,12 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
   pollInterval?: number;
 
   /**
-   * When enabled, the provider will re-fetch flags whenever the page becomes visible
-   * (e.g. user switches back to the tab).
+   * When set to true, disables the automatic flag re-fetch that occurs whenever the
+   * page becomes visible (e.g. user switches back to the tab).
    *
-   * Default: false (disabled)
+   * Default: false (visibility refresh is enabled by default, per ADR-0010)
    */
-  refreshOnVisibilityChange?: boolean;
+  disableVisibilityRefresh?: boolean;
 
   /**
    * cacheMode controls whether and how the provider uses local persistent storage.
--- a/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
+++ b/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
@@ -10,8 +10,9 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
    * pollInterval is the time in milliseconds to wait between calls to the OFREP
    * API to get the latest evaluation of your flags.
    *
-   * If the value is 0 or negative, polling is disabled.
-   * Default: 30000
+   * If a negative number or 0 is provided, the provider will not poll the OFREP API.
+   * This is the default behavior. Polling is available as an opt-in configuration.
+   * Default: 0 (disabled)
    */
   pollInterval?: number;
 
--- a/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
+++ b/libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
@@ -17,6 +17,14 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
   pollInterval?: number;
 
   /**
+   * When enabled, the provider will re-fetch flags whenever the page becomes visible
+   * (e.g. user switches back to the tab).
+   *
+   * Default: false (disabled)
+   */
+  refreshOnVisibilityChange?: boolean;
+
+  /**
    * cacheMode controls whether and how the provider uses local persistent storage.
    *
    * - `'local-cache-first'` (default): load from the persisted cache immediately on startup
--- a/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
+++ b/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
@@ -57,6 +57,7 @@ export class OFREPWebProvider implements Provider {
   private _flagSetMetadataCache?: MetadataCache;
   private _context: EvaluationContext | undefined;
   private _pollingIntervalId?: number;
+  private _isFetching = false;
   private _visibilityChangeHandler = this._onVisibilityChange.bind(this);
 
   constructor(options: OFREPWebProviderOptions, logger?: Logger) {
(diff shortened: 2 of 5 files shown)


libs/providers/ofrep-web/src/lib/model/ofrep-web-provider-options.ts
```diff
@@ -5,8 +5,9 @@ export type OFREPWebProviderOptions = OFREPProviderBaseOptions & {
    * pollInterval is the time in milliseconds to wait between we call the OFREP
    * API to get the latest evaluation of your flags.
    *
-   * If a negative number is provided, the provider will not poll the OFREP API.
-   * Default: 30000
+   * If a negative number or 0 is provided, the provider will not poll the OFREP API.
+   * This is the default behavior. Polling is available as an opt-in configuration.
+   * Default: 0 (disabled)
```
jsalaber, line 15: does the issue specify changing the polling behaviour?

marcozabel: > 2. On SSE connection failure, fall back to polling if polling is enabled; otherwise continue SSE reconnection attempts

This reads to me like it is not enabled by default. Did I misread that @toddbaert ?

jsalaber: my mistake, it is part of this new adr that hasn't been merged yet https://github.com/open-feature/protocol/pull/69

jsalaber (approved):

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -319,4 +347,15 @@ export class OFREPWebProvider implements Provider {
...
+
+  /**
+   * Handler for visibility changes (page/app becoming visible)
+   * Re-fetches flags when the document becomes visible
+   * @private
+   */
+  private _onVisibilityChange() {
+    if (typeof document !== 'undefined' && document.visibilityState === 'visible') {
+      this._refreshFlags('visibility change');
+    }
```
lukas-reining:
```suggestion
    if (document?.visibilityState === 'visible') {
      this._refreshFlags('visibility change');
    }
```
Nit: I think this is a little more concise

lukas-reining: Mhh I get why we added this but it was not defined that we refetch on visibility changes isn't it? @open-feature/spec-evaluation-approvers
Maybe this is something we should add to the provider spec then.

lukas-reining (commented):
This looks good to me! 
I would just like to confirm that we want to have the visibility refresh.

jsalaber (approved):

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -86,6 +88,11 @@ export class OFREPWebProvider implements Provider {
         this.startPolling();
       }
 
+      // Listen for page/app visibility changes to refetch flags when becoming visible
+      if (typeof document !== 'undefined') {
+        document.addEventListener('visibilitychange', this._visibilityChangeHandler);
+      }
```
MattIPv4: I feel this should be opt-in, or at least opt-out? Some folks may well want to load flags once on app start and ensure they don't change again?

MattIPv4 (changes requested):

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -319,4 +347,15 @@ export class OFREPWebProvider implements Provider {
...
   }
+
+  /**
+   * Handler for visibility changes (page/app becoming visible)
+   * Re-fetches flags when the document becomes visible
+   * @private
+   */
+  private _onVisibilityChange() {
+    if (document?.visibilityState === 'visible') {
+      this._refreshFlags('visibility change');
```
jonathannorris: Once SSE lands (tracking ADR-0010 / #1497), this handler will need to coordinate with the SSE connection per the ADR:

> On foreground, if the provider has an active SSE connection, no re-fetch is needed as the SSE connection handles change detection. If the SSE connection was closed during inactivity... the provider should reconnect and perform a full unconditional re-fetch.

Worth leaving a `// TODO(SSE):` marker here so it's not missed when SSE is wired up. Something like:

```ts
private _onVisibilityChange() {
  if (typeof document !== 'undefined' && document.visibilityState === 'visible') {
    // TODO(SSE): once SSE is implemented, skip refresh if SSE is connected,
    // and perform an unconditional re-fetch if SSE was closed during inactivity.
    // See ADR-0010: https://github.com/open-feature/protocol/pull/69
    this._refreshFlags('visibility change');
  }
}
```

jonathannorris:
One thing worth considering: `_refreshFlags` now emits `ClientProviderEvents.Stale` on any failure, including visibility-change triggers. Previously this was polling-only, where sustained refresh failure is a reasonable staleness signal. Applied to a single visibility-change failure (e.g. user tabs back before Wi-Fi reconnects), it flips the provider into `STALE` from one missed refresh even though the in-memory cache is still valid.

Two possible directions:

1. Only emit `STALE` from the polling path; log-and-swallow for one-shot triggers like visibility-change
2. Leave as-is

jonathannorris:
Reviewed against ADR-0010 (https://github.com/open-feature/protocol/pull/69) overall looks good.

Two non-blocking follow-up observations:

**Tests.** No changes to `ofrep-web-provider.spec.ts`. Worth adding coverage for the new default (no polling when unset), the `visibilitychange` → fetch path, and `onClose` listener cleanup.

**Docs.** The `README.md` headline example still shows `pollInterval: 60000`, which now misrepresents the default. ADR-0010 also calls out that SDK docs should note the behavior change and the opt-in path for polling.

Looks good! thanks!

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
MattIPv4: I imagine this will be especially true in micro-frontend architectures where there may be _many_ providers defined in different domains -- if they all refresh every time the page visiblity changes, that is a lot of network requests being fired off at once.

marcozabel: @lukas-reining @MattIPv4 thanks for the suggestion, I added a config option for opt-in 👍

MattIPv4: Thank you :)

MattIPv4 (approved):
Very much a fan of this, we disable polling by default in all our usages currently.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
jonathannorris: +1 to MattIPv4's "or at least opt-out" framing, but I'd strongly recommend going with opt-out that rather than opt-in.

ADR-0010 (https://github.com/open-feature/protocol/pull/69) explicitly lists page visibility as one of the default event-driven refresh triggers replacing timer polling. Defaulting to off means most users get a strictly worse default than the polling behavior this PR replaces. The micro-frontend case is real but niche, and is exactly where opt-out fits: those integrators know their setup and can disable it.

MattIPv4: 👍 No major concerns w/ this being opt-out, though I'm going to have quite a bit of work ahead of me to try and catch this update in every microfrontend we have to set this opt-out before we start spamming requests 😓

jonathannorris: Another option would be exporting a simple like `OFREPMircoWebProvider` or something that has it disabled by default.

MattIPv4: I don't think that'd really change much, we'd still need to do a code change in every microfrontend we have to switch to that, or adopt the new opt-out.

toddbaert:
@lukas-reining Do you approve? I'd like your opinion on this one.

toddbaert (approved):

lukas-reining (approved):
Approved, looks good!
I was waiting for this: https://github.com/open-feature/js-sdk-contrib/pull/1510#discussion_r3129704211

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
lukas-reining: Resolved due to: https://github.com/open-feature/js-sdk-contrib/pull/1510#discussion_r3129704211

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
jonathannorris: Circling back once more before this merges: the migration cost cuts both ways. Opt-in means every existing app that wants ADR-aligned behavior has to discover the new option and add it; opt-out means microfrontend setups need a one-line addition.

ADR-0010 is also explicit that visibility refresh is a default trigger, so shipping this as opt-in means the reference implementation diverges from the spec on day one.

toddbaert: I think I agree with @jonathannorris on this one.

I appreciate the microfrontend example @MattIPv4 , but I suspect that's not the majority of cases, so I think we should align defaults with the ADR.

toddbaert:
Since this is a breaking behavioral change, I've added a `!` to the commit prefix.

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
MattIPv4: Yeah, makes sense, as I said, no major concerns w/ this being opt-out, I'll just try to chase down all our teams using this before they update!

jonathannorris (approved):

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
beeme1mr: I'm going to go ahead and merge this and tag @MattIPv4 on the release PR so we can let you know when the update is going live.

MattIPv4: Works for me 👍 This is marked as breaking now, so hopefully we'll catch it, but would appreciate the heads up!

PR #1518, opened 2026-04-17
feat: initial localStorage provider

MattIPv4:
## This PR

- Introduces a new client provider based on values in `localStorage`

### Notes

This is heavily based on the existing env var provider.

### Follow-up Tasks

I think I'd like to ideally find a way for the provider to "reconcile" when changes are observed in `localStorage` (scoped to the `prefix` provided), so that flag values can update as `localStorage` updates rather than needing a page refresh. Can be a follow-up, though, not part of the initial implementation.


libs/providers/localstorage/README.md
```diff
@@ -0,0 +1,104 @@
...
+The default value will be returned if there is no matching `localStorage` entry or the value can't be cast to the desired type.
+
+```typescript
+import { OpenFeature, MultiProvider } from '@openfeature/web-sdk';
+import { LocalStorageProvider } from '@openfeature/localstorage-provider';
+
+// Register the localStorage provider globally, combined with any other providers you may be using
+OpenFeature.setProvider(
+  new MultiProvider([{ provider: new LocalStorageProvider() }, { provider: new OtherProvider() }]),
+);
```
lukas-reining, line 32: I see why you chose this example but maybe put one sentence why the example uses the multi provider.

libs/providers/localstorage/README.md
```diff
@@ -0,0 +1,104 @@
...
+For example, if you have a flag with the key `new-feature`, the provider will look for a `localStorage` entry with the key `openfeature.new-feature` to determine the value of that flag.
+This allows you to define feature flags in `localStorage` without worrying about naming collisions with other data stored there.
+
+## Examples
+
+### Boolean example
+
+```javascript
+// In the browser's developer console set a localStorage entry for the flag
+localStorage.setItem('openfeature.enable-new-feature', 'true');
```
lukas-reining, line 77: I am wondering if the provider could contain a method for initializing. 
Typically app authors never directly interact with the provider but something like `myProvider.setFlags(...)` and `myProvider.clearFlags()` that can be called after initialization could be good to have. 
`setFlags` would simply set what it has, ignoring any other flag, `clearFlags` would clear all flags by the prefix. 
I would not use the constructor because there idempotency and startup behavior would have to be defined too general I think, while the methods let the app author decide.

lukas-reining (approved):
This looks great @MattIPv4!
Left one proposal..

lukas-reining (approved):
Yes thats what I meant, thanks @MattIPv4!

libs/providers/env-var/src/lib/env-var-provider.spec.ts
```diff
@@ -110,18 +110,18 @@ describe('Environment Variable Provider', () => {
...
     });
 
     it('should return openfeature', async () => {
-      process.env['OBJ_VALUE'] = 'true';
+      process.env['OBJ_VALUE'] = '"openfeature"';
       await expect(envVarProvider.resolveObjectEvaluation('obj-value')).resolves.toMatchObject({
         reason: 'STATIC',
-        value: true,
+        value: 'openfeature',
       });
```
toddbaert, line 125: The way our release automation works, this will cause a new "feature" (and CHANGELOG entry/release) in the env var provider with the title `feat: initial localStorage provider` - can you revert these and keep your changes just to the new package?

libs/providers/localstorage/src/lib/localstorage-provider.ts
```diff
@@ -0,0 +1,165 @@
...
+} from '@openfeature/web-sdk';
+
+export type Config = {
+  /**
+   * Prefix to use when mapping localStorage keys to flag keys.
+   * This allows you to avoid potential naming conflicts with other data in localStorage.
+   *
+   * @default 'openfeature.'
+   */
+  prefix?: string;
```
toddbaert, line 19: I was hoping you'd include such an option. 👍

libs/providers/env-var/src/lib/env-var-provider.spec.ts
MattIPv4: Ah yeh, sure thing

libs/providers/localstorage/src/lib/localstorage-provider.ts
```diff
@@ -0,0 +1,165 @@
...
+  type Provider,
+  type ResolutionDetails,
+  ClientProviderEvents,
+  FlagNotFoundError,
+  OpenFeatureEventEmitter,
+  ParseError,
+  StandardResolutionReasons,
+} from '@openfeature/web-sdk';
+
+export type Config = {
```
toddbaert, line 12: We could add an option to sub out `localstorage` with `sessionstorage`. They have the exact same API but obviously session is more transient. Just an idea; not sure if it's worth it.

toddbaert (approved):

libs/providers/localstorage/src/lib/localstorage-provider.ts
MattIPv4: Ah interesting idea, could be a good follow-up PR.

libs/providers/localstorage/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@openfeature/localstorage-provider",
+  "version": "0.0.1",
```
toddbaert:
```suggestion
  "version": "0.1.0",
```

This should match the manifest. The automation would probably sync it up, but just to be sure...

toddbaert:
LGTM - will merge/release tomorrow if I don't see objections.

libs/providers/localstorage/package.json
MattIPv4: Will add to the commit that fixes the other issues with the scaffolding, no worries 🫠

libs/providers/localstorage/src/lib/localstorage-provider.ts
```diff
@@ -0,0 +1,165 @@
...
+      flagsChanged: flagKeys,
+    });
+  }
+
+  private evaluateLocalStorage<T extends JsonValue>(
+    flagKey: string,
+    parse: (value: string) => T,
+  ): ResolutionDetails<T> {
+    const localStorageKey = `${this.options.prefix ?? ''}${flagKey}`;
+    const value = typeof localStorage !== 'undefined' ? localStorage.getItem(localStorageKey) : null;
```
toddbaert, line 144: This guard doesn't exist in set/clear flags.

MattIPv4: I didn't add it there, as I think if a user is explicitly calling those methods, they should get the error if they're using them in a non-compatible environment?

I only added it here to ensure that if this gets initialized or a flag is evaluated during SSR, it won't throw.

PR #1533, opened 2026-04-27
fix: added internal and bad request ofrep errors

jsalaber:
## This PR

- fixes not throwing proper error for 400s and 500s, catches any other errors as well that don't have `response.ok`


toddbaert:
@jsalaber and I both thought 400/500 were handled - they are, just in a special area to result in particular OpenFeature exceptions.

PR #1542, opened 2026-04-30
chore: migrate to latest nx

toddbaert:
This PR just runs `nx migrate latest` and the generated migrations. Currently packaging/publishing is broken due to some recent dep updates, but this fixes it and clears some security issues.

I smoke tested one of the built packages and all the bundling etc worked fine.

Issue #101, opened 2022-09-08
Flagd gRPC object deserialization is not working properly


beeme1mr:
Objects returned by flagd are not deserialized as expected:

http:
```
{
  "key": "val"
}
```

grpc
```
{
  "fields": {
    "key": {
      "kind": {
        "oneofKind": "stringValue",
        "stringValue": "val"
      }
    }
  }
}
```

beeme1mr:
FYI, @james-milligan

toddbaert:
I'll take this one, just a little missing parsing, but I'm going to add some tests.

toddbaert:
I'm also using this chance to remove HTTP support from the provider as we discussed.

PR #1544, opened 2026-05-08
chore: resolve open dependabot security alerts

jonathannorris:
## Summary
- Merged latest `main` and resolved the resulting lockfile/package.json conflict.
- Resolved all 56 currently-open Dependabot alerts.
- Root-cause fixes (direct/parent bumps) where possible; scoped `overrides` only for deep dev-tooling transitives with no reachable parent bump, each upper-bounded to a major version verified compatible with its consumer(s).

## Dependabot Alerts Resolved (56 of 56)

| Package | Alerts | Severity | Fix |
|---|---|---|---|
| `axios` | 16 | high/medium/low | resolved to `1.16.0` |
| `undici` | 9 | high/medium/low | resolved to `7.28.0` / `6.27.0` |
| `lodash` | 3 | high/medium | resolved to `4.18.1` |
| `protobufjs` | 3 | high/medium | resolved to `7.6.4` |
| `@grpc/grpc-js` | 2 | high | direct bump to `^1.14.4` |
| `fast-uri` (root + flagd-core) | 4 | high | resolved to `3.1.2` |
| `fast-xml-builder` (root + aws-ssm) | 4 | high/medium | resolved to `1.2.0` |
| `js-yaml` (root + go-feature-flag) | 3 | medium | override `>=4.2.0 <5.0.0`; go-feature-flag already at `4.2.0` |
| `uuid` | 2 | medium | override `>=11.1.1 <12.0.0` (capped below v12, which dropped CJS support and broke `jest-cucumber`) |
| `qs` | 1 | medium | override `>=6.15.2` |
| `ws` | 1 | high | override `>=8.21.0` |
| `@xhmikosr/decompress` | 1 | critical | override `>=10.2.1 <11.0.0` |
| `form-data` | 1 | high | resolved to `4.0.6` (via `nx` bump) |
| `piscina` | 1 | high | resolved to `4.9.3` |
| `tmp` | 1 | high | resolved to `0.2.6`/`0.2.7` (via `nx` bump) |
| `follow-redirects` | 1 | medium | resolved to `1.16.0` |
| `minimatch` | 1 | high | resolved to `7.4.9` |
| `@babel/plugin-transform-modules-systemjs` | 1 | high | resolved to `7.29.4` (via `nx` bump) |
| `@babel/core` | 1 | low | resolved to `7.29.7` |

Also synced `tools/workspace-plugin/package.json`'s `@nx/*` pins to `22.7.6` to match root (was failing `@nx/dependency-checks` lint).

## Verification
- `nx affected` `lint`, `build`, `test` vs. `origin/main` all pass.
- `npm ci --dry-run` confirms the lockfile is in sync.


MattIPv4 (commented):
Same thought as in the other PR, do we need to keep the overrides, or can they be removed now that the lockfiles have been bumped?

jonathannorris:
Removing the `overrides` block from root `package.json`. These only affect the monorepo's dev environment — they're not inherited by consumers of any published package, so they don't improve security posture for SDK users.

The following alerts remain open, all tied to dev tooling we don't control:

| Alert(s) | Package | Severity | Root cause |
|----------|---------|----------|------------|
| #256 | `@babel/plugin-transform-modules-systemjs` | High | `nx@22.7.1` pulls in a vulnerable version |
| #217 | `follow-redirects` | Medium | `nx@22.7.1` pins `1.15.11` exactly |
| #135 | `minimatch` | High | transitively via `verdaccio`'s `@verdaccio/core` |
| #202, #201, #101 | `lodash` | High/Medium | `@verdaccio/local-storage-legacy` pins an older version |

These will resolve as `nx` and `verdaccio` release updates with patched transitive deps.

libs/shared/flagd-core/package-lock.json
```diff
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "imurmurhash": "^0.1.4",
-        "json-logic-engine": "^4.0.2",
+        "json-logic-engine": "5.0.7",
```
MattIPv4, line 13: ++ to what gemini flagged here, this top-level dep has been bumped without the matching package.json change? It has also lost its semver range character?

MattIPv4:
The bump to the v3 lockfile makes this PR pretty unreviewable, and it looks like the lockfile isn't actually valid? How was it generated?

jonathannorris:
@MattIPv4 yea the lock file change does make this hard to review. I'll try and revert that, it seemed fairly safe based on this research: 

```
Root `package-lock.json` upgraded from lockfileVersion 2 → 3 (17 of 22 lockfiles in the repo are already v3; CI matrix uses Node 20/22/24, all of which ship with npm 10+; release pipeline explicitly pins npm `^11.5.1`)
```

libs/shared/flagd-core/package-lock.json
jonathannorris: This is actually the lock file catching up with an already-committed `package.json` change. PR #1519 (custom operator conformance fixes) bumped `json-logic-engine` to `5.0.7` (pinned, no range) in `flagd-core/package.json` but never regenerated the lock file. Regenerating the locks here resolved it consistently with `package.json`. The missing `^` was also introduced in #1519, not here.

MattIPv4: Ah interesting, ty! I wonder how that landed with a bad lockfile that didn't break CI

MattIPv4:
Yeh, I think v2 -> v3 _should_ be safe, but it doesn't seem like it was done correctly here? I'd probably lean toward just resolving the security alerts in this PR, keeping it v2, and then doing the v2 -> v3 in a dedicated PR with a reproducible command so we can verify it is just a format change and not including any dependency changes.

MattIPv4 (approved):

.nxignore
MattIPv4, line 1: Non-blocking, but intentionally including this?

package.json
```diff
@@ -54,15 +54,15 @@
...
-    "@nx/workspace": "22.7.1",
+    "@nx/devkit": "22.7.5",
+    "@nx/eslint": "22.7.5",
+    "@nx/eslint-plugin": "22.7.5",
+    "@nx/jest": "22.7.5",
+    "@nx/js": "22.7.5",
+    "@nx/plugin": "22.7.5",
+    "@nx/rollup": "22.7.5",
+    "@nx/web": "22.7.5",
+    "@nx/workspace": "22.7.5",
```
toddbaert: Did you use the `nx migrate` command for this? If not, I would suggest not touching NX deps. They often come with specific migration scripts only run with the above command.

jonathannorris: yea, `nx migrate` wasn't used here. The intent was to pull in `tmp@0.2.6` (alert #264), but `nx@20.3.1` already declares `tmp: ~0.2.1`, which covers 0.2.6 — so a lockfile-only bump would've been enough. I'll revert the NX changes and use `npm update tmp` instead.

jonathannorris: actually, scratch that — `nx@22.7.2` (already on the branch from a prior merge) pins `tmp` at exactly `0.2.4`, so there's no range to exploit with a lockfile-only bump. The 22.7.2 → 22.7.5 bump was necessary to get `tmp@0.2.6`. I ran `nx migrate 22.7.5` locally to verify — no migration scripts exist between those two versions, so the bare version bump is equivalent.

sago2k8 (approved):
LGTM

package.json
```diff
@@ -97,18 +97,18 @@
     "jest-cucumber": "^4.4.0",
     "jest-environment-jsdom": "30.0.5",
     "jest-fetch-mock": "^3.0.3",
+    "jest-util": "30.0.5",
     "jest-websocket-mock": "^2.4.0",
```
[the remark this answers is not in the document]
jonathannorris: jest-util is already present on main (unrelated to this PR's diff — it came in via the nx migration in #1542). Since it's not part of the dependabot fix set, I'll leave removing it to a separate cleanup PR rather than scope-creep this one.

lukas-reining (approved):

toddbaert (approved):

PR #1546, opened 2026-05-11
fix: wasm memory copying in wrapper

AleksandrasMaliuginas-vinted:
## This PR

Fixes a memory-copying issue in the WASM wrapper. Wrapper used to take `inputStr.length` for byte buffer to encode the `inputStr`. 
The problem is when inputStr contains characters that are UTF-8 but encoded in two or more bytes. This results in trying to write to non-allocated memory space, resulting in invalid memory access.

This PR fixes this by first encoding the string and using byte array lenghts for allocation.


thomaspoignant (approved):

thomaspoignant:
@AleksandrasMaliuginas-vinted thanks a lot for this pull request, and good catch.
I'll release a new version Today for this change.

PR #1535, opened 2026-04-30
chore(main): release ofrep-web-provider 0.4.0

openfeaturebot:
:robot: I have created a release *beep* *boop*
---

### ⚠ BREAKING CHANGES

* **ofrep:** Disable polling by default ([#1510](https://github.com/open-feature/js-sdk-contrib/issues/1510))

### 🐛 Bug Fixes

* **ofrep-web:** fall back to FNV-1a hash when crypto.subtle is unavailable ([#1549](https://github.com/open-feature/js-sdk-contrib/issues/1549)) ([00b2457](https://github.com/open-feature/js-sdk-contrib/commit/00b2457f9a8aa593200cef0d486044a20ebeb8c6))

### ✨ New Features

* added local persistence to ofrep-web provider ([#1508](https://github.com/open-feature/js-sdk-contrib/issues/1508)) ([3103f2c](https://github.com/open-feature/js-sdk-contrib/commit/3103f2cb92037886ee32a2076542a717b4fcc52f))
* **ofrep-web:** add SSE event stream support ([#1497](https://github.com/open-feature/js-sdk-contrib/issues/1497)) ([#1509](https://github.com/open-feature/js-sdk-contrib/issues/1509)) ([7b79536](https://github.com/open-feature/js-sdk-contrib/commit/7b7953650b6437d46c9aa06b18377a2f49505261))
* **ofrep:** Disable polling by default ([#1510](https://github.com/open-feature/js-sdk-contrib/issues/1510)) ([d860849](https://github.com/open-feature/js-sdk-contrib/commit/d860849ee0b3a5b0c29819d8202a7620dc0a6a6d))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #1549, opened 2026-05-12
fix(ofrep-web): fall back to FNV-1a hash when crypto.subtle is unavailable

jonathannorris:
## Summary

- `Storage._hashInput` used `crypto.subtle.digest` unconditionally, which throws in non-secure contexts (HTTP on non-localhost IPs, some older browsers)
- The error propagated through `_fetchFlags` and caused `initialize` to fail in those environments
- Added a `_fnvHash` fallback (double-pass FNV-1a-32, 16 hex chars) used when `crypto.subtle` is not available; SHA-256 behavior is unchanged in secure contexts
- Added tests covering the non-secure-context path

### Notes

An alternative approach would be to use `btoa(input)` as the fallback instead of FNV — it's universally available including non-secure contexts and requires no custom logic. The trade-off is that `btoa` is just base64 encoding, so the targeting key is directly recoverable from the localStorage key name (breaking the privacy property the hash provides). FNV-1a preserves that property without requiring the Crypto API, which is why it was chosen here.

### Related Issues

Addresses review feedback in #1535


beeme1mr (approved):

lukas-reining (approved):
Makes sense to me!

PR #1551, opened 2026-05-20
chore(launchdarkly)!: upgrade to 4.x

### LINKED ISSUE #1543: [launchdarkly-client] Migrate provider to the new package at @launchdarkly/js-client-sdk
It looks like LaunchDarkly moved their SDK package over to https://www.npmjs.com/package/@launchdarkly/js-client-sdk. The OpenFeature provider is still based on the older v3.x https://www.npmjs.com/package/launchdarkly-js-client-sdk.

LD's JS client has also been bumped to v4.x which introduces some breaking changes and fixes. See their migration guide: https://launchdarkly.com/docs/sdk/client-side/javascript/migration-3-to-4

Happy to help contribute but unsure if support somehow needs to be maintained for older versions or if this can be a major bump.

afharo:
## This PR

- upgrades the launchdarkly client from 3.x to 4.x (which involves a dependency rename)

### Related Issues

Fixes #1543

### Notes

The migration guide is https://launchdarkly.com/docs/sdk/client-side/javascript/migration-3-to-4


afharo:
@kinyoklion / @sago2k8 , I'd appreciate your input on this one. 

v3 has a dependency on an old library that could pose a security risk.

beeme1mr:
Thanks @afharo! We should probably mark this as a breaking change since it requires the user to upgrade to the latest LD SDK due ot the peer dependency update. FYI @toddbaert @kinyoklion @lukas-reining

afharo:
@beeme1mr that makes sense! Is there something I need to do?

lukas-reining (approved):
> @beeme1mr that makes sense! Is there something I need to do?

No @afharo, the PR title needed to include the `!` to mark it as breaking. 
I changed that :)

@kinyoklion can you also have a quick look?

kinyoklion:
Taking a look.

In regards to the dependency, I assume you mean uuid. We didn't use any of the offending methods (we used v1 without a pre-allocated buffer, and all the problems were for different versions with pre-allocated buffers) and the latest 3.x removes the dependency entirely and uses the implementation we made for 4.x. 

Of course we will still want to update to 4.x as soon as possible.

libs/providers/launchdarkly-client/src/lib/launchdarkly-client-provider.ts
kinyoklion, line 145: One thing worth pointing out here is that there is a default timeout on identify now of 5 seconds. So we may want to plumb the timeout option to this.

The other thing is we support "shedding" now. Which basically means if I rapid fire call identify then intermediates will be discarded.

```
identify({key: 'A'});
identify({key: 'B'});
identify({key: 'C'});
identify({key: 'D'});
identify({key: 'E'});
```

If you did this, then only A and E would identify and the intermediates would be shed. This won't have any impact if these onContextChanges are already sequentially awaited. 

You can disable shedding with an identify option.

kinyoklion (approved):

joker23 (approved):

kinyoklion:
LGTM. A couple minor notes.

Thank you for getting this updated!

toddbaert (approved):

PR #1566, opened 2026-06-18
fix(ofrep-web-provider): use base URL + eval context for cache key

MattIPv4:
## This PR

Updates the cache key logic to use the base URL + full evaluation context. This ensures that two OFREP providers with different URLs or evaluation contexts but the same targeting key do not attempt to share a cache key, as they may have different flag results.

### Related Issues

N/A

### Notes

As caching is implemented at the provider level, rather than the SDK level, this breaks out of the isolation that the domain paradigm usually gives, especially in micro-frontend setups. So, we need to be very careful to make sure that two providers running at the same time are truly independent of each other unless they are doing exactly the same thing.

### Follow-up Tasks

N/A

### How to test

Test suite updated, ensure caching still works from an end-user perspective as expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local cache persistence so storage entries are keyed by the endpoint base URL and the full evaluation context, preventing collisions across context changes.
  * When the evaluation context changes (even if the targeting key stays the same), previously persisted cached state is now cleared to avoid stale ETag/flags reuse.
* **New Features**
  * Added an optional `cacheKeyTransformer` option to customize (or omit) parts of the evaluation context used for cache key generation.

### DIFF
--- a/libs/providers/ofrep-web/src/lib/store/storage.ts
+++ b/libs/providers/ofrep-web/src/lib/store/storage.ts
@@ -13,7 +13,8 @@ const STORAGE_NS = 'ofrep-web-provider';
 export interface PersistedEntry {
   /** Schema version — used to discard entries written by older provider versions. */
   version: number;
-  /** Hex digest (16 chars) of `targetingKey` (or `cacheKeyPrefix + ":" + targetingKey` when configured). SHA-256 when crypto.subtle is available, FNV-1a fallback otherwise. */
+  /** Hex digest (16 chars) of `baseUrl + ":" + context` (or `cacheKeyPrefix + ":" + baseUrl + ":" + context` when configured).
+   *  SHA-256 when crypto.subtle is available, FNV-1a fallback otherwise. */
   cacheKeyHash: string;
   /** ETag returned by the server for this evaluation, used for efficient revalidation. */
   etag: string | null;
@@ -43,12 +44,12 @@ export class Storage {
   }
 
   /**
-   * Hashes the URL + evaluation context (with optional prefix) to a 16-char hex string.
+   * Hashes the base URL + evaluation context (with optional prefix) to a 16-char hex string.
    * Uses SHA-256 via crypto.subtle when available (secure contexts).
    * Falls back to a double-pass FNV-1a-32 in non-secure contexts where crypto.subtle is absent.
    */
-  private async _hashInput(url: string, context: EvaluationContext | undefined): Promise<string> {
-    const input = [this._cacheKeyPrefix ?? '', url, context ? JSON.stringify(context) : ''].join(':');
+  private async _hashInput(baseUrl: string, context: EvaluationContext | undefined): Promise<string> {
+    const input = [this._cacheKeyPrefix ?? '', baseUrl, context ? JSON.stringify(context) : ''].join(':');
     if (typeof crypto !== 'undefined' && crypto.subtle) {
       const encoded = new TextEncoder().encode(input);
       const hashBuffer = await crypto.subtle.digest('SHA-256', encoded);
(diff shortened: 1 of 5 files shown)


libs/providers/ofrep-web/src/lib/store/storage.ts
MattIPv4, line 71: I'm concerned that slicing to just the first 16 chars here risks collisions still happening unexpectedly?

jonathannorris:
@MattIPv4 thanks for picking this up. I think the collision case you're solving here is already handled by the existing `cacheKeyPrefix` option, and going to full-context hashing actually conflicts with [ADR 0009](https://github.com/open-feature/protocol/blob/main/service/adrs/0009-local-storage-for-static-context-providers.md), so I'd recommend we go that route instead.

For the micro-frontend setup you described in [the Slack thread](https://cloud-native.slack.com/archives/C06E4DE6S07/p1780526010480869) (same `targetingKey`, different URLs/instances), setting `cacheKeyPrefix` to the base URL (or an env/instance identifier) should already isolate the providers without any code change. That's exactly what the prefix was added for: the ADR resolved URL/instance namespacing as the opt-in prefix rather than baking the base URL into the default key.

On the full-context part, ADR 0009 is already accepted and explicitly decided against keying on the full evaluation context:

> The cache key is intentionally derived from `targetingKey` alone rather than the full evaluation context. [...] hashing the full context is impractical for local-cache-first startup because many implementations set volatile context properties on initialization (e.g. `lastSessionTime`, `lastSeen`, `sessionId`) that would change the hash on every app restart, defeating the purpose of persistence.

We went back and forth on exactly this in the original ADR discussion ([protocol#64](https://github.com/open-feature/protocol/pull/64), see [this thread](https://github.com/open-feature/protocol/pull/64#discussion_r3075409817) and @guidobrei's point on volatile context props). The concrete issue with hashing `JSON.stringify(context)` by default: any app that sets session/time-based context properties (`sessionId`, `lastSeen`, etc.) gets a new cache key on every restart, so the persisted entry never hits and the offline-bootstrap behaviour the cache exists for is lost.

If we do want to revisit keying on more than `targetingKey`, I think that should go through a spec/ADR update first so the other OFREP providers (kotlin, swift) stay consistent, rather than diverging here. Happy to discuss further with @toddbaert if you'd like.

MattIPv4:
While I understand that the `cacheKeyPrefix` option is available here, I do not believe that solves the issue I'm raising.

As discussed in Slack, I think a fix of some sort needs to be made here. The current default behavior is _unsafe_ and _will_ result in flag evaluations being returned from the cache that do not match the context set in the provider, in situations where multiple providers are running at the same time with different contexts that share the same targeting key.

That is dangerous behavior for a library that shouldn't be considered acceptable as a default.

I understand the concerns around volatile props, but I believe that OpenFeature should be defaulting to safe behavior without silent failure modes (ensuring the cache key is based on the full configuration of the provider), rather than catering to handling all situations (volatile props).

I also believe that this is likely a problem unique to the JS SDK, as this breaks out of the domain isolation that you'd usually get when running multiple providers at once.

MattIPv4:
As a solution to the volatile props concern with hashing the full context, this PR could instead accept a transformer method, which would take in the URL + full context. By default, it would return the URL + full context for hashing, ensuring that the default guarantees safe and consistent behavior. However, if folks are sending volatile props, they can then pass their own transformer method that omits parts of the full context when returning it for hashing.

As was noted in the ADR, another alternative would be to pass the domain down so that you get the same isolation guarantee in this caching that you do in the rest of the SDK, but I suspect that is a much more complicated change to make, given that providers do not know the domain at all currently and can be used for multiple domains at once.

PR #1568, opened 2026-06-25
chore: add js web providers docs for unleash

irstavr:
As the provider targets the OpenFeature Web SDK, and the React/Angular SDKs are built on top of the Web SDK — so the same `UnleashWebProvider` works in all of them unchanged.

We update the docs to notify users how to use properly the provider for React and Angular frameworks.

### DIFF
--- a/libs/providers/unleash-web/README.md
+++ b/libs/providers/unleash-web/README.md
@@ -23,7 +23,7 @@ Suitable for connecting to an Unleash instance
 ## Installation
 
 ```shell
-$ npm install @openfeature/unleash-web-provider @openfeature/web-sdk
+npm install @openfeature/unleash-web-provider @openfeature/web-sdk
 ```
 
 ## Usage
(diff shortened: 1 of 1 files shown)


libs/providers/unleash-web/README.md
```diff
@@ -107,6 +107,126 @@ const evaluationCtx: EvaluationContext = {
...
+```tsx
+import { useFlag } from '@openfeature/react-sdk';
+
+function Page() {
+  const { value: isAwesomeFeatureEnabled } = useFlag('isAwesomeFeatureEnabled', false);
+  return <div>{isAwesomeFeatureEnabled ? 'Awesome feature is enabled!' : 'Not enabled.'}</div>;
+}
+```
+
+See the [OpenFeature React SDK documentation](https://openfeature.dev/docs/reference/technologies/client/web/react) for hooks, suspense, and re-render behavior.
```
jarebudev, line 158: When I originally wrote this provider I used the [react-test-app ](https://github.com/open-feature/react-test-app ) as a means to demonstrate locally as did not need much alteration, just steps 1 and 2 as described above.  As you are expanding the documentation it may be worth also considering including a link to that repo?

libs/providers/unleash-web/README.md
```diff
@@ -107,6 +107,126 @@ const evaluationCtx: EvaluationContext = {
...
+#### 3. Evaluate a flag with a directive
+
+```html
+<div *booleanFeatureFlag="'isAwesomeFeatureEnabled'; default: false; else: disabled">
+  This is shown when the feature flag is enabled.
+</div>
+<ng-template #disabled>This is shown when the feature flag is disabled.</ng-template>
+```
+
+See the [OpenFeature Angular SDK documentation](https://openfeature.dev/docs/reference/technologies/client/web/angular/) for the full set of directives and their `initializing`/`reconciling` templates.
```
jarebudev, line 224: similar to above, there is a https://github.com/open-feature/angular-test-app that you may also want to consider including as a link.

jarebudev (approved):

libs/providers/unleash-web/README.md
irstavr: thanks for your comments and review @jarebudev! :) 

addressed both below commented lines.

lukas-reining (approved):

lukas-reining:
Thank you @irstavr!

PR #1569, opened 2026-06-26
feat(ofrep-web): ADR-0009 domain-aware cache key + domainScoped

jonathannorris:
## Summary

- Replace `cacheKeyPrefix` with a `cacheKeyGenerator` per the [protocol ADR-0009 amendment](https://github.com/open-feature/protocol/pull/80). The default generator returns JSON-encoded key material from `baseUrl`, auth credential, bound `domain`, and `targetingKey`; the provider hashes that into `cacheKeyHash`.
- Declare `domainScoped: true` and create `Storage` only in `initialize(context, domain?)`, so nothing touches `localStorage` before the bound domain is known.
- Derive auth for the cache key from known auth headers only (`Authorization`, `Api-Key`, `X-Api-Key`, `X-Auth-Token`, `X-Access-Token`), matched case-insensitively and serialized with lowercased header names.
- Persist under schema v2 (`ofrep-web-provider:v2:{hash}`); v1 entries are discarded on read.
- Document caching, cache-key customization, and domain scoping in the provider README.
- Bump `@openfeature/web-sdk` peer dependency to `^1.10.0` (domain forwarding and `domainScoped` enforcement landed in [open-feature/js-sdk#1433](https://github.com/open-feature/js-sdk/pull/1433)).

## Motivation

OFREP web persistence needs the bound OpenFeature `domain` at init time to scope cache keys per ADR-0009. This supersedes #1566 (URL + full eval context keying).

Requires `@openfeature/web-sdk@1.10.0+` so the SDK forwards `domain` on `initialize` and enforces single-domain binding for `domainScoped` providers.

## Notes

- Use `cacheKeyGenerator` to namespace instances, drop auth for rotating tokens, or include stable context fields.
- `cacheKeyPrefix` is removed.

## Related Issues

- Supersedes #1566
- Relates to open-feature/js-sdk#1433, open-feature/protocol#80, open-feature/spec#393

### DIFF
--- a/libs/providers/ofrep-web/src/lib/store/cache-key.ts
+++ b/libs/providers/ofrep-web/src/lib/store/cache-key.ts
@@ -1,5 +1,14 @@
 import type { OFREPProviderBaseOptions } from '@openfeature/ofrep-core';
 
+/** Header names treated as auth credentials for cache key derivation (matched case-insensitively). */
+const AUTH_HEADER_NAMES = new Set([
+  'authorization',
+  'api-key',
+  'x-api-key',
+  'x-auth-token',
+  'x-access-token',
+]);
+
 export type CacheKeyParts = {
   cacheKeyPrefix?: string;
   baseUrl: string;
@@ -8,16 +17,18 @@ export type CacheKeyParts = {
   targetingKey: string;
 };
 
+function isAuthHeader(name: string): boolean {
+  return AUTH_HEADER_NAMES.has(name.toLowerCase());
+}
+
 /**
- * Derives a stable auth credential string from static and factory-supplied headers.
+ * Serializes known auth headers from static and factory-supplied options for cache keying.
  * Rotating tokens will change the cache key on each rotation; stable credentials separate caches as intended.
  */
 export async function deriveAuthCredential(options: OFREPProviderBaseOptions): Promise<string> {
   const entries = [...(options.headers ?? []), ...((await options.headersFactory?.()) ?? [])];
-  const credentialHeaders = entries
-    .filter(([name]) => name.toLowerCase() !== 'content-type')
-    .sort(([a], [b]) => a.localeCompare(b));
-  return JSON.stringify(credentialHeaders);
+  const authHeaders = entries.filter(([name]) => isAuthHeader(name)).sort(([a], [b]) => a.localeCompare(b));
+  return JSON.stringify(authHeaders);
 }
 
 /**
--- a/libs/providers/ofrep-web/src/lib/store/cache-key.ts
+++ b/libs/providers/ofrep-web/src/lib/store/cache-key.ts
@@ -1,0 +1,35 @@
+import type { OFREPProviderBaseOptions } from '@openfeature/ofrep-core';
+
+export type CacheKeyParts = {
+  cacheKeyPrefix?: string;
+  baseUrl: string;
+  auth: string;
+  domain: string;
+  targetingKey: string;
+};
+
+/**
+ * Derives a stable auth credential string from static and factory-supplied headers.
+ * Rotating tokens will change the cache key on each rotation; stable credentials separate caches as intended.
+ */
+export async function deriveAuthCredential(options: OFREPProviderBaseOptions): Promise<string> {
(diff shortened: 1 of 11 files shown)


MattIPv4:
👀 I will flag the same other concern I had in #1566 here, which is that while we're now passing all of this in as input to the cache key, it is still being truncated down to just the first 16 chars, and that concerns me that collisions will happen unexpectedly and result in unsafe behaviour still?

libs/providers/ofrep-web/src/lib/store/cache-key.ts
```diff
@@ -0,0 +1,46 @@
+import type { OFREPProviderBaseOptions } from '@openfeature/ofrep-core';
+
+/** Header names treated as auth credentials for cache key derivation (matched case-insensitively). */
+const AUTH_HEADER_NAMES = new Set([
```
jonathannorris, line 5: This is a bit of a choice to make, do we try and include only "known" auth header keys, or do we potentially include all headers in the cache key (which could potentially have headers that break the consistent cache like time-based headers).

jonathannorris:
@MattIPv4 re https://github.com/open-feature/js-sdk-contrib/pull/1569#issuecomment-4810865247

> I will flag the same other concern I had in #1566 here, which is that while we're now passing all of this in as input to the cache key, it is still being truncated down to just the first 16 chars, and that concerns me that collisions will happen unexpectedly and result in unsafe behaviour still?

The 16-char truncation is inherited from the original ADR implementation (first 64 bits of SHA-256). Worth being explicit about what that means in practice.

Collision risk depends on how many entries coexist in `localStorage` on the same origin, not how large the `targetingKey` space is. A deployment might have millions of possible targeting keys, but this provider clears the previous entry on `targetingKey` change and typically only holds one active entry per url/auth/domain binding. In practice that's a small number of keys per website.

For *n* simultaneous entries, collision probability scales roughly with **n² / (2 × 2⁶⁴)**:

| Coexisting entries | Approx. collision risk |
|--------------------|-------------------------|
| 5 | ~5 × 10⁻¹⁹ |
| 10 | ~2 × 10⁻¹⁸ |
| 100 | ~3 × 10⁻¹⁶ |

At that scale a collision would mean serving the wrong cached flags, but the odds are not something I'd optimize for here. Open to bumping to 32 hex chars if you feel strongly.

MattIPv4:
👍 Not strongly held at all, just want to make sure that has been thought through, which it seems it has!

libs/providers/ofrep-web/src/lib/store/storage.ts
```diff
@@ -93,17 +118,17 @@ export class Storage {
...
     metadata?: Record<string, unknown>,
     eventStreams?: EventStream[],
   ): Promise<void> {
     if (this._disabled) return;
-    const key = await this.getStorageKey(targetingKey);
+    const key = await this.getStorageKey(context);
     const entry: PersistedEntry = {
       version: SCHEMA_VERSION,
-      cacheKeyHash: await this._hashInput(targetingKey),
+      cacheKeyHash: await this._hashInput(context),
```
toddbaert: This essentially calls `_hashInput` twice, and I suppose it's return value could even change between calls - can we derive it once and re-use it?

libs/providers/ofrep-web/src/lib/store/storage.ts
```diff
@@ -93,17 +118,17 @@ export class Storage {
...
-    targetingKey: string,
+    context: EvaluationContext,
     flags: FlagCache,
     etag: string | null,
     metadata?: Record<string, unknown>,
     eventStreams?: EventStream[],
   ): Promise<void> {
     if (this._disabled) return;
-    const key = await this.getStorageKey(targetingKey);
+    const key = await this.getStorageKey(context);
```
toddbaert: Especially since this is `async`, should it perhaps be in the try/catch?

toddbaert (approved):
I left some small concerns - but overall makes sense to me.

One thing we should also do is bump the min JS SDK peer for this dep to the latest version (`@openfeature/web-sdk@1.10.0`), since only that supplies the domain.

lukas-reining (approved):
This looks good to me, besides what was already mention by @toddbaert regarding calling hash twice.

PR #1575, opened 2026-07-07
feat(azure-app-configuration): add Azure App Configuration provider

wilqq:
## Summary

- Add a new `@openfeature/azure-app-configuration-provider` server-side provider that resolves feature flags from Azure App Configuration using `@azure/app-configuration-provider` and `@microsoft/feature-management`. Supports connection string and Entra ID credential auth, configurable selectors and refresh polling, boolean flag evaluation (with targeting/filters), and variant-based string/number/object resolution.
- Fix boolean flag resolution to return the caller's default value with reason `DEFAULT` when a flag is not found, instead of delegating to `FeatureManager.isEnabled` (which could return `false` regardless of the default).
- Fix the OpenFeature workspace generator to detect `jest.config.cts` vs `jest.config.ts` (Nx + Jest 30), preventing a crash when scaffolding new providers/hooks.

## Test plan

- [x] Run `nx test azure-app-configuration` — all unit tests pass
- [x] Run `nx package azure-app-configuration` — library builds successfully
- [x] Verify boolean evaluation returns the default value when a flag key does not exist in Azure App Configuration
- [x] Verify boolean evaluation honors targeting filters (targeting key, groups) for existing flags
- [x] Verify variant resolution (`getStringValue`, `getNumberValue`, `getObjectValue`) returns variant configuration and raises `TypeMismatchError` on type mismatch
- [x] Verify `ConfigurationChanged` events are emitted when refresh detects flag changes (`enableRefresh: true`)
- [x] Smoke-test provider initialization with both connection string and `endpoint` + `DefaultAzureCredential`

### DIFF
--- a/libs/providers/azure-app-configuration/README.md
+++ b/libs/providers/azure-app-configuration/README.md
@@ -1,0 +1,96 @@
+# Azure App Configuration Provider
+
+## What is Azure App Configuration?
+
+[Azure App Configuration](https://learn.microsoft.com/azure/azure-app-configuration/overview) is a managed service that provides a central place to manage application settings and feature flags. Its feature management capabilities support simple on/off toggles as well as advanced scenarios such as targeting, time windows, percentage rollouts, and variant feature flags.
+
+This provider lets OpenFeature applications resolve feature flags stored in Azure App Configuration. Under the hood it uses the official [`@azure/app-configuration-provider`](https://www.npmjs.com/package/@azure/app-configuration-provider) to load and refresh feature flags, and [`@microsoft/feature-management`](https://www.npmjs.com/package/@microsoft/feature-management) to evaluate them.
+
+## Installation
+
+```
+$ npm install @openfeature/azure-app-configuration-provider @azure/identity
+```
+
+`@openfeature/server-sdk` and `@azure/identity` are peer dependencies. Install `@azure/identity` if you authenticate with a token credential (recommended). It is not required when authenticating with a connection string.
+
+## Usage
+
+The provider is a server-side provider. Register it with OpenFeature and wait for it to become ready.
+
+### Using a Microsoft Entra ID token credential (recommended)
+
+```typescript
+import { OpenFeature } from '@openfeature/server-sdk';
+import { AzureAppConfigurationProvider } from '@openfeature/azure-app-configuration-provider';
+import { DefaultAzureCredential } from '@azure/identity';
+
+const provider = new AzureAppConfigurationProvider({
+  endpoint: 'https://<your-store>.azconfig.io',
+  credential: new DefaultAzureCredential(),
+});
+
+await OpenFeature.setProviderAndWait(provider);
+
+const client = OpenFeature.getClient();
+const isEnabled = await client.getBooleanValue('Beta', false);
+```
+
+### Using a connection string
+
+```typescript
+import { OpenFeature } from '@openfeature/server-sdk';
+import { AzureAppConfigurationProvider } from '@openfeature/azure-app-configuration-provider';
+
+const provider = new AzureAppConfigurationProvider({
+  connectionString: process.env.AZURE_APPCONFIG_CONNECTION_STRING!,
+});
+
+await OpenFeature.setProviderAndWait(provider);
+```
+
+## Configuration
+
+Provide exactly one of `connectionString` or `endpoint` + `credential`.
+
+| Property              | Type                | Description                                                                                                     | Default                |
+| --------------------- | ------------------- | --------------------------------------------------------------------------------------------------------------- | ---------------------- |
+| `connectionString`    | `string`            | Connection string for the Azure App Configuration store. Mutually exclusive with `endpoint`/`credential`.       | —                      |
+| `endpoint`            | `string`            | Endpoint URL of the Azure App Configuration store. Requires `credential`.                                       | —                      |
+| `credential`          | `TokenCredential`   | Token credential (e.g. `DefaultAzureCredential`) used with `endpoint`.                                          | —                      |
+| `selectors`           | `SettingSelector[]` | Feature flag selectors (`keyFilter`, `labelFilter`, ...) used to filter which feature flags are loaded.         | `[{ keyFilter: '*' }]` |
+| `enableRefresh`       | `boolean`           | Whether to periodically refresh feature flags and emit `ConfigurationChanged` events when a change is detected. | `true`                 |
+| `refreshIntervalInMs` | `number`            | Polling interval (in milliseconds) used to refresh feature flags. Must be greater than 1000.                    | `30000`                |
+
+## Flag resolution semantics
+
+Azure App Configuration feature flags map onto OpenFeature evaluation methods as follows:
+
+- **Boolean** (`getBooleanValue`): resolves the feature flag's enabled state via `FeatureManager.isEnabled`. This honors any configured feature filters (targeting, time window, percentage rollout, etc.).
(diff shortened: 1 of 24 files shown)


libs/providers/azure-app-configuration/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+[Azure App Configuration](https://learn.microsoft.com/azure/azure-app-configuration/overview) is a managed service that provides a central place to manage application settings and feature flags. Its feature management capabilities support simple on/off toggles as well as advanced scenarios such as targeting, time windows, percentage rollouts, and variant feature flags.
+
+This provider lets OpenFeature applications resolve feature flags stored in Azure App Configuration. Under the hood it uses the official [`@azure/app-configuration-provider`](https://www.npmjs.com/package/@azure/app-configuration-provider) to load and refresh feature flags, and [`@microsoft/feature-management`](https://www.npmjs.com/package/@microsoft/feature-management) to evaluate them.
+
+## Installation
+
+```
+$ npm install @openfeature/azure-app-configuration-provider @azure/identity
+```
```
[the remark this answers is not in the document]
wilqq: Current solution is common practice across other providers

libs/providers/azure-app-configuration/tsconfig.spec.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../dist/out-tsc",
```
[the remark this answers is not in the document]
wilqq: This file was auto generated by nx. All new providers, like [aws-ssm](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/aws-ssm) are using the same path, as provided solution

libs/providers/azure-app-configuration/src/lib/azure-app-configuration-provider.ts
```diff
@@ -0,0 +1,197 @@
...
+        errorMessage: `Flag '${flagKey}' was not found.`,
+      };
+    }
+
+    const value = await featureManager.isEnabled(flagKey, targetingContext);
+
+    return {
+      value,
+      reason: StandardResolutionReasons.TARGETING_MATCH,
+    };
```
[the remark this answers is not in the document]
wilqq: fixed

libs/providers/azure-app-configuration/src/lib/resolution-reason.ts
```diff
@@ -0,0 +1,145 @@
...
+  const audienceContextId = `${userId ?? ''}\n${hint}`;
+  const contextMarker = createHash('sha256').update(audienceContextId).digest().readUInt32LE(0);
+  const contextPercentage = (contextMarker / 0xffffffff) * 100;
+
+  if (to === 100) {
+    return contextPercentage >= from;
+  }
+
+  return contextPercentage >= from && contextPercentage < to;
+}
```
toddbaert: Am I understanding correctly that this is a re-implementation of  `@microsoft/feature-management`'s bucketing/random assignment? Is this just to supply a proper resolution reason?

If so I think this is a bad idea for 2 reasons:

1 - it might drift
2 - it's doing extra work for what is an optional reporting/telemetry signal

The proper solution would be to get some kind of reason code from the underlying system (a change in the library), and map it to OpenFeature. Otherwise, I would just to "best effort" for reasons or omit them entirely (they are optional).

wilqq: @toddbaert Thank you for review!

Good catch, that was a re-implementation of the library's bucketing just to get a reason, and I agree it's not worth it. I've removed it

Since reason is optional, I switched to best-effort using data the library already gives us

Can you please review it again

libs/providers/azure-app-configuration/src/lib/azure-app-configuration-provider.ts
```diff
@@ -0,0 +1,210 @@
...
+      return {
+        value: defaultValue,
+        reason: StandardResolutionReasons.ERROR,
+        errorCode: ErrorCode.FLAG_NOT_FOUND,
+        errorMessage: `Flag '${flagKey}' was not found.`,
+      };
+    }
+
+    const enabled = await featureManager.isEnabled(flagKey, targetingContext);
+    const variant = await featureManager.getVariant(flagKey, targetingContext);
```
toddbaert: I think we can run these in parallel to speed things up? (`Promise.all`)

libs/providers/azure-app-configuration/README.md
```diff
@@ -0,0 +1,96 @@
...
+| `refreshIntervalInMs` | `number`            | Polling interval (in milliseconds) used to refresh feature flags. Must be greater than 1000.                    | `30000`                |
+
+## Flag resolution semantics
+
+Azure App Configuration feature flags map onto OpenFeature evaluation methods as follows:
+
+- **Boolean** (`getBooleanValue`): resolves the feature flag's enabled state via `FeatureManager.isEnabled`. This honors any configured feature filters (targeting, time window, percentage rollout, etc.).
+- **String / Number / Object** (`getStringValue`, `getNumberValue`, `getObjectValue`): resolve [variant feature flags](https://learn.microsoft.com/azure/azure-app-configuration/howto-variant-feature-flags-javascript) via `FeatureManager.getVariant`. The variant's `configuration` value is returned, and the variant name is exposed as the resolution `variant`. If no variant is assigned, the provided default value is returned with reason `DEFAULT`. A `TypeMismatchError` is raised if the variant configuration does not match the requested type.
```
toddbaert, line 70: This is a great thing to mention. 👍

libs/providers/azure-app-configuration/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@openfeature/azure-app-configuration-provider",
+  "version": "0.0.1",
```
toddbaert: This should really match whatever is in the manifest: https://github.com/open-feature/js-sdk-contrib/pull/1575/changes#diff-e38bb01e41ed5baa3c9e8111b86009ca62d4498d2aefd14d367cd6300dd8b126R28

Either `0.0.1` or `0.1.0` is fine, but sync them up.

Note than the first release will bump it one more time, but that's fine. As long as you don't specific it will stay sub v1.

libs/providers/azure-app-configuration/src/lib/azure-app-configuration-provider.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    if (!variant || variant.configuration === undefined || variant.configuration === null) {
```
toddbaert, line 175: If the flag exists but no variant/configuration is assigned, this returns `FLAG_NOT_FOUND`. The README says this case returns the caller’s default with reason `DEFAULT`, which seems more accurate. Can we align the implementation with that behavior?

(or am I misreading?)

toddbaert (approved):
Approving, with 3 nits: [here](https://github.com/open-feature/js-sdk-contrib/pull/1575/changes#r3573437961) and [here](https://github.com/open-feature/js-sdk-contrib/pull/1575#discussion_r3573467138)and [here](https://github.com/open-feature/js-sdk-contrib/pull/1575#discussion_r3573545250), and a request to add yourself as the [component-owner](https://github.com/open-feature/js-sdk-contrib/blob/main/.github/component_owners.yml) for this.

Once these are done I will merge. :+1:

beeme1mr (approved):

wilqq:
@toddbaert @beeme1mr  thank you again for your review 👍 
All changes made. I think you can merge now

PR #1576, opened 2026-07-10
fix(flipt, flipt-web): honor default variant in variant flag evaluation

JessieAMorris:
## This PR
Flipt serves a flag's configured default variant (enabled flag, no matching rules) with match=false and DEFAULT_EVALUATION_REASON. The string/number/object resolver only checked match and discarded the served variant, returning the application-provided default value instead. The boolean resolver already handles this reason correctly.

Return the served variant (or its attachment for object flags) with resolution reason DEFAULT, and keep falling back to the application default when no default variant is configured.

### How to test
Add a string type variant to flipt with only the default value, like so:
Screenshot 2026-07-10 at 13 13 53

Then, use the feature flag with a different default value passed into OpenFeature. The result will return the default from the code call, not the default from the flipt server. Expected value is the value from the flipt server.


lukas-reining (approved):
This makes sense to me. Thanks @JessieAMorris!
The same counts for the server side flipt provider right @markphelps @JessieAMorris?

JessieAMorris:
> This makes sense to me. Thanks @JessieAMorris! The same counts for the server side flipt provider right @markphelps @JessieAMorris?

I just added the server side flipt changes as well.

PR #1577, opened 2026-07-12
fix(ofrep-core): encode flagKey in the evaluate request path

### LINKED ISSUE #1324: OFREP providers don't encode flagKey
It seems to me that we aren't properly encoding the /{flagKey} path segment in OFREP requests [here](https://github.com/open-feature/js-sdk-contrib/blob/4e1202982efdc3b08a3dec951155729a2d775e7f/libs/shared/ofrep-core/src/lib/api/ofrep-api.ts#L113C70-L113C77).

This means that flags with certain characters in them `/` for example, will cause errors, and potentially present security issues.

We should encode this segment.

cc @juanparadox

sarmah-rup:
## Bug

`OFREPApi.postEvaluateFlag(flagKey)` puts the raw `flagKey` into the request path. A key containing `/` (or other reserved characters like `?`, `#`, spaces) produces a malformed URL or adds extra path segments, changing which endpoint is hit. Issue #1324 flags this as a correctness and security concern.

## Root cause

```ts
let url = `${this.baseOptions.baseUrl}/ofrep/v1/evaluate/flags/${flagKey}`;
```

The `flagKey` is interpolated into the path without encoding. The bulk endpoint (`postBulkEvaluateFlags`) does not carry the key in the path, so this is the only affected site.

## Solution

Encode the segment:

```ts
`.../evaluate/flags/${encodeURIComponent(flagKey)}`
```

Fixes #1324

## Testing

Added a unit test to `ofrep-api.spec.ts` using the suite's existing `jest.fn` fetch mock to capture the outgoing `Request.url`, asserting `postEvaluateFlag('flag/with/slash')` encodes to `.../flags/flag%2Fwith%2Fslash`. Fails before the change (raw slashes), passes after. Full `ofrep-core` suite: 45/45; lint clean.

### DIFF
--- a/libs/shared/ofrep-core/src/lib/api/ofrep-api.spec.ts
+++ b/libs/shared/ofrep-core/src/lib/api/ofrep-api.spec.ts
@@ -175,6 +175,25 @@ describe('OFREPApi', () => {
         expect(result.httpStatus).toEqual(200);
       });
 
+      it('encode the flagKey in the evaluate path so reserved characters do not alter the URL', async () => {
+        const capturedUrls: string[] = [];
+        const mockFetch = jest.fn(async (input: RequestInfo | URL) => {
+          capturedUrls.push((input as Request).url);
+          return new Response(
+            JSON.stringify({ key: 'flag/with/slash', value: true, reason: 'STATIC', variant: 'on' }),
+            {
+              status: 200,
+              headers: { 'Content-Type': 'application/json' },
+            },
+          );
+        }) as unknown as FetchAPI;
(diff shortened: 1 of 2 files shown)


MattIPv4 (approved):
This seems reasonable, as there doesn't appear to be any restrictions on what characters can be in a flag key string: https://openfeature.dev/specification/glossary#flag-key

lukas-reining (approved):
This makes sense, thank you @sarmah-rup!

thomaspoignant (approved):

PR #1583, opened 2026-07-20
feat: Allow overriding SSE message parsing for compat.

JamieSinn:
## This PR
Not all SSE vendors (namely ably) support directly deserializing the data object with custom body content, and instead wrap it inside of another layer (data.data in Ably's case).
For compatibility with a lot more providers, non-JSON payloads, etc. this will allow more broad implementation adoption.

### Related Issues

https://github.com/open-feature/protocol/pull/82

### DIFF
--- a/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
+++ b/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
@@ -492,6 +492,9 @@ export class OFREPWebProvider implements Provider {
           this._options.inactivityDelaySec,
           this._logger,
           this._options.baseUrl,
+          // Leave EventSource implementation as the default; only tests inject a mock
+          undefined,
+          this._options.sseEventParser,
         );
       }
       this._sseManager.connect(result.eventStreams);
--- a/libs/providers/ofrep-web/src/lib/sse-manager.ts
+++ b/libs/providers/ofrep-web/src/lib/sse-manager.ts
@@ -1,4 +1,4 @@
-import type { BulkEvaluationRefetchMetadata, EventStream } from '@openfeature/ofrep-core';
+import type { BulkEvaluationRefetchMetadata, EventStream, EventStreamMessage } from '@openfeature/ofrep-core';
 import type { Logger } from '@openfeature/web-sdk';
 
 const DEFAULT_INACTIVITY_DELAY_SEC = 120;
@@ -21,6 +21,20 @@ export interface SseManagerCallbacks {
   onStale: () => void;
   onError: () => void;
 }
+
+/**
+ * Parses the payload of an incoming SSE {@link MessageEvent} into the object the
+ * SseManager inspects for `type`/`etag`/`lastModified`. Override to support
+ * non-JSON payloads or custom wire formats.
+ */
+export type SseEventParser = (event: MessageEvent) => EventStreamMessage;
+
+/**
+ * Default parser: JSON-parses string payloads, otherwise passes the data through
+ * as-is (e.g. when the EventSource implementation already deserialized it).
+ */
+export const defaultSseEventParser: SseEventParser = (event) =>
+  typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
 
 /**
  * Resolves the connection URL from an EventStream entry.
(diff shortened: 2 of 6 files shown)


libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
```diff
@@ -492,6 +492,9 @@ export class OFREPWebProvider implements Provider {
           this._options.inactivityDelaySec,
           this._logger,
           this._options.baseUrl,
+          // Leave EventSource implementation as the default; only tests inject a mock
+          undefined,
+          this._options.sseEventParser,
```
jonathannorris, line 497: I wonder if we should re-order these...

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -22,6 +22,20 @@ export interface SseManagerCallbacks {
   onError: () => void;
 }
 
+/**
+ * Parses the payload of an incoming SSE {@link MessageEvent} into the object the
+ * SseManager inspects for `type`/`etag`/`lastModified`. Override to support
+ * non-JSON payloads or custom wire formats.
+ */
+export type SseEventParser = (event: MessageEvent) => EventStreamMessage;
```
jonathannorris, line 30: should `SSE` be capitalized here?

JamieSinn: This follows naming conventions, you generally don't capitalize acronyms in pascal case

libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts
JamieSinn: Not super opinionated here, but I also don't see an immediately logical reason to do this as it's only done for tests. This class isn't exposed for users to manage/configure.

libs/providers/ofrep-web/src/lib/sse-manager.ts
toddbaert: Ya this is proper TS casing AFAIK.

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -63,6 +77,8 @@ export class SseManager {
     private _baseUrl?: string,
     // Allows tests to inject a mock instead of the global EventSource
     private _EventSourceImpl: new (url: string) => EventSource = EventSource,
+    // Overrides how incoming SSE payloads are parsed; defaults to JSON parsing
+    private _parseEvent: SseEventParser = defaultSseEventParser,
```
toddbaert, line 81: Probably worth a test that asserts a custom parser is used if supplied.

toddbaert (approved):
Approved in theory, but I'd like to see a [test](https://github.com/open-feature/js-sdk-contrib/pull/1583/changes#r3625258911).

libs/providers/ofrep-web/src/lib/sse-manager.ts
```diff
@@ -22,6 +22,20 @@ export interface SseManagerCallbacks {
...
+ * SseManager inspects for `type`/`etag`/`lastModified`. Override to support
+ * non-JSON payloads or custom wire formats.
+ */
+export type SseEventParser = (event: MessageEvent) => EventStreamMessage;
+
+/**
+ * Default parser: JSON-parses string payloads, otherwise passes the data through
+ * as-is (e.g. when the EventSource implementation already deserialized it).
+ */
+export const defaultSseEventParser: SseEventParser = (event) =>
```
erka, line 36: Out of curiosity, why should it be available to the public?

JamieSinn: my thought was that you can extend or take advantage of the base formatter as a step in your own formatter without having to re-write and copy/paste code. purely a DRY concept.
eg. Ably's SSE wrapper is just a single layer encapsulation, so the parser would be just wrapping the extraction of the upper `data` property and then calling the existing parser on that unwrapped object.

erka: thanks for explaining. It might be worth adding test(s) for this too, so we can catch any future changes that break the contract.

JamieSinn: we have some here for the format parsing/testing - https://github.com/open-feature/js-sdk-contrib/pull/1583/changes#diff-021a03ec38152040c9823ee0641451fed19475b688036a6c87e84ea17c6250bfR304-R375

erka: I don't see any tests which call `defaultSseEventParser` directly.

JamieSinn: Ah that makes more sense. Will update

JamieSinn: Added

jonathannorris (approved):

lukas-reining (approved):

erka (approved):

PR #1588, opened 2026-07-23
feat(flagd): Support client-certificate mTLS

### LINKED ISSUE #1587: flagd provider: add client-certificate mTLS support to the gRPC channel
## Problem

The flagd provider does not support mutual TLS (mTLS). The provider builds the gRPC channel with `credentials.createSsl()`. It calls this function with no arguments. Because of this, the provider cannot send a client certificate.

The `Config` interface has one TLS field for the server certificate (`certPath`). It has no field for a client certificate. It has no field for a client key. The provider has no other way to give a client certificate to the channel.

The transport layer supports mTLS. The function `credentials.createSsl(rootCerts, privateKey, certChain)` from `@grpc/grpc-js` accepts a client key and a client certificate. The provider does not use these two parameters.

## Effect

A user cannot connect the flagd provider to a flagd endpoint that requires a client certificate. Some deployments configure flagd to accept only clients that present a valid client certificate. The provider cannot connect to these endpoints. This is also a problem for local development, when a developer uses mTLS to test a secure connection.

The Java flagd library and the Go flagd library support client-certificate mTLS. The JavaScript provider does not. This is a difference between the language implementations.

## Proposed change

1. Add two options to the `Config` interface:
   - `clientCertPath` — the path to the client certificate.
   - `clientKeyPath` — the path to the client private key.
2. Add two environment variables:
   - `FLAGD_CLIENT_CERT_PATH`
   - `FLAGD_CLIENT_KEY_PATH`
3. Change the function that builds the channel credentials. If the user sets both `clientCertPath` and `clientKeyPath`, and both files exist, build the credentials with `credentials.createSsl(rootCerts, privateKey, certChain)`.
4. Keep the current behavior for all other conditions. If the user does not set the two client options, the provider works as before.

## Compatibility

The change is backward compatible. The mTLS path starts only when both new options are present. The default behavior does not change.

## Scope

The change is small. It touches the credential-build function, the configuration, and the documentation. Both resolver types use the same credential-build function. Because of this, the change applies to the RPC resolver and the in-process resolver.

tuhinkanti:
## This PR

Closes #1587

Adds client-certificate mutual TLS (mTLS) support to the flagd provider gRPC channel.

## Problem

The provider builds the gRPC channel with `credentials.createSsl()` and calls it with no arguments, so it cannot present a client certificate. The `Config` interface has a server-certificate field (`certPath`) but no field for a client certificate or client key. The `@grpc/grpc-js` function `credentials.createSsl(rootCerts, privateKey, certChain)` supports mTLS, but the provider does not use these parameters.

## Change

- Add `clientCertPath` and `clientKeyPath` options to `Config`.
- Add `FLAGD_CLIENT_CERT_PATH` and `FLAGD_CLIENT_KEY_PATH` environment variables.
- Update `createChannelCredentials`: when both client paths are set and both files exist, build credentials with `credentials.createSsl(rootCerts, privateKey, certChain)`.
- Thread the two options through the RPC resolver (`grpc-service.ts`) and the in-process resolver (`grpc-fetch.ts`).
- Document the two options in the README.

## Compatibility

Backward compatible. The mTLS path starts only when both new options are present. The default behavior does not change.

## Tests

New unit tests for `createChannelCredentials` cover insecure, server-side TLS, and mTLS credential construction. They use real temporary cert files and spy on the `@grpc/grpc-js` `credentials` factory to assert the exact arguments.

```
$ npx nx test providers-flagd -- -t "createChannelCredentials"

  createChannelCredentials
    ✓ should create insecure credentials when TLS is disabled
    ✓ should create SSL credentials with system trust store when no cert paths are set
    ✓ should create SSL credentials with a root cert when certPath is set
    ✓ should create mTLS credentials when both client cert and key paths are set
    ✓ should create mTLS credentials with null root certs when certPath is omitted
    ✓ should fall back to server-side TLS when only the client cert path is set

Tests: 6 passed, 6 total
```

Full provider suite is green with the change:

```
$ npx nx test providers-flagd

Test Suites: 6 passed, 6 total
Tests:       165 passed, 165 total
```

### DIFF
--- a/libs/providers/flagd/README.md
+++ b/libs/providers/flagd/README.md
@@ -34,6 +34,8 @@ Options can be defined in the constructor or as environment variables. Construct
 | tls                   | FLAGD_TLS                      | boolean  | false                                                          |                                              |
 | socketPath            | FLAGD_SOCKET_PATH              | string   | -                                                              |                                              |
 | certPath              | FLAGD_SERVER_CERT_PATH         | string   | -                                                              |                                              |
+| clientCertPath        | FLAGD_CLIENT_CERT_PATH         | string   | -                                                              | mTLS; requires clientKeyPath and tls         |
+| clientKeyPath         | FLAGD_CLIENT_KEY_PATH          | string   | -                                                              | mTLS; requires clientCertPath and tls        |
 | resolverType          | FLAGD_RESOLVER                 | string   | rpc                                                            | rpc, in-process                              |
 | offlineFlagSourcePath | FLAGD_OFFLINE_FLAG_SOURCE_PATH | string   | -                                                              |                                              |
 | selector              | FLAGD_SOURCE_SELECTOR          | string   | -                                                              | rpc & in-process (see [Selector](#selector)) |
--- a/libs/providers/flagd/src/lib/configuration.spec.ts
+++ b/libs/providers/flagd/src/lib/configuration.spec.ts
@@ -40,6 +40,8 @@ describe('Configuration', () => {
     const tls = true;
     const socketPath = '/tmp/flagd.socks';
     const certPath = '/etc/cert/ca.crt';
+    const clientCertPath = '/etc/cert/client.crt';
+    const clientKeyPath = '/etc/cert/client.key';
     const maxCacheSize = 333;
     const cache = 'disabled';
     const resolverType = 'in-process';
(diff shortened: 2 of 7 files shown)


Baltsat (commented):
Reviewed exact head `64dce6f115e497a9d79e4ccd4860736d13406cb7` against #1587. Confirmed the constructor and environment configuration paths, `createSsl(rootCerts, privateKey, certChain)` argument order, wiring through both RPC implementations, and the documented one-way TLS fallback when the client pair is incomplete.

Local verification on Node 24:
- `npx nx test providers-flagd --skip-nx-cache`: 6 suites / 165 tests passed
- `npx nx lint providers-flagd --skip-nx-cache`: 0 errors (10 existing warnings)
- Prettier check: all 7 changed files passed

No blocking findings on this head.

tuhinkanti:
Any feedback?

tuhinkanti:
PR #1630 supersedes this path-specific change. It adds generic `ChannelCredentials` injection and preserves the current `tls` and `certPath` behavior when the new option is absent.

The caller owns credential loading and rotation. The flagd provider continues to own its gRPC transport. Issue #1628 contains the design and compatibility requirements.

PR #1631 and issue #1629 track the separate readiness and shutdown fixes that support safe provider replacement.

PR #1589, opened 2026-07-25
fix(flagd): stop reconnecting after disconnect

### LINKED ISSUE #1484: flagd-provider 0.13.4: disconnect() does not clear reconnect timers, causing Node.js process hang
reconnect() in grpc-service.ts and grpc-fetch.ts creates a setTimeout without storing the reference
disconnect() never clears this timer → Node.js event loop stays alive
After the timer fires, listen() is called on a closed client → infinite reconnect loop
Fix: store the timer ref, clear it in disconnect(), add a _disconnected guard flag

Introduced in PRs #1437, #1438, #1445, #1458

Baltsat:
## This PR

- retains and cancels pending reconnect timers in both flagd gRPC transports
- prevents ready and retry callbacks from reopening clients after shutdown
- rejects an in-flight initial connection when shutdown happens before the first ready payload
- covers retry cancellation and both startup race windows with deterministic tests

### Related Issues

Fixes #1484

### Notes

Both gRPC implementations scheduled reconnects without retaining the timer handle. A queued callback could therefore call `listen()` after `disconnect()` had closed the client. Shutdown during initial connection setup could also leave the connection promise unsettled.

### How to test

- `npx jest --config libs/providers/flagd/jest.config.ts --runInBand libs/providers/flagd/src/lib/flagd-provider.spec.ts libs/providers/flagd/src/lib/service/in-process/grpc/grpc-fetch.spec.ts` — 2 suites, 29 tests passed
- `npx nx run providers-flagd:test --skip-nx-cache --runInBand` — 6 suites, 165 tests passed
- `npx nx run providers-flagd:lint --skip-nx-cache` — passed with 0 errors and the existing warning baseline
- `npx tsc --noEmit -p libs/providers/flagd/tsconfig.lib.json` — passed
- Prettier and `git diff --check` — passed

### DIFF
--- a/libs/providers/flagd/src/lib/flagd-provider.spec.ts
+++ b/libs/providers/flagd/src/lib/flagd-provider.spec.ts
@@ -553,6 +553,59 @@ describe(FlagdProvider.name, () => {
         // eventStream should have been called twice (initial + reconnect attempt)
         expect(streamingServiceClientMock.eventStream).toHaveBeenCalledTimes(2);
       });
+
+      it('should not reconnect after disconnect when an error has scheduled a retry', async () => {
+        const provider = new FlagdProvider(
+          undefined,
+          undefined,
+          new GRPCService(
+            { deadlineMs: 100, streamDeadlineMs: 600000, host: '', port: 123, tls: false, cache: 'lru' },
+            streamingServiceClientMock,
+          ),
+        );
+        const initializePromise = provider.initialize();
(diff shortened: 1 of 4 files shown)


Baltsat:
Rebased onto the current main. The branch remains focused on making disconnect terminal for both transport implementations, with regression coverage for pending reconnect callbacks. Fresh CI is running now; a maintainer re-review would be appreciated once it is green.

PR #1591, opened 2026-07-27
feat(flagsmith-client)!: implement experimental tracking

Zaimwa9:
## Summary

- **Breaking**: migrate the peer dependency from `flagsmith` to `@flagsmith/flagsmith` `^12.1.2` — the Flagsmith JS SDK moved to the npm org scope at v12; the unscoped package stops at 10.x
- Populate `ResolutionDetails.variant` for multivariate flags (spec [2.2.4](https://openfeature.dev/specification/sections/providers#requirement-224))
- Populate `flagMetadata` on every resolution: `enabled`, `featureId`, plus `experiment.arm` / `experiment.active` / `experiment.unit` on multivariate flags
- Report honest resolution reasons: `DEFAULT`, `DISABLED` (configured value still returned), `CACHED`, `STATIC`, `TARGETING_MATCH`
- Implement `onClose()`: stop listening and flush buffered events; stop leaking identity traits into provider event metadata
- Implement experimental `track()` (spec [§6](https://openfeature.dev/specification/sections/tracking)): the reserved event name `feature_flag.exposure` (exported as `EXPOSURE_TRACKING_EVENT`) routes to Flagsmith exposure tracking; any other name becomes a plain Flagsmith event. All tracking surface is marked `@experimental`
- Fix `typeFactory` coercion: `"0"` now resolves to `0`; booleans are strict (non-boolean values throw `TypeMismatchError`)
- Fix `resolveBooleanEvaluation` ignoring the caller's `defaultValue` for missing flags (spec [1.4.3](https://openfeature.dev/specification/sections/flag-evaluation#requirement-143))

## Motivation

The provider was pinned several majors behind the Flagsmith SDK. Flagsmith 12.1+ ships an experimentation surface (variants, exposure events, product events), and this PR makes it reachable through the vendor-neutral OpenFeature API: evaluation details now carry the variant and experiment metadata, and `track()` records exposures and conversions without importing Flagsmith directly.

## Notes

- Consumers must swap `flagsmith` → `@flagsmith/flagsmith` in their dependencies and imports; hence `feat!`
- Tracking is experimental on both sides (OpenFeature spec §6 and Flagsmith's events pipeline) and is a no-op unless the provider is constructed with `enableEvents: true`
- Exposure events require an identified context (`targetingKey`); anonymous exposures are skipped and logged
- `TARGETING_MATCH` is approximated as *identified + fresh server flags* — the Flagsmith client API doesn't expose per-flag provenance (segment match vs environment default). Documented in the README
- The boolean-default fix changes behavior for missing flags (previously always `false`); it rides the major
- Verified functionally against a live Flagsmith environment side by side with the raw SDK: evaluation parity across experiment arms, and byte-identical exposure/custom-event wire payloads

## Test plan

- [x] `npx nx test providers-flagsmith-client` — 54 tests
- [x] `npx nx lint providers-flagsmith-client`
- [x] `npx nx run providers-flagsmith-client:package`

## OpenFeature
image


talissoncosta (approved):
Looks good!

libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts
```diff
@@ -166,26 +238,66 @@ export class FlagsmithClientProvider implements Provider {
...
-   * Based on Flagsmith's loading state and feature resolution, determine the Open Feature resolution reason
+   * Based on the flag state and Flagsmith's loading state, determine the Open Feature resolution reason
    * @private
    */
-  private parseReason(value: unknown): ResolutionReason {
-    if (value === undefined) {
-      return 'DEFAULT';
+  private parseReason(flag: IFlagsmithFeature): ResolutionReason {
+    if (!flag.enabled) {
+      return 'DISABLED';
```
toddbaert, line 298: Heads up: this looks like a new behavior change unrelated to tracking. On `main` a boolean flag resolving `false` reports `STATIC`/`CACHED`; this `!flag.enabled -> DISABLED` short-circuit makes every `false` boolean report `DISABLED` (returning the real value, not the default). Your call, but that reason-mapping is a fundamental provider decision, so might be worth splitting out from this PR.

libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts
```diff
@@ -108,14 +128,65 @@ export class FlagsmithClientProvider implements Provider {
...
+        this._logger?.info(`Exposure for "${flagKey}" skipped: no targetingKey in the evaluation context.`);
+        return;
+      }
+      if (typeof variant === 'string') {
+        this._client.trackExposureEvent(flagKey, { identifier, value: variant, metadata });
+      } else {
+        if (Object.keys(metadata).length > 0) {
+          this._logger?.debug(`Exposure for "${flagKey}": metadata is ignored when no variant is provided.`);
+        }
+        this._client.getExperimentFlag(flagKey);
```
toddbaert: Minor: the explicit-variant path above uses `context.targetingKey`, but this variant-less `getExperimentFlag()` resolves against Flagsmith's internal identity. Usually the same, but a quick code comment noting the difference would help future readers.

toddbaert (approved):
Approving - but I made 2 callouts - just take a look and make sure.

Otherwise happy to release this as soon as you guys want.

libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts
Zaimwa9: Agreed, missed this one. Fixed in [6df638e](https://github.com/open-feature/js-sdk-contrib/pull/1591/commits/6df638ec142ff9b5f1c8de6efd394e3892629ac7) to make sure it won't drift

libs/providers/flagsmith-client/src/lib/flagsmith-client-provider.ts
Zaimwa9: Yep true, that's deliberate. I looked into separating in 2 distinct PRs but it's not as easy/straightforward. The experiment hook relies on `TARGETING_MATCH` implying `enabled` to not expose users out of an experiment.
I documented in the readme [here](https://github.com/open-feature/js-sdk-contrib/pull/1591/changes#diff-9d62b3184a7136f71a42bb08dc776b1dff96d9c2903083e108907993d14ef930R98-R103) but let me know if it's not clear enough.

Also, we plan to quickly iterate on this implementation, especially for when we will surface experimentation descriptors (that our SDK lacks atm) and we will be able to guard these paths better

Zaimwa9:
@toddbaert thanks a lot for the review, I appreciate the reactivity! Gave the rationales for the first callout and addressed the second one 👍

toddbaert:
@Zaimwa9 here is the release (the release notes mention the break, but not the specific changes that are breaking - if you want to change them, just open a PR targeting the release branch: https://github.com/open-feature/js-sdk-contrib/pull/1594)

Zaimwa9:
@toddbaert done in #1595, thanks!

lukas-reining:
I merged #1595 @Zaimwa9 :)

PR #1625, opened 2026-09-14
chore: pin the flagd testbed and schemas submodules to their release tags

### LINKED ISSUE #1624: Restore the .gitmodules `branch` pin so Renovate tracks testbed versions again
open-feature/js-sdk-contrib#1461 dropped the `branch = v2.10.2` pin from the `libs/shared/flagd-core/test-harness` submodule in `.gitmodules`. It reads as harmless — the exact SHA is still pinned — but it is the line Renovate uses to track testbed versions, and testbed updates here have been dead ever since.

The submodule is at `b308c1b` (**v3.10.1**), which is current, so nothing is broken right now. The four bumps that got it there were all manual, carried along inside unrelated feature PRs. Without the pin, staying current depends on someone noticing.

The PR restores the pin at `v3.10.1`.

## Why this matters

Renovate's `git-submodules` manager reads `branch` from `.gitmodules` as the dependency's **current version**. With the pin present it resolves against the testbed's release tags and opens an ordinary version bump, rewriting the pin itself:

```diff
-	branch = v2.10.1
+	branch = v2.10.2
```

Without it there is no version to compare against, so Renovate falls back to chasing the tip of the testbed default branch — a digest update. Those don't match the `community-tooling` preset's `matchSourceUrls` auto-create exemption, so they get parked behind a dashboard checkbox instead of becoming PRs. Net effect: no automated testbed updates in this repo since the removal.

## The fix

The attached PR restores `branch = <tag>` on the testbed submodule, set to the tag it currently points at. That is the whole fix — no Renovate config change is needed, because the preset's existing exemption starts matching again as soon as the dependency resolves to a version.

Parent: open-feature/flagd-testbed#398

aepfli:
Closes #1624
Parent: open-feature/flagd-testbed#398

Restores the dropped testbed `branch` pin and extends it to the `flagd-schemas` submodules, so the repo is covered in one pass. Every pin is set to the tag the submodule already points at, so nothing moves.

Renovate reads `branch` in `.gitmodules` as the submodule's current version. With it, a release arrives as an ordinary version PR that rewrites both the pointer and the pin. Without it there is no version to compare against, so Renovate can only chase the upstream default branch — which is why this repo has had no automated testbed updates since the pin was removed, and which open-feature/community-tooling#126 turns off entirely.

| submodule | commit | pin |
|---|---|---|
| `libs/shared/flagd-core/test-harness` | `b308c1b` | `v3.10.1` |
| `libs/providers/flagd/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |
| `libs/providers/flagd-web/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |
| `libs/shared/flagd-core/flagd-schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |

`json/json-schema-*` is a release-please component tag and does not parse as semver, so the three schemas pins stay inert until open-feature/community-tooling#126 merges. Harmless until then, and the comment block in `.gitmodules` says so.

Left unpinned: the two `spec` submodules (untagged; pinning would roll them back to `v0.9.0`) and `wasm-releases`, which has no tags — only `publish-wasm-evaluation-v*` branches, and `TARGET_WASM_VERSION` in `copy-latest-wasm.js` would have to move in step.

Refs: open-feature/community-tooling#125


aepfli:
Pushed a commit extending this PR to the three `flagd-schemas` submodules, so the whole repo is covered in one pass rather than leaving a second round for later.

All three already sit exactly on `json/json-schema-v0.2.15`, so nothing moves:

| submodule | recorded commit | tag |
|---|---|---|
| `libs/providers/flagd/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |
| `libs/providers/flagd-web/schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |
| `libs/shared/flagd-core/flagd-schemas` | `1daf5ff` | `json/json-schema-v0.2.15` |

One caveat worth knowing: that is a release-please component tag and does not parse as semver, so on its own it resolves to nothing. It needs the prefix rule in open-feature/community-tooling#126 to become a real version. Until that merges the pin is inert, but harmless — and the comment block in `.gitmodules` says so.

Still deliberately unpinned here: the two `spec` submodules (pending the spec-cadence question in open-feature/community-tooling#125) and `wasm-releases`, which has no tags — only `publish-wasm-evaluation-v*` branches. That one also needs `TARGET_WASM_VERSION` in `copy-latest-wasm.js` kept in step, so a pin alone is not the whole change.

This branch is 1 commit behind `main`. Still mergeable, and git's three-way merge keeps main's copy of anything I did not touch.

